PR #1, opened 2022-05-27
ENG 1147: Updates linter configurations and runs linter

vsreekanti:
I first tried to use the TSDX linter, but it seems to have some very restrictive settings and/or is broken [based on this thread](https://github.com/jaredpalmer/tsdx/issues/428). It didn't allow much configurability, and obviously we want to keep code formatting uniform across our components, so this PR:

* clobbers the TSDX lint commands and replaces them with the standard eslint command
* ports over our `.eslintrc.js` and `.prettierrc.js`
* makes the linter configurations uniform across `common` and `app`
* runs the linter (hence the massive LOC changed)


src/ui/app/pages/index.tsx
```diff
@@ -1,25 +1,25 @@
...
-    return null;
-  }
+    if (!success) {
+        const router = useRouter();
+        router.push('/login');
+        return null;
+    }
 
-  //return <GettingStartedTutorial user={user} />;
+    //return <GettingStartedTutorial user={user} />;
```
agiron123, line 20: we can remove this comment, i left this in on accident :)

PR #2, opened 2022-05-31
Adds setup script to backend Python connector package

cw75:
With this change, we can `cd` into `src/python` and install the latest development version of the backend Python package via `pip3 install .`.


src/python/setup.py
```diff
@@ -0,0 +1,24 @@
...
+import os
+
+from setuptools import find_packages, setup
+
+
+install_requires = open("requirements.txt").read().strip().split("\n")
+
+setup(
+    name="aqueduct-ml",
+    version="0.0.1",
```
saurav-c, line 11: Is this something that should ever be changed, or is it supposed to be hard-coded?

saurav-c (approved):
LGTM! Left a question about versioning and just a reminder to add the `typing-extensions` package to the requirements.

src/python/setup.py
cw75: We should bump this when we are ready to cut another release.

cw75:
@saurav-c added the package into the requirements file.

PR #6, opened 2022-05-31
Add Github Actions for Integration Tests

kenxu95:
Runs the suite of integration tests with the publish flag with all our supported python versions (3.8 and 3.9). Each takes about 5 minutes to run.

The tests are triggered on push, and also by any comment that starts with "ci". I want us to start running "ci" before we merge PRs, since reverting a merged PR is more work then just checking it beforehand.

The server logs are uploaded to GH actions as an artifact, so you can inspect what went wrong on the backend.


.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,101 @@
...
+        env:
+          SERVER_ADDRESS: localhost:8080
+          INTEGRATION: aqueduct_demo
+        run: pytest . -rP --publish -n 5
+
+      - uses: actions/upload-artifact@v3
+        if: ${{ always() }}
+        with:
+          name: Server Logs
+          path: $SERVER_LOGS_FILE
```
saurav-c, line 100: Where are these logs uploaded? Is that something we can access after the action completes?

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,101 @@
...
+      - name: Install aqueduct-ml
+        working-directory: src/python
+        run: pip3 install .
+
+      - name: Startup the server and wait for it to initialize
+        run: (aqueduct server > $SERVER_LOGS_FILE 2>&1 &) && sleep 10
+
+      # Grabs the pid of the process bound to port 8080 and kills it.
+      - name: Kill the server
+        run: kill -9 $(lsof -nP -iTCP -sTCP:LISTEN | grep 8080 | awk '{print $2}')
```
saurav-c, line 73: Nit: Can we capture the PID of the server process with something like this instead of checking the PID of the process listening on port 8080: https://stackoverflow.com/a/1911387. It seems less hacky.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,101 @@
...
+
+      # install_local.py requires ~/.aqueduct/server scaffolding to exist.
```
saurav-c, line 75: Just to confirm, we start and then kill the server just so ~/.aqueduct/server exists?

saurav-c (commented):
Left a few questions below. Otherwise looks good!

.github/workflows/integration-tests.yml
kenxu95: Yep, you'll see them as artifacts in the job summary, which you can download :D 

image

.github/workflows/integration-tests.yml
kenxu95: Yeah the "$!" syntax doesn't work for me though, it returns the pid of the background python process, not the golang server actually bound to the port. Killing the golang server stops both processes. Killing the pid returned by $! doesn't.

.github/workflows/integration-tests.yml
kenxu95: yep its purely there for scaffolding purposes. Chengang and I discussed this and it seemed better to have something a bit ugly in this one internal instance than to figure out how to share code nicely between /bin/python/aqueduct and scripts/

sdk/setup.py
```diff
@@ -27,5 +27,5 @@
     classifiers=[
         "Programming Language :: Python :: 3",
     ],
-    python_requires=">=3.7,<=3.10",
```
likawind, line 30: cc @kenxu95 @cw75 I think this removal is necessary to enable supports for `3.10`. Should we do `<=3.11` instead?

kenxu95: I'd say just get rid of the upper restriction :)

kenxu95: Also can you make a separate PR for the 310 stuff, just to keep this one from getting more cluttered?

.github/workflows/integration-tests.yml
saurav-c: Sweet!

saurav-c (approved):

PR #7, opened 2022-06-01
Add package version command to cli

cw75:
This is the first step towards supporting versions. This PR allows users to retrieve the package version via `aqueduct version`. In a subsequent PR, we will code this version information to `~/.aqueduct` and add support for version update.

### DIFF
--- a/src/python/bin/aqueduct
+++ b/src/python/bin/aqueduct
@@ -16,6 +16,7 @@ base_directory = os.path.join(os.environ["HOME"], ".aqueduct")
 server_directory = os.path.join(os.environ["HOME"], ".aqueduct", "server")
 ui_directory = os.path.join(os.environ["HOME"], ".aqueduct", "ui")
 
+package_version = "0.0.1"
 
 def update_config_yaml(file):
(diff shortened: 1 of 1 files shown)


src/python/bin/aqueduct
```diff
@@ -206,6 +207,9 @@ def apiKey():
 def clear():
     execute_command(["rm", "-rf", base_directory])
 
+def version():
```
vsreekanti, line 210: How are we going to ensure that the package version logged here is in sync with the executables in `~/.aqueduct`?

cw75: Yeah, in a subsequent PR, my plan is to persist this version information into `~/.aqueduct`, and then whenever users do `aqueduct server` or `aqueduct ui`, we will compare the version number here with the persisted version number, and if there is any discrepancies (`aqueduct version` is ahead of the one persisted in `~/.aqueduct`), we will then pull the right version from S3 to upgrade the binaries. Does this design make sense?

kenxu95 (approved):
lgtm as first step

PR #10, opened 2022-06-01
Restrict supported python version to <3.10

kenxu95:
The expression `<=3.10` doesn't make sense since it only includes 3.10.0 and not 3.10.1, 3.10.2, etc.

It doesn't look like our current OSS even supports 3.10, I've filed a task for this: https://linear.app/aqueducthq/issue/ENG-1174/python-310-doesnt-work-with-oss

Error message:
```
Operator success_on_single_metric_input Failed! Error message:
 Traceback (most recent call last):
  File "/Users/kennethxu/.aqueduct/server/storage/operators/3fb8a254-1f92-4324-8e95-21b8ebba936f/op/model.py", line 10, in predict
    return self.func(*args)
  File "/Users/kennethxu/work/aqueduct/integration_tests/sdk/checks_test.py", line -1, in success_on_single_metric_input
SystemError: unknown opcode
```

We should address this issue soon, but in the meantime we should be less confusing with our supported python ranges.

### DIFF
--- a/sdk/setup.py
+++ b/sdk/setup.py
@@ -27,5 +27,5 @@ setuptools.setup(
     classifiers=[
         "Programming Language :: Python :: 3",
     ],
-    python_requires=">=3.7,<=3.10",
+    python_requires=">=3.7,<3.10",
 )


HarisChoudhary (approved):

cw75:
So it seems that we are supporting 3.10.x? In this case should we change to `<3.11`?

kenxu95:
3.10 doesn't work. @HarisChoudhary is working on it right now: https://linear.app/aqueducthq/issue/ENG-1174/python-310-doesnt-work-with-oss

I would prefer to remove this bound entirely, but then people with 3.10 will just see errors all the time. Do we want to be permissive here like we do for requirements, or do we want to be explicit with our supported python versions?

PR #12, opened 2022-06-01
Implement support for version upgrade

cw75:
This PR enables our users to update OSS package versions. When they perform `pip3 install --upgrade aqueduct-ml`, if we have a newer version, then when they subsequently execute `aqueduct server` or `aqueduct ui`, we will detect that their current version is older than the package version and fetch the latest binaries from S3. Their existing database state will be untouched.

## Test
Made a dummy version 0.0.2 and verified that the version upgrading from 0.0.1 to 0.0.2 worked.

### DIFF
--- a/src/python/bin/aqueduct
+++ b/src/python/bin/aqueduct
@@ -61,13 +61,14 @@ def generate_version_file(file_path):
     with open(file_path, 'w') as f:
         f.write(package_version)
 
+# Returns a bool indicating whether we need to perform a version upgrade.
 def require_update(file_path):
     if not os.path.isfile(file_path):
         return True
     with open(file_path, 'r') as f:
         current_version = f.read()
         if package_version < current_version:
-            raise Exception("Attempting to install an older version %s but found existing installation %s" % (package_version, current_version))
+            raise Exception("Attempting to install an older version %s but found existing newer version %s" % (package_version, current_version))
         elif package_version == current_version:
             return False
         else:
--- a/src/python/bin/aqueduct
+++ b/src/python/bin/aqueduct
@@ -57,6 +57,76 @@ def update_executable_permissions():
     execute_command(["chmod", "755", os.path.join(server_directory, "bin", "executor")])
     execute_command(["chmod", "755", os.path.join(server_directory, "bin", "migrator")])
 
+def generate_version_file(file_path):
+    with open(file_path, 'w') as f:
+        f.write(package_version)
+
+def require_update(file_path):
+    if not os.path.isfile(file_path):
+        return True
+    with open(file_path, 'r') as f:
+        current_version = f.read()
+        if package_version < current_version:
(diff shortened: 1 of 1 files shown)


src/python/bin/aqueduct
```diff
@@ -57,6 +57,77 @@ def update_executable_permissions():
...
+        f.write(package_version)
+
+# Returns a bool indicating whether we need to perform a version upgrade.
+def require_update(file_path):
+    if not os.path.isfile(file_path):
+        return True
+    with open(file_path, 'r') as f:
+        current_version = f.read()
+        if package_version < current_version:
+            raise Exception("Attempting to install an older version %s but found existing newer version %s" % (package_version, current_version))
```
kenxu95, line 71: Are we not allowing users to downgrade our packages at all? What if they want to go back to and older version because of breaking changes we introduced?

src/python/bin/aqueduct
```diff
@@ -57,6 +57,77 @@ def update_executable_permissions():
...
+    elif system == "Darwin" and arch == "arm64":
+        print("Operating system is Mac with architecture arm64.")
+        execute_command(["curl", os.path.join(s3_prefix, "bin/darwin_arm64/server"), "--output", os.path.join(server_directory, "bin/server")])
+        execute_command(["curl", os.path.join(s3_prefix, "bin/darwin_arm64/executor"), "--output", os.path.join(server_directory, "bin/executor")])
+        execute_command(["curl", os.path.join(s3_prefix, "bin/darwin_arm64/migrator"), "--output", os.path.join(server_directory, "bin/migrator")])
+    else:
+        raise Exception("Unsupported operating system and architecture combination: %s, %s" % (system, arch))
+
+def update_server_version():
+    print("Updating server version to %s" % package_version)
```
kenxu95, line 116: I don't see where `package_version` is ever updated from 0.0.1 - am I missing something?

src/python/bin/aqueduct
cw75: The problem with downgrading right now is that we haven't implemented the down migration of sqlite. The reason was that in SQLite, lots of column drop and constraint drop operations aren't supported, which makes it pretty tedious to implement down migration; we basically need to create a new table with old constraints and dump data from the current table into that. I think we should eventually still implement this and support it, but not sth we can support right now, so that's why I'm disabling downgrading now.

src/python/bin/aqueduct
cw75: When we decide to cut a release, we will make a PR that updates this hardcoded version number from 0.0.1 to 0.0.2.

kenxu95 (approved):

src/python/bin/aqueduct
saurav-c: Yeah, that makes sense for why we can't support downgrades right now (without wiping the database.)

PR #14, opened 2022-06-01
[2/4] Reorganize server's utils code

likawind:
This PR reorganizes server's utils code. It mainly achieves two goals for code reorg:
* moved any sharable helper functions to `/lib` instead of `/cmd/server`
* split the large `utils` package into smaller, specific pieces. For future code dev, I would like to push away from using package named 'utils' (which is vague) and do the following instead:
  * external helpers (intended to import by others) should have specific package name reflecting the 'function' of the package
  * For 'internal' helpers, we could still have `utils.go` files in certain package as collections of util functions. This doesn't violate the above for using more explicit package names

More specific changes:
* all resp utils goes to `cmd/server/response` package
* renamed `cmd/server/request_parser` to `cmd/server/request` to be symmetric with `response`. Moved previous `utils.ExtractHttpPayload` to this package.
* splitted previous `utils/constants` into specific packages. Most of them, including headers and url params, goes to `cmd/server/route`
* A few validation utils goes to `lib/workflow/dag` and `lib/workflow/operator`
* Built-in utils goes to `lib/workflow/demo`
* Moved all `commonArgs` parsing to `lib/context`. The main purpose is to be able to import `contextKey` constants from a commonly shared ancestor package. I think this is a valuable move in long run since we could certainly have request ID tracking for executors which would involve the same context passing around and parsings.

Not covered:
* I'm not exactly sure what we should do for `cmd/server/queries`. Planning to save this for future clean-up projects.
* We will clean up migrator code in similar way.

## Tests
This PR mostly contains moving code from A to B. Should be sufficient as long as all components compiles.


src/golang/cmd/server/middleware/authentication/authentication.go
```diff
@@ -4,8 +4,10 @@ import (
 	"context"
 	"net/http"
 
-	"github.com/aqueducthq/aqueduct/cmd/server/utils"
+	"github.com/aqueducthq/aqueduct/cmd/server/response"
+	"github.com/aqueducthq/aqueduct/cmd/server/routes"
 	"github.com/aqueducthq/aqueduct/lib/collections/user"
+	aq_context "github.com/aqueducthq/aqueduct/lib/context"
```
kenxu95, line 10: The PR description says this is named `context_parser` btw. Having to alias this import everywhere may be a bit annoying - what about "request_context"?

kenxu95 (approved):
One nit but generally agree with what is in the PR description.

likawind:
@kenxu95 sry I just saw your naming comments, I will submit follow up PRs to fix this

PR #15, opened 2022-06-01
Add script for building aqueduct locally

kenxu95:
Runs through all the disparate steps needed to update your local aqueduct package with the current state of the repo. Does not include UI changes yet.

### DIFF
--- a/scripts/install_local.py
+++ b/scripts/install_local.py
@@ -1,5 +1,7 @@
 """
-Installs aqueduct from the local repo. No aqueduct server can be running.
+Installs aqueduct from the local repo. Run with `python3 scripts/install_loca.py`.
+
+Requires no aqueduct server to be running.
 
 After this script completes, running `aqueduct server` will start the backend with the local changes.
 The sdk will also be updated with any local changes.
--- a/scripts/install_local.py
+++ b/scripts/install_local.py
@@ -48,6 +48,12 @@ def update_config_yaml(file):
 
 
 if __name__ == "__main__":
+    print("Current directory should be the root directory of the aqueduct repo.")
+    cwd = os.getcwd()
+    if not cwd.endswith("aqueduct"):
+        print("Your working directory is %s" % cwd)
+        exit(1)
+
     # Create the ~/.aqueduct directory is it does not already exist. Copied from `/src/python/bin/aqueduct`.
     if not os.path.isdir(server_directory):
         try:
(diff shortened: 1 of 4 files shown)


scripts/install_local.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+if __name__ == "__main__":
+    print("Current directory should be the root directory of the aqueduct repo.")
+    cwd = os.getcwd()
+    if not cwd.endswith("aqueduct"):
+        print("Your working directory is %s" % cwd)
+        exit(1)
+
+    # Create the ~/.aqueduct directory is it does not already exist. Copied from `/src/python/bin/aqueduct`.
```
kenxu95, line 57: This and some of the helpers in this script are copied from the CLI. I'm not sure if I should devote time right now trying to extract out these common helpers.

scripts/install_local.py
```diff
@@ -0,0 +1,129 @@
+"""
+Installs aqueduct from the local repo. Run with `python3 scripts/install_loca.py`.
```
cw75, line 2: `install_local.py`

cw75 (commented):
My concern with this is that if in the future we make changes to the code copied from the CLI, this will also need to be changed accordingly. What if we assume the `~/.aqueduct/server` already exists (we already performed the initialization step), then I think this script can be simplified with just lines 117-129.

Basically, before we execute this script to update stuff, we require developers to execute `pip3 install .` from `src/python` and run `aqueduct server`. This only needs to be done once. Then whenever they wanna update anything they just run `install_local.py` (maybe rename to `update_local.py`) which only contains lines 117-129.

kenxu95:
I think thats totally fair, I'll put the requirements for running this script up at the top.

scripts/install_local.py
```diff
@@ -0,0 +1,51 @@
...
+        exit(1)
+
+    # Build the local backend binaries.
+    execute_command(["make", "server"], cwd=os.path.join(cwd, "src"))
+    execute_command(["cp", "./src/build/server", os.path.join(server_directory, "bin", "server")])
+    execute_command(["make", "executor"], cwd=os.path.join(cwd, "src"))
+    execute_command(["cp", "./src/build/executor", os.path.join(server_directory, "bin", "executor")])
+
+    # Install the local python operators.
+    execute_command(["pip", "install", ".", "--user"], cwd=os.path.join(cwd, "src", "python"))
```
cw75, line 46: When I tested stuff on my end, I never use the `--user` when pip installing stuff. So now when this is executed and if I already installed it in non `--user` mode, will it still pick up the new changes?

cw75 (commented):
Left a question otherwise lgtm.

scripts/install_local.py
kenxu95: hmm yeah I think I'll get rid of the `--user` flag actually. I think the --user install gets ignored if its already installed in the system.

cw75 (approved):

PR #17, opened 2022-06-02
[SDK] Eng 1105 allow users to execute annotated

Fanjia-Yan:
This PR enable `.local()` method addition to an **annotated function**(such as @op, @metric,@check) that allows users to execute that code locally if they would like. The changes are:

- In the `wrapped()` function which is the result of adding a decorator, we add a attribute `local` and set it to a lambda function mapping from all the input `artifacts` to the undecorated, original function, taking in `artifact.get()` for each `artifacts`. The lambda function outputs a panda dataframe, which will be our `.local()` output.

How do I test it:

- I use Sentiment Model.ipynb, and run the @op `sentiment_prediction` both locally and in the cloud. `sentiment_table1` is from cloud and `sentiment_table2` is from local. The result match for several significant digits while there are some floating point inaccuracy which I am not sure if that's a result of different machine.
- Screen Shot 2022-06-02 at 8 29 16 AM
- When running locally, I encounter this warning, while haven't received anything during the execution in the cloud. I think this is a issue either with my setup or the parameter we pass in when we create it using `pipeline("sentiment-analysis") `, not our codebase issue.
Screen Shot 2022-06-02 at 8 35 58 AM


vsreekanti:
@Fanjia-Yan, nice work! If I understand correctly this allows us to call the function locally on `TableArtifact`s, but this wouldn't work if I passed in a vanilla `pd.DataFrame`, right? It would be nice if we could do some typechecking and work in both cases.

Fanjia-Yan:
> @Fanjia-Yan, nice work! If I understand correctly this allows us to call the function locally on `TableArtifact`s, but this wouldn't work if I passed in a vanilla `pd.DataFrame`, right? It would be nice if we could do some typechecking and work in both cases.

@vsreekanti , yes, you are right, currently it is only taking `TableArtifacts` as input and convert it into `pd.DataFrame.` I can make another commit to enable typechecking and make it take `pd.DataFrame` as a possible input

sdk/aqueduct/tests/decorator_local_method_test.py
```diff
@@ -0,0 +1,20 @@
...
+
+@check()
+def check_fn_with_parentheses(df):
+    pass
+print(check_fn_with_parentheses.local)
+
+@metric()
+def metric_fn_with_parentheses(df):
+    pass
+print(metric_fn_with_parentheses.local())
```
kenxu95, line 20: these aren't really tests, they just print stuff but don't check the output. Since `.local()` is a user-facing feature, I'd prefer if we wrote some integration tests for them (see `integration_tests/sdk`). Lemme know if you need help figuring out how to run these. Check out the README here: https://github.com/aqueducthq/integration-test (I need to port this over to the OSS)

Fanjia-Yan:
@kenxu95 I have done some style checks on my `decorate.py` and create `local_test.py` under the integration-tests/sdk folder. The tests check on:

> if a function decorated with @op ,@metric, or @check, will the result of local execution matches the cloud executation.
> if .local() can take both `InputArtifact` and `Dataframe` as input types.
> if .local() can process multiple inputs.

integration_tests/sdk/local_test.py
```diff
@@ -0,0 +1,51 @@
...
+    run_sentiment_model,
+    should_run_complex_models
+)
+
+def test_local_operator(sp_client):
+    db = sp_client.integration(name=get_integration_name())
+    sql_artifact = db.sql(query=SENTIMENT_SQL_QUERY)
+    output_artifact = run_sentiment_model(sql_artifact)
+    output_cloud = output_artifact.get()
+    output_local = sentiment_model.local(sql_artifact) if should_run_complex_models() else dummy_sentiment_model.local(sql_artifact)
```
kenxu95, line 21: Can we make this a function in `utils.py`? Same as `run_sentiment_model` but `run_sentiment_model_local`

integration_tests/sdk/local_test.py
```diff
@@ -0,0 +1,51 @@
...
+    assert output_cloud.count()[0] == output_local.count()[0]
+    assert (output_cloud["positivity"] == output_local["positivity"]).all()
```
kenxu95, line 23: is there a reason we can't compare the entire dataframe for equality? They should be exactly the same right?

kenxu95 (commented):
Getting there! Mostly comments around the testing :)

integration_tests/sdk/local_test.py
Fanjia-Yan: Yes I can check the entire Dataframe for equity. my initial thought is that the testing operation only add a column "positivity" to the Dataframe but it will be good to check if other columns are affected or not.

integration_tests/sdk/local_test.py
```diff
@@ -0,0 +1,51 @@
...
+    db = sp_client.integration(name=get_integration_name())
+    sql_artifact = db.sql(query=SENTIMENT_SQL_QUERY)
+    check = success_on_single_table_input
+    assert check(sql_artifact)
+    assert check.local(sql_artifact)
+
+def test_local_dataframe_input(sp_client):
+    db = sp_client.integration(name=get_integration_name())
+    sql_artifact = db.sql(query=SENTIMENT_SQL_QUERY)
+    output_local = dummy_sentiment_model.local(sql_artifact.get())
```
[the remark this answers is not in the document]
Fanjia-Yan: Thanks for the feedback! I will update my integration tests and give it another try

Fanjia-Yan:
Just push a new commit with changes:

- encapsulate the `run_sentiment_model _local` into `util.py` which simplify the code
- when checking whether the local execution and cloud execution is the same, I use `.equals()` instead of comparing the added column
- add "Attribute section" under "Returns section" in `decorate.py` to describe the usage of `.local `method

integration_tests/sdk/local_test.py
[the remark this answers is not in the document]
kenxu95: any reason this isn't comparing `run_sentiment_model()` and `run_sentiment_model_local()`?

sdk/aqueduct/decorator.py
```diff
@@ -142,7 +143,17 @@ def op(
...
+    
+    Returns:
+        A decorated function that once called with artifact input(s) will be
+        executed in the cloud and returns a TableArtifact.
+
+            Attributes:
+                local:
+                    A function that once called will be executed locally that conduct
+                    the same operations as the decorated function. The function takes in
+                    either TableArtifact or Dataframe, and return a Dataframe.
```
kenxu95, line 155: I feel like this description might be a bit hard for users to understand (and the cloud part is not necessarily correct, the user could be running it all locally). Also, we seem to already describe the "return" part in the first few lines of the docstring.

What about just something like (get rid of the headers and just stick this above "Args" on line 137:

```
To run the wrapped code locally, without Aqueduct, use the `local` attribute. Eg:
>>> compute_recommendations.local(customer_profiles, recent_clicks)
```

kenxu95 (approved):
Cool, just one more request around the docstring, which is a little wordy and can be more direct.

PR #19, opened 2022-06-02
[4/4] Reorg handler codes

likawind:
This PR adds slightly more clean up to `server` code base by separating out all single-handler codes from server's handler execution codes. This makes it easier to look for handler implementations vs server implementations.

## Tests
Compiles


kenxu95 (approved):

cw75 (approved):
LGTM!

PR #23, opened 2022-06-03
ENG-1193 Fix Hooks Errors

agiron123:
Resolves:
https://linear.app/aqueducthq/issue/ENG-1193/fix-bug-where-workflowid-page-does-not-show-and-crashes-app

Fixes bug where the worfklow/[id] page crashed the app. 
Also fixes console.log error where we were calling hooks out of order.

### DIFF
--- a/src/ui/common/src/components/workflows/ReactFlowCanvas.tsx
+++ b/src/ui/common/src/components/workflows/ReactFlowCanvas.tsx
@@ -33,6 +33,7 @@ const ReactFlowCanvas: React.FC<ReactFlowCanvasProps> = ({
     (state: RootState) => state.openSideSheetReducer
   );
   const { fitView } = useReactFlow();
+  useEffect(() => { fitView() }, []);
 
   useEffect(() => {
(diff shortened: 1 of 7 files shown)


src/ui/common/src/components/workflows/ReactFlowCanvas.tsx
```diff
@@ -33,6 +33,7 @@ const ReactFlowCanvas: React.FC<ReactFlowCanvasProps> = ({
     (state: RootState) => state.openSideSheetReducer
   );
   const { fitView } = useReactFlow();
+  useEffect(() => { fitView() }, []);
```
agiron123, line 36: TODO: Will need to bump version in package.json and re-publish before merging this PR.

vsreekanti (approved):

src/ui/common/src/components/workflows/ReactFlowCanvas.tsx
agiron123: NOTE: Will bump after #42 is merged so we can just do it all in one go.

PR #24, opened 2022-06-03
Pipe executor logs to the logs directory

kenxu95:
As we discussed, any jobs being run by the executor should be piped to a file in the `server/logs` directory. We do this by setting the output of the `logrus` library. The naming rules are currently:
- `logs/default`: all logs that were generated in the brief period of time before we are able to redirect logs to the correct file are here.
- `logs/<job name>`: all logs that are generated by the workflow job. Includes all the function/extract/etc logs.
- `logs/Workflow Job Retention`: all logs generated by the long-running workflow job retention job.

There are two bugs I found related to the returned execution status of a job in poll() being SUCCESS when it should have been FAIL. This seems like issues since we rely on poll() within our orchestration engine.
https://linear.app/aqueducthq/issue/ENG-1196/processjobmanagerpoll-returns-successful-execution-status-for
https://linear.app/aqueducthq/issue/ENG-1195/processjobmanagerpoll-returns-successful-execution-status-for

BELOW IS NO LONGER DIRECTLY RELATED TO THE CONTENTS OF THIS PR -------------------------------------------------------------
 
I spent the afternoon poking at the server with bad functions and trying to figure out how to get them to display in the server logs. The solution in this PR might not be the way we want to structure things long term, but it prints out all our operator errors to the console.

In order for us to be able to print the logs of a job launched by the process manager, we need to poll on it. This can be easily done to the cron job function. However, refresh_workflow currently only launches the job and then immediately returns. How will we be able to grab the logs if we don't poll? I've added a comment at the place I'd like to discuss.

### DIFF
--- a/src/golang/lib/job/process.go
+++ b/src/golang/lib/job/process.go
@@ -30,11 +30,13 @@ const (
 
 	BinaryDir          = "bin/"
 	OperatorStorageDir = "storage/operators/"
+	LogsDir            = "logs/"
 )
 
 var (
 	defaultBinaryDir          = path.Join(os.Getenv("HOME"), ".aqueduct", BinaryDir)
 	defaultOperatorStorageDir = path.Join(os.Getenv("HOME"), ".aqueduct", OperatorStorageDir)
+	defaultLogsDir            = path.Join(os.Getenv("HOME"), ".aqueduct", "server", LogsDir)
 )
 
 type Command struct {
@@ -128,7 +130,7 @@ func NewProcessJobManager(conf *ProcessConfig) (*ProcessJobManager, error) {
 	}, nil
 }
 
-func (j *ProcessJobManager) mapJobTypeToCmd(spec Spec) (*exec.Cmd, error) {
+func (j *ProcessJobManager) mapJobTypeToCmd(jobName string, spec Spec) (*exec.Cmd, error) {
 	if spec.Type() == WorkflowJobType {
 		workflowSpec, ok := spec.(*WorkflowSpec)
 		if !ok {
@@ -140,10 +142,15 @@ func (j *ProcessJobManager) mapJobTypeToCmd(spec Spec) (*exec.Cmd, error) {
 			return nil, err
 		}
 
+		logFilePath := path.Join(defaultLogsDir, jobName)
+		log.Infof("Logs for job %s are stored in %s", jobName, logFilePath)
+
 		return exec.Command(
 			fmt.Sprintf("%s/%s", j.conf.BinaryDir, workflowExecutorBinary),
 			"--spec",
 			specStr,
+			"--logs-path",
+			logFilePath,
 		), nil
 	} else if spec.Type() == FunctionJobType {
 		functionSpec, ok := spec.(*FunctionSpec)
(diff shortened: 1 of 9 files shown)


src/golang/cmd/server/server/refresh_workflow.go
```diff
@@ -108,5 +109,13 @@ func (h *RefreshWorkflowHandler) Perform(ctx context.Context, interfaceArgs inte
...
 		return nil, http.StatusInternalServerError, errors.Wrap(err, "Unable to trigger this workflow.")
 	}
+
+	// TODO: debate this in the PR
+	status, err := job.PollJob(ctx, jobName, h.JobManager, 2*time.Second, 5*time.Minute)
+	if err != nil {
+		return nil, http.StatusInternalServerError, errors.Wrap(err, "Unable to poll this workflow.")
+	}
+	fmt.Printf("Cron job completed with status %s\n", status)
+
```
kenxu95, line 119: I don't think polling on the job here is the long term solution. refresh_workflow is called by register_workflow, and it seems silly to block `publish_flow()` until an entire flow run completes.

Any ideas? What if we had a long-running goroutine on every process manager that listens for certain jobs to complete, and flushes the stdout and stderr of those jobs to the console + deletes the cmd? We wouldn't have to poll() after every launch() to get the logs then, we could just launch() and then register the jobname with this background thread.

kenxu95: Actually yeah I think you're right! I guess I didn't retest correctly after making all the changes.

src/golang/cmd/executor/executor/log.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"github.com/aqueducthq/aqueduct/lib/job"
+	"github.com/dropbox/godropbox/errors"
+	log "github.com/sirupsen/logrus"
+	"os"
+	"path"
+)
+
+var LogsDir = path.Join(os.Getenv("HOME"), ".aqueduct", "server", "logs")
```
likawind, line 11: what's the behavior for enterprise version?

src/golang/cmd/executor/executor/log.go
```diff
@@ -0,0 +1,33 @@
+package executor
```
likawind, line 1: I'm wondering if the log file location will be shared across the stack, outside of executor. In that case I would consider moving it to `/lib`

src/golang/cmd/executor/executor/log.go
likawind: As a quick fix maybe we can have a command line flag to specify the redirected log file. This could be empty in enterprise version, in which case we will still print in stdout / stderr

src/golang/lib/job/process.go
```diff
@@ -98,6 +98,8 @@ func (j *ProcessJobManager) mapJobTypeToCmd(spec Spec) (*exec.Cmd, error) {
 			fmt.Sprintf("%s/%s", j.conf.BinaryDir, workflowExecutorBinary),
 			"--spec",
 			specStr,
+			"--job-name",
```
likawind, line 101: Is this because of the deserialization bug?

likawind (changes requested):
The OSS changes looks great! I'm blocking mainly to confirm:
* If we should print out the path in server log rather than asking user to follow the agreed format to figure out the file location
* make sure we have clear plan for how to proceed with enterprise version

src/golang/lib/job/process.go
kenxu95: Nope its because refresh and register workflow both set the "name" of the workflow spec to the workflow object name, which isn't the job name (edit workflow sets it as the job name though). We can decide the make the "name" of a workflow spec consistent and use that, but that runs into your aforementioned deserialization bug.

likawind: I see, it makes sense to go with the current solution then

src/golang/cmd/executor/executor/log.go
kenxu95: Moved the remaining helper to main.py. I think we can refactor this to /lib if we find another use case.

kenxu95:
Logged the path that the logs are at. I have a question about workflow retention, but I'll ask it on slack.

likawind (approved):
Looks great! Thanks for addressing comments!

PR #25, opened 2022-06-03
Add support to let the user configure whether to expose the server to the public

cw75:
By default, this is set to False. But if the user wants to expose the port to the public, she can do `aqueduct server --expose`.

### DIFF
--- a/src/golang/cmd/server/server/aqueduct_server.go
+++ b/src/golang/cmd/server/server/aqueduct_server.go
@@ -230,7 +230,12 @@ func (s *AqServer) Log(ctx context.Context, key string, req *http.Request, statu
 	logging.LogRoute(ctx, key, req, excludedHeaderFields, statusCode, utils.Server, s.Name, err)
 }
 
-func (s *AqServer) Run() {
+func (s *AqServer) Run(expose bool) {
+	ip := ""
+	if !expose {
+		ip = "localhost"
+	}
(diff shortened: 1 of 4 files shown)


src/golang/cmd/server/server/aqueduct_server.go
```diff
@@ -230,7 +230,12 @@ func (s *AqServer) Log(ctx context.Context, key string, req *http.Request, statu
...
-func (s *AqServer) Run() {
+func (s *AqServer) Run(expose bool) {
+	ip := ""
+	if !expose {
+		ip = "localhost"
+	}
+
 	log.Infof("%s Starting HTTP server on port %d\n", time.Now().Format("2006-01-02 03:04:05 PM"), connection.ServerInternalPort)
-	log.Fatal(http.ListenAndServe(fmt.Sprintf(":%d", connection.ServerInternalPort), s.Router))
+	log.Fatal(http.ListenAndServe(fmt.Sprintf("%s:%d", ip, connection.ServerInternalPort), s.Router))
```
saurav-c, line 240: So if `ip = ""` the server will be exposed?

cw75: Yeah, I can add a comment clarifying this.

saurav-c (approved):
LGTM!

PR #27, opened 2022-06-03
Sets up tooling to create single executable UI package

vsreekanti:
This PR adds support for using the [pkg](https://github.com/vercel/pkg) library to generate a single executable for the UI. In order to do that, we add our own custom entrypoint (`server.js`) which is a very thin wrapper around the default `next` server. 

To generate the package, we can run `pkg --compress -t node16-linux-x64,node16-macos-x64,node16-macos-arm64 . --out-path pkg`, which will generate the 3 executables we care about in the `pkg` directory. 

## Tests
Works on an Ubuntu machine. Still needs to be validated in other environments.


src/ui/common/src/components/workflows/ReactFlowCanvas.tsx
```diff
@@ -42,7 +42,7 @@ const ReactFlowCanvas: React.FC<ReactFlowCanvasProps> = ({
     setTimeout(fitView, 200);
   }, [openSideSheetState]);
 
-  useEffect(fitView);
+  useEffect(fitView, []);
```
agiron123, line 45: Should be resolved in https://github.com/aqueducthq/aqueduct/pull/23 can you update here to match?

src/ui/app/server.js
```diff
@@ -0,0 +1,31 @@
+require('dotenv').config()
```
agiron123, line 1: Any changes here that needed to happen to properly handle cookies? Or was I just hitting issues with env variables not being properly set?

agiron123:
Could you update our makefile to add a package action, and add that to a .github action as well please?

agiron123 (approved):
Looks good! Excited to see this working.

Will pull down shortly to test out on my M1 mac.

src/ui/app/server.js
vsreekanti: I think it was the env variables issue. Once I fixed this, everything worked as expected for me.

PR #30, opened 2022-06-06
UI-1035-reduce text in notification

Boyuan-Deng:
### This PR:
- clean up the messages displayed in the notification list item card

#### Before Clean Up:
image

#### After Clean Up:
image

### DIFF
--- a/src/ui/common/src/components/notifications/NotificationListItem.tsx
+++ b/src/ui/common/src/components/notifications/NotificationListItem.tsx
@@ -54,7 +54,6 @@ export const NotificationListItem: React.FC<Props> = ({
               variant="body1"
               gutterBottom
               sx={{
-                fontWeight: 'light',
                 fontFamily: 'Monospace',
                 '&:hover': { textDecoration: 'underline' },
               }}
--- a/src/ui/common/src/components/notifications/NotificationListItem.tsx
+++ b/src/ui/common/src/components/notifications/NotificationListItem.tsx
@@ -54,16 +54,13 @@ export const NotificationListItem: React.FC<Props> = ({
               variant="body1"
               gutterBottom
               sx={{
+                fontWeight: 'medium',
                 fontFamily: 'Monospace',
                 '&:hover': { textDecoration: 'underline' },
               }}
             >
               {notification.workflowMetadata.name}
             </Typography>
-
-            {notification.level === NotificationLogLevel.Success
-              ? ' succeeded'
-              : ' failed'}
           </Typography>
         </Box>
       ) : (
@@ -97,24 +94,13 @@ export const NotificationListItem: React.FC<Props> = ({
           icon={faXmark}
           style={{
             cursor: 'pointer',
-            color: 'gray.600',
+            color: 'gray600',
           }}
           onClick={() =>
             dispatch(handleArchiveNotification({ user, id: notification.id }))
           }
         />
       </Box>
-
-      <Typography
-        variant="h6"
-        gutterBottom
-        sx={{
-          fontFamily: 'Monospace',
-          '&:hover': { textDecoration: 'underline' },
-        }}
-      >
-        {notification.content}
-      </Typography>
 
       <Box
         sx={{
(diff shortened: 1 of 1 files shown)


Boyuan-Deng:
Design Decisions
- delete notification.content & notification type based on [linear discussions](https://linear.app/aqueducthq/issue/ENG-1035/reduce-text-in-notification)
- changed some font to make it cleaner

src/ui/common/src/components/notifications/NotificationListItem.tsx
```diff
@@ -97,25 +94,14 @@ export const NotificationListItem: React.FC<Props> = ({
           icon={faXmark}
           style={{
             cursor: 'pointer',
-            color: 'gray.600',
+            color: 'gray600',
```
vsreekanti, line 97: The MUI theme uses `gray.600` as the style. If you look at the X icon in the screenshot you provided, the styling isn't being reflected currently.

vsreekanti (approved):
@Boyuan-Deng, this looks good to me, but can you verify the styling is working as expected. I think MUI requires the `.` in the color palette. cc @agiron123

agiron123:
Hmm, what happened to the sentence "Workflow save_workflow_2 has succeeded!" 
I was hoping final result would look something like this, without the word "workflow" on the bottom left.
image

src/ui/common/src/components/notifications/NotificationListItem.tsx
agiron123: Yes, it requires a period to separate the color and it's weight.
so ```gray.600``` would be the correct way to do this.

In the future, we should add some constants to have a safer way to code these. Missing a period is pretty easy to do and not something we can detect easily via our tools, so some consts will help. I'll file a linear task for this.

src/ui/common/src/components/notifications/NotificationListItem.tsx
```diff
@@ -126,18 +112,25 @@ export const NotificationListItem: React.FC<Props> = ({
         <Box>
           <Typography
             variant="body1"
-            sx={{ fontWeight: 'medium', color: 'gray600' }}
+            sx={{
+              fontWeight: 'light',
+              color: 'gray600',
```
agiron123, line 117: Can we change the weight back to medium here?

Should be gray.600

Also, the comment below in line 121 no longer reflects what is happening. Can you fix that up too?

src/ui/common/src/components/notifications/NotificationListItem.tsx
```diff
@@ -54,16 +54,13 @@ export const NotificationListItem: React.FC<Props> = ({
               variant="body1"
               gutterBottom
               sx={{
+                fontWeight: 'medium',
```
agiron123, line 57: Can we put the fontweight back to what it was before please?

src/ui/common/src/components/notifications/NotificationListItem.tsx
```diff
@@ -126,18 +112,25 @@ export const NotificationListItem: React.FC<Props> = ({
...
+              : ' Failure'}
           </Typography>
         </Box>
         <Box>
           <Typography
             variant="body1"
-            sx={{ fontWeight: 'light', color: 'gray600' }}
+            sx={{
+              fontWeight: 'light',
+              color: 'gray600',
```
agiron123, line 132: Also gray.600

agiron123 (changes requested):
See comments, couple changes before this is 100% ready.

src/ui/common/src/components/notifications/NotificationListItem.tsx
```diff
@@ -105,43 +110,21 @@ export const NotificationListItem: React.FC<Props> = ({
...
-            variant="body1"
-            sx={{ fontWeight: 'light', color: 'gray600' }}
-          >
-            {`${dateString(notification.createdAt)}`}
-          </Typography>
-        </Box>
+        <Typography
+          variant="body1"
+          sx={{
+            fontWeight: 'light',
```
agiron123, line 122: Can you post a screenshot of light vs medium font weight please? Curious what the two look like.

agiron123 (approved):
Couple comments, once you respond we should be good to go.

Boyuan-Deng:
> @Boyuan-Deng Can you post a screenshot of final result?

See the screen shot above inside the PR content. After Clean up is the final look.

src/ui/common/src/components/notifications/NotificationListItem.tsx
Boyuan-Deng: image

agiron123:
> > @Boyuan-Deng Can you post a screenshot of final result?
> 
> See the screen shot above inside the PR content. After Clean up is the final look.

@Boyuan-Deng Nice work. Looks great. 🚢 it!

PR #36, opened 2022-06-06
ENG-1109 Export all from index.ts

agiron123:
Part of: https://linear.app/aqueducthq/issue/ENG-1109/refactor-ui-enterprise-to-import-from-ui-components-package

This PR makes it so that we're now exporting everything that is exported inside of the ui/common module from a single index.ts file. 

As a result, our code now runs through the Typescript compiler and gets output to a build bundle.

This means we no longer need to transpile modules in next.js, and can now import from the root of our package like so: @aqueducthq/common/foobar


src/ui/common/src/components/primitives/LoadingButton.styles.tsx
```diff
@@ -2,7 +2,9 @@ import LoadingButton, { loadingButtonClasses } from '@mui/lab/LoadingButton';
 import { buttonClasses } from '@mui/material/Button';
 import { styled } from '@mui/material/styles';
 
-const AqueductLoadingButton = styled(LoadingButton)(({ theme }) => {
+import { theme } from '../../styles/theme/theme';
```
vsreekanti, line 5: Not sure if this is intentional, but `theme` is actually passed in by default by MUI. 

I think the idea is that if you wanted to have multiple nested themes rather than just 1, this would inherit the theme from the closest location.

vsreekanti (approved):
Just one tiny question. LGTM otherwise!

src/ui/common/src/components/primitives/LoadingButton.styles.tsx
agiron123: I actually got some compiler errors here and was having trouble getting this to work with common. I'm not really sure how we can go about typing it, but open to suggestions.

agiron123:
This PR's work has been superseded by #42

PR #38, opened 2022-06-06
Add compile option to remove symbol tables, debugging info, and path info from the Go binary

cw75:
The `-lean` Make option does the trick. We should use this flag to compile the binaries and upload them to S3.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -21,20 +21,22 @@ server:
 	mkdir -p build
 	cp $$GOPATH/bin/server build/server
 
-migrator-lean:
+migrator-release:
 	cd golang/cmd/migrator && go install -ldflags="-s -w" -trimpath
 	mkdir -p build # Make the build directory if it doesn't exist
 	mv $$GOPATH/bin/migrator build/migrator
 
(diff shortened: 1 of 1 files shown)


src/Makefile
```diff
@@ -21,6 +21,21 @@ server:
 	mkdir -p build
 	cp $$GOPATH/bin/server build/server
 
+migrator-lean:
```
vsreekanti, line 24: Tiny, tiny naming nit -- the `lean` flag is an implementation detail. Can we call this `make migrator-release` and then add a target called `release` which is just a dummy target for all 3 of these?

vsreekanti (approved):

vsreekanti:
@cw75, out of curiosity, I don't see `-lean` anywhere in the codebase. How is this specified?

cw75:
@vsreekanti Oh I was planning to run `make release` manually to compile and then upload to S3. So there's no where else in the repo that uses this flag. Do you have anything else in mind?

vsreekanti:
Nope, just curious. This LGTM!

PR #39, opened 2022-06-06
Fix deprecated cli command in the integration test

cw75:


kenxu95:
ci

kenxu95 (approved):
I triggered an integration test run by commenting "ci" on this pull request, once that passes you can merge :)

kenxu95:
Actually Fanjia's latest PR broke the tests. If the tests get past the server spinup part, we should be good.

cw75:
@kenxu95 I don't think the integration test is picking up my change though.

kenxu95:
urgh I hate hate hate github actions. Its impossible to test anything well. The comment-trigger seems to be broken.

Whatever, just merge it. Can you add the `integration-test.yml` file itself to the paths that would trigger a run on commit btw?

.github/workflows/integration-tests.yml
```diff
@@ -69,7 +69,7 @@ jobs:
         run: pip3 install .
```
kenxu95, line 69: Can we add `.github/workflows/integration-test.yml` to the `paths` field at the beginning of this test

cw75: like this? see my last commit.

kenxu95: should be "integration-tests.yml"

PR #41, opened 2022-06-07
Improve thread-safety of job manager

likawind:
This PR improves thread-safety of process based job manager as the same instance is being used in multiple go routines (e.g. gateway handlers). We added read / write locks to map manipulation so that concurrent IO doesn't crash the server. This change is not necessary for K8s manager where we rely on k8s to manage the concurrency.

I think the current solution is better than a few other options:
* Technically we could require job manager to be only used in single-thread. This is obviously hard to enforce and is very risky as a process.
* There are publicly available concurrent map in go: https://github.com/orcaman/concurrent-map . Given the repo is maintained by rather small group of ppl and that our use case is small, I don't think we should use it in a critical piece in our backend.

Resolves https://linear.app/aqueducthq/issue/ENG-1210/investigate-server-crash-due-to-concurrency

## Tests:
integration test passed with concurrency: https://github.com/aqueducthq/aqueduct/runs/6765132149?check_suite_focus=true


kenxu95 (approved):
Makes sense to me, but might be good to get @cw75's opinion too

cw75 (changes requested):
@likawind Looks good to me! One nit is that I think we can use two mutex: one for `cmds` and the other one for `cronMapping` to enable a bit more concurrency. These two maps are independent, so we don't have to block access to `cronMapping` when we are updating `cmds`, for example.

cw75 (approved):

PR #46, opened 2022-06-07
Eng 1105 allow users to execute annotated REVISED

Fanjia-Yan:
The last PR #17  fails the integration test and is reverted since I did not recompile the sdk locally and run the test(my bad :(  ). Here is the update I make to fix the issue and pass the integration test:

1.` input.get() if type(input) in InputArtifact_list` instead of ` input.get() if type(input) in InputArtifact` since `InputArtifact` is a Union of three types, using `in` does not necessarily test for whether the input is a TableArtifact. This results in passing TableArtifact into the locally executed function. Therefore, I create a list of InputArtifact `InputArtifact_list` instead of Union so the type checking works 
2. `run_sentiment_model_local(artifact: TableArtifact) -> DataFrame` instead of `run_sentiment_model_local(artifact: TableArtifact) -> TableArtifact` because `.local(*arg)` method will return a DataFrame

The current commit passes all the integration tests with the most up-to-date sdk installed.


sdk/aqueduct/decorator.py
```diff
@@ -18,10 +18,13 @@
     artifact_name_from_op_name,
 )
 from aqueduct.error import AqueductError
+from pandas import DataFrame
 
 # Valid inputs and outputs to our operators.
 OutputArtifact = Union[TableArtifact, MetricArtifact, CheckArtifact]
 InputArtifact = Union[TableArtifact, MetricArtifact, ParamArtifact]
+InputArtifact_list = [TableArtifact, MetricArtifact, ParamArtifact]
+InputArtifact_local = Union[TableArtifact, MetricArtifact, ParamArtifact,DataFrame]
```
kenxu95, line 27: These variables are a mix of camelcase and underscores. Lets just do camelcase for types in python. Underscores for everything else in python.

kenxu95: Also, `it needs to be `InputArtifactList = List[TableArtifact, MetricArtifact...]`. Mypy would catch this once I get that in.

kenxu95 (changes requested):
Cool, thanks for testing it! Just ping me against once the style and typing things are fixed :)

sdk/aqueduct/decorator.py
Fanjia-Yan: Just fix the issues! Since "a list is a homogenous structure containing values of one type", I can't necessarily use a List to wrap three types. Therefore, I turn to a solution on StackOverflow. Let me know if you think that works or not.
https://stackoverflow.com/questions/45957615/check-a-variable-against-union-type-at-runtime-in-python-3-6

sdk/aqueduct/decorator.py
```diff
@@ -18,10 +18,12 @@
     artifact_name_from_op_name,
 )
 from aqueduct.error import AqueductError
+from pandas import DataFrame
 
 # Valid inputs and outputs to our operators.
 OutputArtifact = Union[TableArtifact, MetricArtifact, CheckArtifact]
 InputArtifact = Union[TableArtifact, MetricArtifact, ParamArtifact]
+InputArtifactLocal = Union[TableArtifact, MetricArtifact, ParamArtifact,DataFrame]
```
kenxu95, line 26: nit: space in front of "DataFrame"

kenxu95 (approved):
Thanks! Yeah I realize now I meant Union not List

PR #60, opened 2022-06-08
Eng 944 update discoverroute to show only base

eunice-chan:
Before, the Discover route (and Integration Details Page) shows all tables in the integration. Now, it only shows the base tables--user-created tables (loadspec) are filtered out.


src/golang/lib/workflow/operator/connector/load_params.go
```diff
@@ -39,6 +39,50 @@ type S3LoadParams struct {
 	Format   string `json:"format"`
 }
 
+func CastToRelationalDBLoadParams(params LoadParams) (*RelationalDBLoadParams, bool) {
```
cw75, line 42: If all we care about is whether the load param is a relational db load param, can we just cast it to `RelationalDBLoadParams` instead of trying to cast to the specific ones?

src/ui/common/package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "@aqueducthq/common",
-  "version": "0.0.6",
+  "version": "0.0.7",
```
cw75, line 3: Is this change intentional? Seem not relevant to this PR.

cw75 (commented):
LGTM, a couple questions inline.

src/golang/lib/workflow/operator/connector/load_params.go
likawind: Is this comment out-dated? It seems that this is exactly what the func is doing

likawind (approved):
The change LGTM!

@cw75 and @vsreekanti one question for you guys is what do we show if the user choose to append to an existing 'base' table? The current behavior will filter that table out, but I don't have good idea on what's a better way though

src/golang/lib/workflow/operator/connector/load_params.go
eunice-chan: I can't do that directly because there is no generic RelationalDBLoadParamsWrapper and there was some weirdness when I was trying to make it a LoadParam. For now, I've decided to do something similar to https://github.com/aqueducthq/aqueduct/blob/377cd4d603b8160d1c647bb1149b63f8b3c0039c/src/golang/lib/workflow/operator/connector/extract_params.go#L70

eunice-chan:
New:
Only CSV-uploaded & existing base tables.
image
Old:
The above plus the data tables saved by the user via `(table_artifact).save`
image

PR #62, opened 2022-06-08
Add Mypy check Github Action on the SDK

kenxu95:
In the process, refactored the `.local()` attribute a bit to conform better to mypy (and also be more readable).


Fanjia-Yan (approved):
Looks good to me!

PR #64, opened 2022-06-08
Port UI-Build Github Action from Enterprise

kenxu95:
Had to massage this to keep it consistent with the new folder structure. Don't know if I'm missing anything tho

### DIFF
--- a/.github/workflows/ui-build.yml
+++ b/.github/workflows/ui-build.yml
@@ -18,9 +18,9 @@ jobs:
         with:
           node-version: "16"
 
-#      - name: Install dependencies
-#        working-directory: ./src/ui/app
-#        run: npm install
+      - name: Install dependencies
+        working-directory: ./src/ui/app
+        run: npm install
 
       - name: Format check
         working-directory: ./src/ui/app
--- a/.github/workflows/ui-build.yml
+++ b/.github/workflows/ui-build.yml
@@ -24,6 +24,6 @@ jobs:
         working-directory: ./src/ui/app
         run: npm run lint
 
-      - name: Build
+      - name: Link and Build
         working-directory: ./src/ui
-        run: make build
+        run: make link-dev && make dist
--- a/.github/workflows/ui-build.yml
+++ b/.github/workflows/ui-build.yml
@@ -17,6 +17,10 @@ jobs:
         with:
           node-version: "16"
 
+      - name: Link and Build
+        working-directory: ./src/ui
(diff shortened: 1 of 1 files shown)


eunice-chan:
Would this pick up changes in the `common` folder? Right now the way I am working on things in ui is running this set of commands before testing my changes:
```
cd common
npm run lint:fix
rm -rf node_modules
npm install
npm run build
sudo npm link
cd ../app
npm run lint:fix
rm -rf node_modules
npm link @aqueducthq/common
npm run build
npm run start
```

kenxu95:
I don't know much about this - maybe @vsreekanti or @agiron123 (when we gets back) can give us some guidance on whether common needs to be built and linked before app?

vsreekanti:
Eunice is right that you need to build and link common locally before it'll pick up the changes. Without that, it'll pull the version of common that is on npm. Some of the steps are unnecessary, but here's what's required: 

```cd common
npm install
npm run build
sudo npm link
cd ../app
npm install
npm link @aqueducthq/common
npm run start # or npm run build if you're generating the compiled version of the code
```

.github/workflows/ui-build.yml
```diff
@@ -0,0 +1,29 @@
...
+
+      - name: Install dependencies
+        working-directory: ./src/ui/app
+        run: npm install
+
+      - name: Format check
+        working-directory: ./src/ui/app
+        run: npm run lint
+
+      - name: Build
```
vsreekanti, line 27: Rather than this, let's do `make link-dev` and then run `make dist`. That should cover everything we need to do.

kenxu95: those commands don't call `npm install` in the `app/` directory though. Was that an oversight or is that ok?

vsreekanti: `npm install` is actually unnecessary. If `npm run build` in the `app` directory succeeds, that means that the code has compiled and built into a bundle that we can use.

vsreekanti: Woah. That's super weird. It's not installing force-resolutions, but it's still working.

cw75:
@kenxu95 what's the status of this PR?

kenxu95:
I gave it to Andre since he had much better context: https://linear.app/aqueducthq/issue/ENG-1185/add-automated-ui-build-github-action-to-the-public-repo

vsreekanti:
This can be closed -- #132 resolves this task.

PR #65, opened 2022-06-08
Port Unit Test Github Actions over

kenxu95:
For both the backend and the sdk.

### DIFF
--- a/sdk/aqueduct/tests/decorator_test.py
+++ b/sdk/aqueduct/tests/decorator_test.py
@@ -24,7 +24,6 @@ def test_packaging_files():
 
         assert not os.path.exists("./test_function/test_function.py")
         assert os.path.exists("./test_function/test_dependency_folder/helper_function.py")
-        assert os.path.exists("./test_function/requirements.txt")
 
     finally:
         delete_zip_folder_and_file("test_function")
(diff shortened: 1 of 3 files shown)


sdk/aqueduct/tests/decorator_test.py
```diff
@@ -24,7 +24,6 @@ def test_packaging_files():
 
         assert not os.path.exists("./test_function/test_function.py")
         assert os.path.exists("./test_function/test_dependency_folder/helper_function.py")
-        assert os.path.exists("./test_function/requirements.txt")
```
eunice-chan, line 27: Out of the loop on this one--why do we no longer have `requirements.txt`? Is this part of this PR?

kenxu95: oh yeah at lot of the context is here: https://linear.app/aqueducthq/issue/ENG-1191/remove-pipreqsnb-for-requirements-discovery

Essentially, we wanted to cut out pipreqsnb because of its weirdness. We will still copy requirements.txt into the zip file if it exists in the same directory as the function - we just won't infer.

likawind (approved):
Can't eyeball anything wrong. Any chance we could verify on this branch? Or we need to verify in production?

sl43-1553109213609

kenxu95:
yep, just tested.

PR #68, opened 2022-06-08
Adds RequireAuth component to project.

agiron123:
Adds component to check if the user is authed.


src/ui/common/src/components/pages/LoginPage.tsx
```diff
@@ -14,8 +14,13 @@ export const LoginPage: React.FC = () => {
   const [apiKey, setApiKey] = useState<string>(
     cookies.load('aqueduct-api-key')
   );
+
   const navigate = useNavigate();
 
+  const [isAuthed, setIsAuthed] = useState<boolean>(false);
+  // used to trigger a re-render of the page after auth has gone through.
+  const [toggle, setToggle] = useState<boolean>(false);
```
agiron123, line 22: Probs don't need this anymore. Can't quite figure out why this works only one time around and why removing cookies and trying again doesn't work.

Removing the cookie does bounce me back here though, which is good.

src/ui/common/src/components/pages/LoginPage.tsx
```diff
@@ -36,18 +41,25 @@ export const LoginPage: React.FC = () => {
...
       setErrorMsg(
         "Invalid key, please copy the key from 'aqueduct apikey' outputs."
       );
+      setIsAuthed(false);
     } else {
       setValidationError(false);
       setUser(apiKey);
-      navigate('/');
+      setIsAuthed(true);
+      navigate('/', { replace: true })
```
agiron123, line 55: This and below are pretty much the same, but was struggling with getting things to work so tried hooks and component way of doing this.

PR #69, opened 2022-06-09
Fix hardcoded path

cw75:
@vsreekanti It turned out to be as simple as adding `server_directory` to the `cmd` argument to popen. Now I'm assuming the ui binaries are in `~/.aqueduct/server/bin/ui`, but you can adjust if you want to. Also I removed the UI-related code from the CLI, but we still need to add the part that download the ui from S3 to `server()`


vsreekanti:
Awesome! Should we just do `~/.aqueduct/ui` for the location of the files? I think your original set up was good.

Shouldn't the download still just be downloading a Zip?

cw75:
@vsreekanti The reason I put everything under `~/.aqueduct/server` is that I thought since we're hosting both the UI and the backend with the same server, they can just go together into the `server` directory. Also previously we were creating two `__version__` files under `~/.aqueduct/ui` and `~/.aqueduct/server`, respectively, and now we only need to do it in one place, which is cleaner and simpler. I agree that the download logic should still be downloading a zip file and unpacking it to whatever directory we end up deciding.

src/python/bin/aqueduct
```diff
@@ -142,13 +128,21 @@ def update_server_version():
 
     execute_command([os.path.join(server_directory, "bin", "migrator"), "--type", "sqlite", "goto", "8"])
 
-def generate_welcome_message(expose_ip):
-    if not expose_ip:
+def generate_welcome_message(expose):
+    if not expose:
         expose_ip = "localhost"
+    else:
+        expose_ip = "<PUBLIC_IP>"
```
vsreekanti, line 135: What's going on here?

vsreekanti (approved):

src/python/bin/aqueduct
cw75: @vsreekanti Since we don't require the user to input the ip anymore, we don't know the public ip right? So I'm putting a placeholder here. Any better ideas?

PR #70, opened 2022-06-09
Eng-1218 Updated job.launch to add workflow retention job

hsubbaraj-spiral:
Verified that the workflow retention job is successfully run: 
```
INFO[0328] Running cron job workflowretentionjob-1654783200
INFO[0328] Successfully ran cron job workflowretentionjob-1654783200
```


src/golang/lib/job/process.go
```diff
@@ -23,7 +23,7 @@ const (
 	defaultPythonExecutorPackage = "aqueduct_executor"
 	connectorPythonPath          = "operators.connectors.tabular.main"
 	paramPythonPath              = "operators.param_executor.main"
-	workflowExecutorBinary       = "executor"
+	exectuorBinary               = "executor"
```
kenxu95, line 26: spelling?

src/golang/lib/job/process.go
```diff
@@ -146,7 +146,28 @@ func (j *ProcessJobManager) mapJobTypeToCmd(jobName string, spec Spec) (*exec.Cm
...
+		specStr, err := EncodeSpec(workflowRetentionSpec, GobSerializationType)
+		if err != nil {
+			return nil, err
+		}
+
+		logFilePath := path.Join(defaultLogsDir, jobName)
+		log.Infof("Logs for job %s are stored in %s", jobName, logFilePath)
+
+		return exec.Command(
+			fmt.Sprintf("%s/%s", j.conf.BinaryDir, exectuorBinary),
```
kenxu95, line 170: this is good, can we also move the python connector case out of the `else` at the end here into an `else if`? Lets error if anything falls through to the `else`. This will force us to modify this function appropriately when any new job types are introduced.

kenxu95 (approved):
unblocking! But have some comments to address

PR #72, opened 2022-06-09
Eng 1204 add sdk command to get local host apikey

Fanjia-Yan:
This PR seeks to add SDK command .apikey() so that the user is able to run: `api_key = aqueduct.apikey()`to get the local api key.

- The implementation follows `src/python/bin/aqueduct.py` `get_apikey()` which enables the user to see the API key when running `aqueduct server` I think.
- I wrote one test for this apikey() where it tests whether the established client's API key matches the result from running the SDK command. I put the test under `integration_tests/sdk/apikey_test.py`


sdk/aqueduct/aqueduct_client.py
```diff
@@ -33,6 +34,21 @@
...
+    Returns:
+        The local API key.
+    """
+    server_directory = os.path.join(os.environ["HOME"], ".aqueduct", "server")
+    config_file = os.path.join(server_directory, "config", "config.yml")
+    with open(config_file, "r") as f:
+        try:
+            return str(yaml.safe_load(f)['apiKey'])
+        except yaml.YAMLError as exc:
+            print(exc)
```
cw75, line 50: Instead of just printing out the exception trace, let's in addition print out a warning saying "This API works only when you are running the server and the SDK on the same machine.", because if the user runs into this issue, it's likely that the server is a remote one, so we can't find the api key locally.

sdk/aqueduct/aqueduct_client.py
```diff
@@ -33,6 +34,21 @@
 import __main__ as main
 import os
 
+def apikey() -> str:
+    """
+    Get the local host API key if the server is running locally.
```
cw75, line 39: "Get the API key if the server is running locally."

sdk/aqueduct/aqueduct_client.py
```diff
@@ -33,6 +34,21 @@
 import __main__ as main
 import os
 
+def apikey() -> str:
+    """
+    Get the local host API key if the server is running locally. 
+
+    Returns:
+        The local API key.
```
cw75, line 42: "The API key."

sdk/aqueduct/aqueduct_client.py
```diff
@@ -33,6 +34,21 @@
 import __main__ as main
 import os
 
+def apikey() -> str:
```
cw75, line 37: let's change this api to `get_apikey()`

integration_tests/sdk/apikey_test.py
```diff
@@ -0,0 +1,6 @@
+from aqueduct import apikey
+
+def test_apikey_local_server(sp_client):
+    sdk_api_key = apikey()
+    assert(sdk_api_key == sp_client._api_client.api_key)
```
cw75, line 5: What is this asserting? What does `client._api_client.api_key` do? Is it just executing the same function as in line 4? If so, what's the goal of this test?

cw75 (changes requested):
LGTM overall! Left a few comments inline.

cw75 (approved):
LGTM!

PR #80, opened 2022-06-09
Improve and add a Golang linter

kenxu95:
Our enterprise linter used to just check a few basic formatting things. However, taking a look at all the cool linters here: https://golangci-lint.run/usage/linters/#errcheck I picked a number of them that either:

1) Would help us catch bugs earlier.
2) Are harmless nice-to-haves (eg. correcting mispellings)

I think there was at least one bug that this exercise found. (`makezero` is a nice one!)

To run linters, all you have to do is `golangci-lint run` from src/golang. All the appropriate configurations are in `.golangci.yml`. We do not run over test files.

These linters are pretty fast and run in parallel so performance is not a concern.

### DIFF
--- a/.github/workflows/lint-golang.yml
+++ b/.github/workflows/lint-golang.yml
@@ -1,0 +1,51 @@
+name: Golang Linter
+
+# Reference: https://github.com/golangci/golangci-lint-action
+
+on:
+  pull_request:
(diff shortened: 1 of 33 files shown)


hsubbaraj-spiral (approved):
This is great!

PR #81, opened 2022-06-09
Fix bug where the logs directory does not exist so executor always fails

kenxu95:
Don't compare with `os.ErrNotExist`, use `os.IsNotExist()` instead.


likawind (approved):
yay!

PR #84, opened 2022-06-10
Wipe binaries in `bin` before copying over newer versions to avoid mismatching Mac signature

cw75:
After doing some research, it seems there's some weird Mac signature issue when we try to `cp` over the new binary to try to replace the old one, according to [here](https://apple.stackexchange.com/questions/258623/how-to-fix-killed-9-error-in-mac-os). The issue seems to be that when we copy over the binary, the content got updated but the Mac signature is still using the cached old signature, and due to the mismatch, Mac immediately kills the binary when starting...

### DIFF
--- a/src/python/bin/aqueduct
+++ b/src/python/bin/aqueduct
@@ -91,7 +91,11 @@ def update_ui_version():
         execute_command(["rm", "-rf", ui_directory])
         exit(1)
 
-def download_golang_binaries(s3_prefix):
+def download_server_binaries():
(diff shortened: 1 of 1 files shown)


vsreekanti (approved):
I'm confused about why this was working on non-M1 Macs now. It seems to be an OS-level feature?

cw75:
@vsreekanti  My guess is that different versions of Mac may have different levels of signature enforcement policy. Anyway now it works on all Oses.

PR #98, opened 2022-06-13
Add ability to trigger existing flows with different parameters

kenxu95:
Ported over from my demo a few weeks ago. Unfortunately, there's not a great way to test that we're actually triggering with the new parameters, until we build the ability to fetch and inspect flow versions.

However, I was able to verify that this works from the UI :)


likawind (approved):
LGTM!

PR #101, opened 2022-06-14
Remove parameter operator from being displayed in _show_dag()

kenxu95:
Parameter operators are completely unnecessary nodes. We can just remove param operators after calculating the node positions for them. Does this look weird? I kinda like how the artifacts are all in the same column.

image

### DIFF
--- a/sdk/aqueduct/dag.py
+++ b/sdk/aqueduct/dag.py
@@ -98,7 +98,6 @@ class DAG(BaseModel):
     def list_operators(
         self,
         filter_to: Optional[List[OperatorType]] = None,
-        exclude: Optional[List[OperatorType]] = None,
         on_artifact_id: Optional[uuid.UUID] = None,
     ) -> List[Operator]:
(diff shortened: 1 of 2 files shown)


eunice-chan:
Any reason why we are stripping the parameters after calculating node positions?
Also, in a future iteration, it might be neat to have an endpoint for SDK and UI to use that strips away nodes / consolidates nodes into super nodes (e.g. artifacts node rather than artifact 1 node, artifact 2 node, etc.) for the creation of simplified DAGs.

kenxu95:
Yeah the reason is when I stripped the node out before `get_node_positions()`, the endpoint took removed other operators from it, which I didn't understand. This is more of a band-aid fix for now anyways. 

I'm thinking, long term, it might just make a lot more sense to just display a link here to a temporary rendering of the graph in the browser. Maintaining two separate versions of the DAG seems unwieldy, and after messing around with plotly a bit a doesn't seem that well suited to the task. Curious if you had thoughts about this?

PR #102, opened 2022-06-14
Implement system metrics for SDK

HarisChoudhary:
Allow the sdk to add system metrics to a table.


sdk/aqueduct/operators.py
```diff
@@ -151,6 +155,7 @@ class OperatorSpec(BaseModel):
     metric: Optional[MetricSpec]
     check: Optional[CheckSpec]
     param: Optional[ParamSpec]
+    systemmetric: Optional[SystemMetricSpec]
```
kenxu95, line 158: some comment as in the backend PR about using underscores to delimit multiple words

sdk/aqueduct/decorator.py
```diff
@@ -127,6 +137,40 @@ def wrap_spec(
...
+                        output_artifacts=[
+                            Artifact(
+                                id=system_output_artifact_id,
+                                name=artifact_name_from_op_name(metric_name),
+                                spec=metric_artifact_spec,
+                            )
+                        ],
+                    ),
+                ],
+            )
```
kenxu95, line 170: hmm why we do add every system metric to the dag here, instead of when the user requests it in `.system_metric()`. It seems somewhat unnecessary to always add it to the dag and force a calculation.

kenxu95 (changes requested):

sdk/aqueduct/decorator.py
HarisChoudhary: yeah, i moved it. I forgot we could request the calling op from the dag itself.

sdk/aqueduct/operators.py
```diff
@@ -135,6 +135,10 @@ class MetricSpec(BaseModel):
     function: FunctionSpec
 
 
+class SystemMetricSpec(BaseModel):
+    metricname: str
```
kenxu95, line 139: nit: same, "metric_name"

sdk/aqueduct/table_artifact.py
```diff
@@ -321,6 +324,66 @@ def internal_std_metric(table: pd.DataFrame) -> float:
...
+            raise AqueductError(
+                "Invalid metric requested: %s. Please choose a valid metric from: %s"
+                % (metric_name, ",".join(list(SYSTEM_METRICS_INFO.keys())))
+            )
+
+        operator = self._dag.must_get_operator(with_output_artifact_id=self._artifact_id)
+        system_operator_id = generate_uuid()
+        system_output_artifact_id = generate_uuid()
+        system_metric_description = SYSTEM_METRICS_INFO[metric_name][0]
+        system_metric_unit = SYSTEM_METRICS_INFO[metric_name][1]
```
kenxu95, line 350: nit: more pythonic -> `description, unit = SYSTEM_METRICS_INFO[metric_name]`

sdk/aqueduct/table_artifact.py
```diff
@@ -321,6 +324,66 @@ def internal_std_metric(table: pd.DataFrame) -> float:
...
+            ],
+        )
+
+        return MetricArtifact(
+            api_client=self._api_client,
+            dag=self._dag,
+            artifact_id=system_output_artifact_id,
+        )
+
+    def _associate_system_metric(self, metric_name: str, id: uuid.UUID) -> None:
```
kenxu95, line 384: THis method is unused?

sdk/aqueduct/table_artifact.py
```diff
@@ -73,6 +75,7 @@ def __init__(
         self._api_client = api_client
         self._dag = dag
         self._artifact_id = artifact_id
+        self._system_metric_map: Dict[str, uuid.UUID] = {}
```
kenxu95, line 78: this is unused

kenxu95 (approved):
Looks good just address the last few comments :)

sdk/aqueduct/table_artifact.py
HarisChoudhary: leftover from prev impl mb

sdk/aqueduct/table_artifact.py
HarisChoudhary: leftover last impl thanks for catching

PR #104, opened 2022-06-15
Add ability to create checks on table artifacts from Expectations

HarisChoudhary:
We have a new method on tables that allows us to take a expectation string and arguments and creates a check that returns a true/false based on the validation of the expectation.

### DIFF
--- a/sdk/aqueduct/table_artifact.py
+++ b/sdk/aqueduct/table_artifact.py
@@ -1,11 +1,25 @@
 from __future__ import annotations
 
 import json
-from typing import Callable, Any, List, Optional, Dict
+from typing import Callable, Any, Optional, Dict, Union, List
 import uuid
+
+from great_expectations.data_context.types.base import (
+    DataContextConfig,
+    DatasourceConfig,
+    FilesystemStoreBackendDefaults,
+)
+from great_expectations.data_context import BaseDataContext
+
+from great_expectations import DataContext
+from great_expectations.core import ExpectationSuite, ExpectationConfiguration
+from great_expectations.core.batch import RuntimeBatchRequest
+from great_expectations.validator.validator import Validator
 
 import pandas as pd
 import aqueduct
+import ruamel
+from ruamel import yaml
 
 from aqueduct.api_client import APIClient
 from aqueduct.artifact import ArtifactSpec
@@ -18,7 +32,7 @@ from aqueduct.dag import (
     RemoveCheckOperatorDelta,
     UpdateParametersDelta,
 )
-from aqueduct.enums import OperatorType, FunctionType, FunctionGranularity
+from aqueduct.enums import CheckSeverity, OperatorType, FunctionType, FunctionGranularity
 from aqueduct.error import (
     InvalidIntegrationException,
     AqueductError,
@@ -31,6 +45,7 @@ from aqueduct.operators import (
     FunctionSpec,
     MetricSpec,
     SystemMetricSpec,
+    CheckSpec,
 )
 from aqueduct.utils import (
     serialize_function,
@@ -43,6 +58,9 @@ from aqueduct.utils import (
 
 from aqueduct.generic_artifact import Artifact
 from aqueduct.metric_artifact import MetricArtifact
+from aqueduct.check_artifact import CheckArtifact
+
+OutputArtifact = Union[MetricArtifact, CheckArtifact]
 
 
 class TableArtifact(Artifact):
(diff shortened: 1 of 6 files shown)


cw75 (commented):
LGTM overall! @kenxu95 since you have more context around this could you give another pair of eyes on this? Thx.

sdk/aqueduct/table_artifact.py
```diff
@@ -156,6 +172,93 @@ def save(self, config: SaveConfig) -> None:
...
+        ge_check.get() // True or False based on expectation passing
+        Args:
+            expectation_name:
+                Name of built-in expectation to run with great_expectations
+            expectation_args:
+                Dictionary of args to pass into the expectation suite for the expectation being run.
+            severity:
+                Optional severity associated with the check created with this expectations
+        Returns:
+            A check artifact that represent the validation result of running the expectation provided on the table.
```
kenxu95, line 196: style nit: Leave a space after the first line, and before each subsection (Args and Returns)

sdk/aqueduct/table_artifact.py
```diff
@@ -156,6 +172,93 @@ def save(self, config: SaveConfig) -> None:
...
+                batch_identifiers={"default_identifier_name": "aq_table_check"},
+            )
+
+            # Constructing Validator by passing in RuntimeBatchRequest
+            my_validator: Validator = context.get_validator(
+                batch_request=runtime_batch_request,
+                expectation_suite=suite,
+            )
+
+            # run validation to return the result
```
kenxu95, line 251: style nit: be consistent with capitalization of the first letter for each comment.

sdk/aqueduct/table_artifact.py
```diff
@@ -156,6 +172,93 @@ def save(self, config: SaveConfig) -> None:
...
+                store_backend_defaults=FilesystemStoreBackendDefaults(root_directory="/tmp"),
+            )
+
+            context = BaseDataContext(project_config=data_context_config)
+
+            # create and load Expectation Suite
+            expectation_suite_name = "aq_check_expectation"
+            context.create_expectation_suite(
+                expectation_suite_name=expectation_suite_name, overwrite_existing=True
+            )
```
kenxu95, line 215: Doesn't `create_expectation_suite` return the suite already? Why do you need `get_expectation_suite()`?

sdk/aqueduct/table_artifact.py
```diff
@@ -156,6 +172,93 @@ def save(self, config: SaveConfig) -> None:
...
+                expectation_suite=suite,
+            )
+
+            # run validation to return the result
+            result = my_validator.validate()
+            return bool(result.success)
+
+        return self._apply_check_to_table(
+            great_expectations_check_method,
+            "ge_table_check: {%s}" % expectation_name,
```
kenxu95, line 257: "ge" might be slightly confused with "greater than or equal to", should we just call it "great expectations check: {name}"?

sdk/aqueduct/table_artifact.py
```diff
@@ -156,6 +172,93 @@ def save(self, config: SaveConfig) -> None:
...
+                Name of built-in expectation to run with great_expectations
+            expectation_args:
+                Dictionary of args to pass into the expectation suite for the expectation being run.
+            severity:
+                Optional severity associated with the check created with this expectations
+        Returns:
+            A check artifact that represent the validation result of running the expectation provided on the table.
+        """
+
+        def great_expectations_check_method(table: pd.DataFrame) -> bool:
```
kenxu95, line 199: Curious why you chose to create a new suite with a single expectation each time, instead of creating a single suite with multiple expectations? Is there an advantage of associating multiple expectations with a single suite? I'm not to familiar with what Great Expectations recommends.

sdk/aqueduct/table_artifact.py
```diff
@@ -321,6 +424,50 @@ def internal_std_metric(table: pd.DataFrame) -> float:
...
+                        )
+                    ],
+                ),
+            ],
+        )
+
+        return CheckArtifact(
+            api_client=self._api_client, dag=self._dag, artifact_id=output_artifact_id
+        )
+
```
kenxu95, line 470: Isn't this method almost exactly the same as `_apply_metric_to_table()`? Can we avoid duplicating all this? 

From what I can tell, the only difference is in the OperatorSpec and ArtifactSpec. Can we do a similar thing as `wrap_spec()`, where that helper just takes in an OperatorSpec and figures out what to return? You can cast the return value to the appropriate artifact.

kenxu95 (changes requested):
Nice! I'm not too familiar with this library, but have some questions

cw75:
@HarisChoudhary is this PR part of the preset project? If so, let's address the comment and get this merged this morning.

sdk/aqueduct/table_artifact.py
HarisChoudhary: Every check will run its own python process so the suite loaded from the context would be different for each check regardless. So we create one within the executor and check the validation we need to. For multiple validations within a check, this was brought up during planning and we need more first class support for certain representations to ensure that and i think its the aim of one of the upcoming projects. The code snippet itself was pieced together from the GE docs on how to load and validate expectations in memory.

sdk/aqueduct/table_artifact.py
HarisChoudhary: you're right, i was snipping code from the ge docs and missed this

kenxu95 (approved):

PR #106, opened 2022-06-15
Fix broken integration tests

kenxu95:
Fixes https://github.com/aqueducthq/aqueduct/actions/runs/2504138116. There were actually a number of issues here.

Ran the suite locally and it all passed.

### DIFF
--- a/sdk/aqueduct/aqueduct_client.py
+++ b/sdk/aqueduct/aqueduct_client.py
@@ -306,7 +306,7 @@ class Client:
     def trigger(
         self,
         flow_id: Union[str, uuid.UUID],
-        parameters: Optional[Dict[str, Any]],
+        parameters: Optional[Dict[str, Any]] = None,
     ) -> None:
         """Immediately triggers another run of the provided flow.
 
--- a/src/golang/cmd/server/request/parameters.go
+++ b/src/golang/cmd/server/request/parameters.go
@@ -3,8 +3,6 @@ package request
 import (
 	"encoding/json"
 	"net/http"
-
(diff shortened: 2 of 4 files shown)


sdk/aqueduct/aqueduct_client.py
```diff
@@ -306,7 +306,7 @@ def publish_flow(
     def trigger(
         self,
         flow_id: Union[str, uuid.UUID],
-        parameters: Optional[Dict[str, Any]],
+        parameters: Optional[Dict[str, Any]] = None,
```
kenxu95, line 309: issue 1: easy fix

src/golang/cmd/server/request/parameters.go
```diff
@@ -15,23 +13,15 @@ const (
...
 func ExtractParamsfromRequest(r *http.Request) (map[string]string, error) {
-	paramsInBytes, err := ExtractHttpPayload(
-		r.Header.Get(routes.ContentTypeHeader),
-		parametersKey,
-		false,
-		r,
-	)
-	if err != nil {
-		return nil, err
-	}
```
kenxu95, line 26: issue 2: for some reason ContentTypeHeader is not set when we post a request without a  `files` parameter. We can just grab the parameter from the body of the request.

agiron123 (approved):

likawind (approved):

src/golang/cmd/server/request/parameters.go
likawind: So parameters were in both header and body?

eunice-chan (approved):

PR #100, opened 2022-06-14
Implement Backend for System Metrics

HarisChoudhary:
We introduce the following:
- Operator for system metrics and the necessary scaffolding to execute it. 
- Changes to existing function executor to ensure that it pipes necessary system information for the metrics to represent.
- Change to get artifact_result endpoint to understand the new metadata format and return requested schema accordingly.

Verified all integration tests passing. 
Screen Shot 2022-06-14 at 2 25 52 PM

PR #107, opened 2022-06-15
Add migrator for system metrics schema change

HarisChoudhary:
We have schema changes to the artifact result when we are introducing system metrics in this PR:
https://github.com/aqueducthq/aqueduct/pull/100
this PR has the migrator to ensure that old flows dont break in the system.


saurav-c (approved):
LGTM! I'm guessing this has all been tested as part of #100?

HarisChoudhary:
> LGTM! I'm guessing this has all been tested as part of #100?

yeah i manually run the migrator to check back compat

PR #109, opened 2022-06-15
Fix build error for parcel/css on m1 mac.

agiron123:
Resolves:
https://linear.app/aqueducthq/issue/ENG-1248/bug-npm-run-start-fails-on-m1-mac

### DIFF
--- a/src/ui/app/package.json
+++ b/src/ui/app/package.json
@@ -9,7 +9,7 @@
     "lint:fix": "eslint '*/**/*.{js,ts,tsx}' --format table --fix"
   },
   "dependencies": {
-    "@aqueducthq/common": "^0.0.5",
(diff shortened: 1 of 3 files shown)


cw75:
@agiron123 how exactly is this change fixing the issue?

agiron123:
> @agiron123 how exactly is this change fixing the issue?

I googled the error message and found this thread:
https://github.com/parcel-bundler/parcel/issues/7962

Adding the @parcel/css package to package.json and removing package-lock.json and re-installing node_modules seemed to do the trick.

My hunch is that whoever last pushed to main didn't test on an M1 mac and used an Ubuntu box instead.

In the future, we should add M1 as a build target to CI/CD to ensure that our builds work for dev machines.

vsreekanti (approved):

PR #110, opened 2022-06-15
eng-1247 account page

eunice-chan:
The account page is back now :)
image

Had to update it to fit with the new UI structure. Here it is before: https://github.com/aqueducthq/aqueduct/blob/085cdf6692dc8419377a658e059dad2c243efd53/src/ui/app/pages/account.tsx

Side note: One thing I noticed was that my browser didn't reflect my UI changes until I switched to incognito and restarted the gateway so if you run into the issue you may want to try that.

### DIFF
--- a/src/ui/app/index.tsx
+++ b/src/ui/app/index.tsx
@@ -2,7 +2,7 @@ import React, { useEffect, useState } from 'react';
 import ReactDOM from 'react-dom/client';
 import { BrowserRouter, Routes, Route, Link, Navigate, useLocation } from "react-router-dom";
 import { Thing } from '../.';
-import { HomePage, DataPage, IntegrationsPage, IntegrationDetailsPage, WorkflowPage, WorkflowsPage, LoginPage } from '@aqueducthq/common';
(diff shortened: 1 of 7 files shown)


likawind:
Thanks for the quick fix! @vsreekanti @agiron123 given we've completely removed account page could you confirm whether this is intentional?

@eunice-chan most changes looks good! I'm wondering if we could bring back the code block for rendering. There are some reference in the older code: https://github.com/aqueducthq/enterprise/blob/main/ui/pages/account.tsx#L148 , https://github.com/aqueducthq/enterprise/blob/main/ui/components/text/codeblock.tsx , https://github.com/aqueducthq/enterprise/blob/main/ui/package.json#L40

vsreekanti:
It absolutely wasn't intentional, and I'm kinda shocked we somehow did this. Thanks for fixing this @eunice-chan!

eunice-chan:
image
@likawind Added syntax highlighting

likawind (approved):
Yay thanks for addressing comments!

PR #124, opened 2022-06-21
Parcel Build dist prefixing.

agiron123:
Adds build:dist option to Makefile to allow for prefixing of all urls in parceljs builds with /dist.

Removes need for using sed after generating bundles.


vsreekanti (approved):
Neat! Much better than my previous hack. 🙂

PR #125, opened 2022-06-21
Show parameters correctly on the UI

kenxu95:
Shows the parameter artifact in the reactflow dag, but not the parameter operator. I picked a Pencil icon because parameters are configurable, but open to other suggestions :P

A parameter named "dictionary" looks like this:

image


agiron123:
Looks good! If it's not already, can we make sure to use a monospace font for the value:
```
{ 'hello': 'world' }
```

part?

agiron123 (approved):

PR #128, opened 2022-06-22
Eng 1104 Allow users to specify function-specific requirements

Fanjia-Yan:
This PR implements the following: 
1. Enable passing in `reqs` as a parameter for `op` which contain the absolute path of a file that specify the requirements to run the specific operator.

The file I changes:
1. `decorator.py`: enable the `op` to take in an extra param `reqs` and pass it in to `serialize_function()`
2. `utils.py`: if the the `reqs` passed in is `None`, nothing changes. If `reqs` is specified, if the path is valid, copy the file into the directory, else raise `FileNotFound` Error

I wrote some integration tests in previous commits but decided to delete it since the test I thought of was a bit trivial, what I wrote is that:
1. passed in a definitely not existing file path and assert it for FileNotFound Error,
3. passed in a requirements.txt from Sentiment Analysis and assert the completeness of the table

Looking for some suggestions!


sdk/aqueduct/decorator.py
```diff
@@ -160,6 +161,9 @@ def op(
             A list of relative paths to files that the function needs to access.
             Python classes/methods already imported within the function's file
             need not be included.
+        reqs:
+            An absolute path to file that specify requirements for the specific
```
kenxu95, line 165: grammar nit: "specifies requirements for this specific operator"

sdk/aqueduct/utils.py
```diff
@@ -184,6 +186,8 @@ def _package_files_and_requirements(
         file_dependencies:
             Paths of file dependencies the function uses. Note that the paths are relative to the
             file the function is defined in.
+        reqs:
+            An absolute path of file that specify the requirements of the operator.
```
kenxu95, line 190: but it can also be relative right? Like I can do `../sibling_dir/requirements.txt`

sdk/aqueduct/decorator.py
kenxu95: Lets call this `reqs_path` for clarity.

sdk/aqueduct/utils.py
```diff
@@ -223,17 +227,27 @@ def _package_files_and_requirements(
...
-    if os.path.exists(REQUIREMENTS_FILE):
-        Logger.logger.info(
-            "%s: requirements.txt file detected in current directory %s, will not self-generate by inferring package dependencies."
-            % (os.getcwd(), func.__name__)
-        )
-        shutil.copy(REQUIREMENTS_FILE, os.path.join(dir_path, REQUIREMENTS_FILE))
+    if reqs:
+        if os.path.exists(reqs):
+            Logger.logger.info(
+                "%s: requirements.txt file detected in user-provided directory, will not self-generate by inferring package dependencies."
```
kenxu95, line 233: not sure I get these comments, what do you mean by self-generate?

sdk/aqueduct/utils.py
```diff
@@ -138,6 +139,8 @@ def serialize_function(
             The function to package
         file_dependencies:
             List of file dependencies the function uses
+        reqs:
+            A path to file that specify requirements
```
kenxu95, line 143: nit: "specifies"

kenxu95 (changes requested):
The first test case you laid out was good. Give the operator a bad path and assert that you get the specific error back.

Other test cases I'd like to see are:
- Run an operator that uses a library not already installed in by the test suite.
1) Do the above for user defined requirements
2) Do the above for default requirements

Fanjia-Yan:
Update the PR with the following changes:
1.change the naming `reqs` to `reqs_path`
2.edit grammar mistake in doc string and log console
3. Integration tests:
    -Provide @op() with a invalid path file, assert for FileNotFound Error
    -Create a default operator which its operation requires `transformers` package. During the execution of the operator, assert for AqueductError(inside it is moduleNotFound Error but show up as AqueductError)
    -Provide a operator with a valid path file to "requirements.txt". "requirements.txt" has `transformers` which the operator needs to execute. After execution, assert for the correctness of the dataframe from the outputartifact.get()

integration_tests/sdk/requirement_test.py
```diff
@@ -0,0 +1,66 @@
...
+    db = client.integration(name=get_integration_name())
+    table = db.sql(query=SENTIMENT_SQL_QUERY)
+    with pytest.raises(FileNotFoundError):
+        invalid_path_table = invalid_valid_path_operator(table)
+
+
+@pytest.mark.skipif(
+    condition=check_if_transformers_exist(),
+    reason="the transformers package already exists so the error can't be triggered.",
+)
```
kenxu95, line 52: I think this is a good enough solution for now. Can you double check that this test always runs in github actions?  There may be issues where if we ever install `transformers` as part of the integration test setup, this test will be skipped. I don't anticipate that happening though.

integration_tests/sdk/requirement_test.py
```diff
@@ -0,0 +1,66 @@
...
+@op(reqs_path=VALID_REQUIREMENTS_PATH)
+def valid_sentiment_prediction(reviews: pd.DataFrame) -> pd.DataFrame:
+    import transformers
+
+    model = transformers.pipeline("sentiment-analysis")
+    return reviews.join(pd.DataFrame(model(list(reviews["review"]))))
+
+
+@op()
+def default_sentiment_prediction(reviews: pd.DataFrame) -> pd.DataFrame:
```
kenxu95, line 35: can we name these clearer (its not obvious that its the reqs path that is invalid/valid):
```
sentiment_prediction_without_reqs_path()
```
and above would be:
```
sentiment_prediction_with_valid_reqs_path()
```
and `sentiment_prediction_with_invalid_reqs_path()`

sdk/aqueduct/utils.py
```diff
@@ -223,17 +228,24 @@ def _package_files_and_requirements(
...
+            Logger.logger.info("Installing requirements found at {path}".format(path=reqs_path))
+            shutil.copy(reqs_path, os.path.join(dir_path, REQUIREMENTS_FILE))
+        else:
+            raise FileNotFoundError("Requirement file provided does not exist.")
     else:
-        Logger.logger.info(
-            "%s: No requirements.txt file detected, self-generating file by inferring package dependencies."
-            % func.__name__
-        )
+        if os.path.exists(REQUIREMENTS_FILE):
```
kenxu95, line 238: Lets not nest conditionals unnecessarily. This can be:
```
elif os.path.exists(REQUIRMENTS_FILE)
```
(and the else below would be unindented too)

kenxu95 (approved):
A couple more comments I'd like to see addressed, but other than that looks good to me :D

cw75:
@Fanjia-Yan reminder to update PR title to complete the sentence.

PR #130, opened 2022-06-22
Fixes error handling in get_apikey

vsreekanti:
There was an extraneous print (L51) with a typo (`exec` instead of `exc`) -- funnily enough, this failed when I was generating API docs with `pydoc-markdown`. 

This PR removes the extraneous print statement.


Fanjia-Yan (approved):
My bad with the typo :(

PR #144, opened 2022-06-25
[Python executor] Improve logging for user function execution

likawind:
This PR adds logging and error handling objects so that we can clearly generate user errors separated to system errors. Specifically:
* Fixed function execution bash script so that its exit code is consistent with commands in the script.
* adds a simple error library allowing error owner to generate 'context' (which is typically a stack trace) and 'tip' (which is a user-friendly message around how to fix).
* adds a logger with io redirect supports.
* based on above change, all user function including extract, load, and op_execution are redirected. All these functions are now associated with detailed tips for fixing.
* I'm not very happy with the naming `ExecutionLogs`, better suggestions are welcomed
* This PR establishes the following pattern to handle user errors:
```
run(storage, spec): # most system code
  logger=Logger()
  try:
    @logger.redirect_user_fn(tip='...') # provides a tip based on the user fn
    def _run_user():
      actual_user_fn()
    _run_user() # execution results will be captured by logger
    utils.write_logs(storage, spec.metadata_path, logger)
  catch Exception as e:
    logger.code = ExecutionCode.FAILED
    logger.failure_reason = ExecutionCode.SYSTEM
    logger.error = Error(
        context=exception_traceback(e),
        tip=TIP_UNKNOWN_ERROR,
    )
    print(f"Failed with system error. Full Logs:\n{logger.json()}")
    utils.write_logs(storage, spec.metadata_path, logger)
    sys.exit(1)

if __name__ == "__main__":
   ...# some parsing
   run(storage, spec)
```

## Not included:
This PR is for python execution change only. We will have additional PRs for golang, SDK, and UI changes to consume and render the errors properly.

https://linear.app/aqueducthq/issue/ENG-1249/cleanly-fail-and-handle-user-vs-system-errors


src/python/aqueduct_executor/operators/utils/logging.py
```diff
@@ -0,0 +1,132 @@
...
+    "We are sorry to see this :(. "
+    f"You could send over a bug report through github issue {_GITHUB_ISSUE_LINK} "
+    " or in our slack channel. We will get back to you as soon as we can."
+)
+TIP_UNKNOWN_ERROR = f"An unexpected error occurred. {_TIP_CREATE_BUG_REPORT}"
+TIP_INTEGRATION_CONNECTION = (
+    "We have trouble connecting to the integration. "
+    "Please check your credentials or your integraiton provider."
+)
+TIP_DEMO_CONNECTION = "We have trouble connecting to demo DB. {_TIP_CREATE_BUG_REPORT}"
```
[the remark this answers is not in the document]
likawind: Demo credentials are not controlled by the user, and if the API key is wrong, there should be trouble before the user's able to reach this far (e.g. client creation error). In either case, I don't think the user's responsible for fixing, so I'm inclined to have us providing help

src/python/aqueduct_executor/operators/utils/logging.py
```diff
@@ -0,0 +1,132 @@
...
+    context: str = ""
+    tip: str = ""
+
+
+class Logs(BaseModel):
+    stdout: str = ""
+    stderr: str = ""
+
+
+class ExecutionLogs(BaseModel):
```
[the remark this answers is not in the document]
likawind: Yeah I'm not super happy about the naming here. Do you have any good suggestions? I feel `ExecutionStatus` could work but that kinda conflicts with the enum name https://github.com/aqueducthq/aqueduct/blob/main/src/golang/lib/collections/shared/types.go#L9

src/python/aqueduct_executor/operators/utils/execution.py
```diff
@@ -0,0 +1,131 @@
...
+
+class Logs(BaseModel):
+    stdout: str = ""
+    stderr: str = ""
+
+
+class ExecutionState(BaseModel):
+    user_logs: Logs
+    status: ExecutionStatus = ExecutionStatus.PENDING
+    failure_type: FailureType = FailureType.SUCCESS
```
kenxu95, line 43: it seems strange to have a `FailureType.SUCCESS`. Any reason this can't just be an `Optional[FailureType]`, and anyone using ExecutionState can just check `ExecutionState.failure_type is not None`?

src/python/aqueduct_executor/operators/utils/execution.py
```diff
@@ -0,0 +1,131 @@
...
+TIP_DEMO_CONNECTION = f"We have trouble connecting to demo DB. {_TIP_CREATE_BUG_REPORT}"
+
+TIP_EXTRACT = "We couldn't execute the provided query. Please double check your query is correct."
+TIP_LOAD = "We couldn't load to the integration. Please make sure the target exists, or you have the right permission."
+TIP_DISCOVER = "We couldn't list items in the integration. Please make sure your credentials have the right permission."
+
+
+class Error(BaseModel):
+    context: str = ""
+    tip: str = ""
```
kenxu95, line 32: Can we document here what these two fields represent, and how they will get surfaced to the user?

src/python/aqueduct_executor/operators/utils/execution.py
```diff
@@ -0,0 +1,131 @@
...
+                        context=stack_traceback(
+                            offset=1
+                        ),  # traceback the first stack frame, which belongs to user
+                        tip=failure_tip,
+                    )
+                    print(f"User failure. Full log: {self.json()}")
+                    return None
+
+                # Include the stack trace within the user's code.
+                fetch_redirected_logs(stdout_log, stderr_log, self.user_logs)
```
kenxu95, line 69: Do we need to call `fetch_redirected_logs() twice? Shouldn't a
```
finally:
    fetch_redirected_logs(...)
```
work the same way.

kenxu95: Also `fetch_redirected_logs()` actually populates the logs argument, can we rename it `set_standard_logs(...)` or something?

src/python/aqueduct_executor/operators/utils/utils.py
```diff
@@ -243,19 +244,17 @@ def _write_json_output(
     )
 
 
-def write_operator_metadata(
+def write_logs(
```
kenxu95, line 247: shouldn't this be called `write_exec_state()`?

src/python/aqueduct_executor/start-function-executor.sh
```diff
@@ -1,16 +1,25 @@
 #!/bin/bash
 JOB_SPEC=$1
 OP_PATH=$(python3 -m aqueduct_executor.operators.function_executor.get_extract_path --spec "$JOB_SPEC")
+EXIT_CODE=$?
+if [ $EXIT_CODE != "0" ]; then exit $(($EXIT_CODE)); fi
+
 python3 -m aqueduct_executor.operators.function_executor.extract_function --spec "$JOB_SPEC"
+EXIT_CODE=$?
+if [ $EXIT_CODE != "0" ]; then exit $(($EXIT_CODE)); fi
```
kenxu95, line 9: This is a bug fix right? Just to clarify, previously the function-executor would just continue on even if extracting the function failed?

src/python/aqueduct_executor/operators/utils/enums.py
```diff
@@ -13,6 +13,19 @@ def __contains__(cls, item: Any) -> Any:
...
 
+class ExecutionStatus(str, Enum, metaclass=MetaEnum):
+    UNKNOWN = "unknown"
+    PENDING = "pending"
+    SUCCEEDED = "succeeded"
+    FAILED = "failed"
+
+
+class FailureType(Enum, metaclass=MetaEnum):
+    SUCCESS = 0
```
kenxu95, line 24: Same comment as below. Can we not have a `FailureType.SUCCESS`?

src/python/aqueduct_executor/operators/utils/enums.py
```diff
@@ -13,6 +13,19 @@ def __contains__(cls, item: Any) -> Any:
         return item in [v.value for v in cast(Iterable[Enum], cls.__members__.values())]
 
 
+class ExecutionStatus(str, Enum, metaclass=MetaEnum):
+    UNKNOWN = "unknown"
```
kenxu95, line 17: When is an execution state ever unknown? Doesn't it always start at PENDING then become either SUCCEEDED or FAILED?

kenxu95 (changes requested):
Sry for the lateness, I think you need to re-request review from me or I won't see it. Generally looks good! but have some feedback

src/python/aqueduct_executor/operators/utils/execution.py
kenxu95: or I guess they would just check if `ExecutionState.error is not None`

kenxu95: Can you note in a comment here what each of these fields means actually? Whats the source of truth of whether an execution state succeeded? Does an error != None also mean that `failure_type` is also none (or SUCCESS), and that status must be `FAILED`? The relationship between all these fields may be unclear to someone not familiar with all  the code.

src/python/aqueduct_executor/start-function-executor.sh
likawind: yes. Also we need to exit with `$((EXIT_CODE))` since the script will exit with 0 even if the underlining cmd failed

src/python/aqueduct_executor/operators/utils/execution.py
likawind: I think `Optional[FailureType]` is a good idea which avoids a lot of confusion

src/python/aqueduct_executor/operators/utils/execution.py
likawind: the subsequent steps of `fetch_redirected_logs()` are different when handling exceptions vs succeeded, so I don't think `finally` would work here.

Could you add more details on the second part? Don't quite get it :)

src/python/aqueduct_executor/operators/utils/enums.py
likawind: It's here mainly to be consistent with go code, and go use `unknown` internally for job status handling.

kenxu95: ok I see its like a default value we use when we return an error or something

src/python/aqueduct_executor/operators/utils/execution.py
kenxu95: Yeah I was just saying since the function takes in a stdout and stderr and writes those streams to `self.user_logs`, so its not really doing a fetch. A "set" might be more appropriate here.

kenxu95 (approved):
lgtm, thanks!

PR #148, opened 2022-06-27
Fix mypy errors on python executors

likawind:
This PR clears all `mypy` issues. Most fixes involves fixing and cleaning import types or function signature typings. The main follow up required is https://linear.app/aqueducthq/issue/ENG-1285/revisit-python-typings-for-connector-code where we need to rethink the implementation in the case of inheritance. 
In addition to the list below, verified `mypy` and `black` is clean:
```
(base) ubuntu@ip-172-31-45-190:~/aqueduct/src/python$ mypy aqueduct_executor --ignore-missing-imports --exclude tests
Success: no issues found in 47 source files
```
- [x] If this is a new feature, I have added unit tests and integration tests. (This is not a new feature)
- N/A I have manually run the integration tests and they are passing.
- N/A All features on the UI continue to work correctly.
- Verified mypy is passing on subsequent PRs: https://github.com/aqueducthq/aqueduct/pull/160


src/python/aqueduct_executor/operators/connectors/tabular/connector.py
```diff
@@ -31,7 +35,12 @@ def extract(self, params: extract.Params) -> pd.DataFrame:
         """
 
     @abstractmethod
-    def load(self, params: load.Params, df: pd.DataFrame) -> None:
+    def load(
+        self,
+        # TODO: https://linear.app/aqueducthq/issue/ENG-1285/revisit-python-typings-for-connector-code
```
kenxu95, line 40: nit: can we format these like `TODO(ENG-1285): {short human readable description}`

src/python/aqueduct_executor/operators/connectors/tabular/main.py
```diff
@@ -117,54 +125,57 @@ def setup_connector(
             PostgresConnector as OpConnector,
         )
     elif connector_name == common.Name.SNOWFLAKE:
-        from aqueduct_executor.operators.connectors.tabular.snowflake import (
+        from aqueduct_executor.operators.connectors.tabular.snowflake import (  # type: ignore
```
kenxu95, line 128: Will all these type ignores be fixed by ENG-1285?

kenxu95 (changes requested):
Two requests:
- Can you update the name of the PR?
- Can you add this mypy command to the `mypy.yml` gh action?

src/python/aqueduct_executor/operators/connectors/tabular/main.py
likawind: No this is another issue where mypy can't deal with case-based imports. I think ignoring mypy here is the right thing since our use case is a bit special.

.github/workflows/mypy.yml
```diff
@@ -29,7 +29,12 @@ jobs:
...
-      - name: mypy
+      - name: mypy sdk
         working-directory: sdk
         run: |
           mypy -m aqueduct --ignore-missing-imports --strict
+      
+      - name: mypy executor
+        working-directory: src/python
+        run: |
+          mypy aqueduct_executor --ignore-missing-imports --exclude tests
```
cw75, line 40: Why is the flags here different from the one above? Also should we add mypy check to the integration tests directory as well? @likawind

likawind: We should do integration tests but I would prefer to do in a separate PR, since there are probably some additional fix involved.

@kenxu95 do you know how the commands are different? I couldn't properly run run using `-m aqueduct_executor` locally, it only check one file instead of the entire tree

kenxu95: I don't know exactly why they're different. I also saw the same issue with `-m`, but I found this command which seemed to give sane results. I wonder if `-m` only looks at the top-level directory?

likawind: @cw75 did you see any difference running w wo `-m` for SDK?

cw75: I'm looking into the right flag to use now. Pretty sure at this point `-m` only looks at the python files at the top level directory.

cw75: @likawind @kenxu95  I think running `mypy .` recursively checks all the Python files. We shouldn't use the `-m` flag.

cw75: @likawind after some investigation with @kenxu95 , we think we should add:
1. "--ignore-missing-imports"
2. "--strict"
3. "--exclude" if necessary
4. "--implicit-reexport"
flags to all mypy commands.

.github/workflows/mypy.yml
```diff
@@ -29,7 +29,35 @@ jobs:
         run : |
           python3 -m pip install types-croniter types-requests types-PyYAML
 
-      - name: mypy
+      - name: mypy sdk
         working-directory: sdk
         run: |
           mypy -m aqueduct --ignore-missing-imports --strict
```
cw75, line 35: @likawind could you also add `--implicit-reexport` so that it's consistent?

cw75 (approved):
@likawind one last comment inline but otherwise LGTM!

.github/workflows/mypy.yml
kenxu95: Would it also be possible to get rid of the `-m` (or in a follow-up?)

.github/workflows/mypy.yml
```diff
@@ -29,7 +29,35 @@ jobs:
...
+
+  executor:
```
kenxu95, line 37: we need to have this trigger on changes to the `src/python/**` paths too right?

src/python/aqueduct_executor/operators/connectors/tabular/connector.py
```diff
@@ -20,7 +20,12 @@ def discover(self) -> List[str]:
         """
 
     @abstractmethod
-    def extract(self, params: extract.Params) -> pd.DataFrame:
+    def extract(  # type: ignore
+        self,
+        # TODO (ENG-1285): Revisit the typing issue that araises from inheritence
+        # https://linear.app/aqueducthq/issue/ENG-1285/revisit-python-typings-for-connector-code
```
kenxu95, line 26: nit: the link isn't strictly necessary once we have the task number.

src/python/aqueduct_executor/operators/connectors/tabular/load.py
```diff
@@ -7,13 +7,13 @@
...
     table: str
-    update_mode: Optional[common.UpdateMode] = common.UpdateMode.REPLACE
+    update_mode: common.UpdateMode = common.UpdateMode.REPLACE
 
     class Config:
         validate_assignment = True
 
     @validator("update_mode")
-    def set_update_mode(cls, update_mode):
+    def set_update_mode(cls, update_mode):  # type: ignore
```
kenxu95, line 16: wait why can't we type `update_mode: common.UpdateMode` here? And have a None return value.

kenxu95 (changes requested):
This is a great step forward! Blocking just cus I want to make sure we trigger the mypy action on the appropriate changed paths.

src/python/aqueduct_executor/operators/connectors/tabular/load.py
likawind: The validator has some implicit assumption on typing, and mypy complains when the input is `common.UpdateMode`. I think it's fine to ignore the typing check here

kenxu95 (approved):

PR #149, opened 2022-06-27
Implement a script to run all linters

cw75:
This PR adds a Python script that the developers can use to run all linters. We should be running this prior to making future PRs so that the code is cleaned before the reviewers review them. The exact usage of the script is described in the script.

This PR also does a cleanup of an unnecessary module import of `ruamel` and includes `ruamel.yaml` to our SDK's `requirements.txt`s.

### DIFF
--- a/scripts/run_linters.py
+++ b/scripts/run_linters.py
@@ -76,9 +76,9 @@ if __name__ == "__main__":
         args.lint_golang = True
         args.lint_ui = True
 
-    print("Current directory should be the root directory of the aqueduct repo.")
     cwd = os.getcwd()
     if not cwd.endswith("aqueduct"):
+        print("Current directory should be the root directory of the aqueduct repo.")
         print("Your working directory is %s" % cwd)
         exit(1)
 
--- a/scripts/run_linters.py
+++ b/scripts/run_linters.py
@@ -29,7 +29,8 @@ def lint_python(cwd):
     execute_command(["black", join(cwd, "src/python"), "--line-length=100"])
     execute_command(["black", join(cwd, "sdk"), "--line-length=100"])
     execute_command(["black", join(cwd, "integration_tests"), "--line-length=100"])
-    execute_command(["mypy", "-m", "aqueduct", "--ignore-missing-imports", "--strict"], join(cwd, "sdk"))
+    execute_command(["mypy", ".", "--ignore-missing-imports", "--strict", "--exclude", "aqueduct/tests", "--implicit-reexport"], join(cwd, "sdk"))
+    execute_command(["mypy", ".", "--ignore-missing-imports", "--strict", "--exclude", "tests", "--implicit-reexport"], join(cwd, "src/python"))
 
 def lint_golang(cwd):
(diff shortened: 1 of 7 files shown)


scripts/run_linters.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def execute_command(args, cwd=None):
+    with subprocess.Popen(args, stdout=sys.stdout, stderr=sys.stderr, cwd=cwd) as proc:
+        proc.communicate()
+        if proc.returncode != 0:
+            raise Exception("Error executing command: %s" % args)
+
+def lint_python(cwd):
+    execute_command(["black", join(cwd, "src/python"), "--line-length=100"])
```
vsreekanti, line 29: This is the type of stuff we should get out of the habit of to be better about supporting Windows in the future. Rather than doing `join(cwd, 'src/python')`, we can just do `join(cwd, 'src', 'python')`

scripts/run_linters.py
```diff
@@ -0,0 +1,95 @@
...
+    execute_command(["mypy", ".", "--ignore-missing-imports", "--strict", "--exclude", "aqueduct/tests", "--implicit-reexport"], join(cwd, "sdk"))
+    execute_command(["mypy", ".", "--ignore-missing-imports", "--strict", "--exclude", "tests", "--implicit-reexport"], join(cwd, "src/python"))
+
+def lint_golang(cwd):
+    execute_command(["golangci-lint", "run", "--concurrency=4", "--fix"], join(cwd, "src/golang"))
+
+def lint_ui(cwd):
+    execute_command(["npm", "install"], join(cwd, "src/ui/common"))
+    execute_command(["npm", "run", "lint", "--", "--fix"], join(cwd, "src/ui/common"))
+    execute_command(["npm", "link"], join(cwd, "src/ui/common"))
```
vsreekanti, line 41: Do we need to link common in order to lint?

scripts/run_linters.py
```diff
@@ -0,0 +1,95 @@
...
+    )
+
+    args = parser.parse_args()
+
+    if not (args.lint_python or args.lint_golang or args.lint_ui):
+        args.lint_python = True
+        args.lint_golang = True
+        args.lint_ui = True
+
+    print("Current directory should be the root directory of the aqueduct repo.")
```
vsreekanti, line 79: Why do we need to print this every time?

vsreekanti (approved):

scripts/run_linters.py
```diff
@@ -0,0 +1,95 @@
...
+
+import argparse
+import os
+import subprocess
+import sys
+
+from os.path import join
+
+
+def execute_command(args, cwd=None):
```
likawind, line 22: not an immediate thing but I think these helper functions could go to some shared library in future

likawind (approved):

scripts/run_linters.py
kenxu95: I actually think this is somewhat helpful to telling the user where to run the script from? Or we can flush out the error message on line 82 if we want to avoid this spew.

kenxu95 (commented):
lgtm but will leave final review for others that have commented more

PR #151, opened 2022-06-28
ENG-1271 Show user a message saying that we do not support previewing base data for an S3 integration

eunice-chan:
image

ENG-1271 Show user a message saying that we do not support previewing base data for an S3 integration

- Not a new feature.
- I don't think we have integration tests for UI.

### DIFF
--- a/src/ui/common/src/components/pages/integration/id/index.tsx
+++ b/src/ui/common/src/components/pages/integration/id/index.tsx
@@ -166,10 +166,10 @@ const IntegrationDetailsPage: React.FC<IntegrationDetailsPageProps> = ({
   }
 
   let preview = (
-    <Alert severity="error" sx={{ width: '80%' }}>
+    <Alert severity="warning" sx={{ width: '80%' }}>
         <>
           We currently do not support listing data in an S3 bucket. But
-          don&apos;t worry--we&apos;re working on adding this feature! If you
+          don&apos;t worry&mdash;we&apos;re working on adding this feature! If you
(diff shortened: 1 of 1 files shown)


src/ui/common/src/components/pages/integration/id/index.tsx
```diff
@@ -158,6 +165,110 @@ const IntegrationDetailsPage: React.FC<IntegrationDetailsPageProps> = ({
     return null;
   }
 
+  let preview = (
+    <Alert severity="error" sx={{ width: '80%' }}>
+        <>
+          We currently do not support listing data in an S3 bucket. But
+          don&apos;t worry--we&apos;re working on adding this feature! If you
```
vsreekanti, line 172: Can we use an m-dash instead of 2 n-dashes here? I think you can do it with `&mdash;`.

src/ui/common/src/components/pages/integration/id/index.tsx
```diff
@@ -158,6 +165,110 @@ const IntegrationDetailsPage: React.FC<IntegrationDetailsPageProps> = ({
     return null;
   }
 
+  let preview = (
+    <Alert severity="error" sx={{ width: '80%' }}>
```
vsreekanti, line 169: Nit -- let's making this `warning` instead of `error`.

vsreekanti (approved):

eunice-chan:
image

vsreekanti:
Perfect, that looks great -- thanks Eunice!

PR #152, opened 2022-06-28
[Backend] Adds Parameterization to SQL Queries

kenxu95:
Adds the ability to execute sql queries with parameters that are defined by the user.

This entails:
- Loosing the restriction that extract operators cannot have any inputs.
- Because the extract operator needs to know the name of the parameter so it can replace the correct in the query, we write the parameter name as the metadata as part of the parameter operator. Then, when the extract operator consumes the parameter operator's output, i'll look at the input metadata for the parameter name. Alternatively, we could try to have the orchestration engine in go attempt to pull this information out and stick it on the ExtractSpec. What do you think? I worry a bit that adding another field to the metadata may make it slightly harder to migrate away from in the future.

ENG-1258

- [x ] I have manually run the integration tests and they are passing.


src/golang/lib/job/spec.go
```diff
@@ -122,11 +123,15 @@ type SystemMetricSpec struct {
...
-	ConnectorConfig    auth.Config             `json:"connector_config"  yaml:"connector_config"`
-	Parameters         connector.ExtractParams `json:"parameters"  yaml:"parameters"`
-	OutputContentPath  string                  `json:"output_content_path"  yaml:"output_content_path"`
-	OutputMetadataPath string                  `json:"output_metadata_path"  yaml:"output_metadata_path"`
+	ConnectorName   integration.Service     `json:"connector_name"  yaml:"connector_name"`
+	ConnectorConfig auth.Config             `json:"connector_config"  yaml:"connector_config"`
+	Parameters      connector.ExtractParams `json:"parameters"  yaml:"parameters"`
+
+	// Inputs can only be parameters.
+	InputContentPaths  []string `json:"input_content_paths" yaml:"input_content_paths"`
```
likawind, line 131: we really need a task to make all paths a part of some struct like `Artifact` and pass the struct around, instead of n dict / lists

src/python/aqueduct_executor/operators/connectors/tabular/extract.py
```diff
@@ -1,13 +1,88 @@
...
 from aqueduct_executor.operators.connectors.tabular import common, models
+from aqueduct_executor.operators.utils import constants
+
+# Regular Expression that matches any substring appearance with
+# "{{ }}" and a word inside with optional space in front or after
+# Potential Matches: "{{today}}", "{{ today  }}""
+#
+# Duplicated in the SDK at `sdk/aqueduct/integrations/sql_integration.py`.
+# Make sure the two are in sync.
+TAG_PATTERN = r"{{\s*\w+\s*}}"
```
likawind, line 13: not an regex expert, does this allow some special characters like underscore and dashes as parameter name?

src/python/aqueduct_executor/operators/utils/utils.py
```diff
@@ -56,11 +59,24 @@ def read_artifacts(
...
+    storage: Storage,
+    input_metadata_paths: List[str],
+) -> List[Dict[str, Any]]:
+    return _read_metadata_key(storage, input_metadata_paths, _METADATA_USER_DEFINED_KEY)
+
+
+def _read_metadata_key(
+    storage: Storage, input_metadata_paths: List[str], key_name: str
+) -> List[Dict[str, Any]]:
+    metadata_inputs = [_read_json_input(storage, input_path) for input_path in input_metadata_paths]
```
likawind, line 76: I believe pydantic will make these reading a lot easier, I have a comment below with more details

src/python/aqueduct_executor/operators/utils/utils.py
```diff
@@ -29,7 +32,7 @@ def read_artifacts(
     input_paths: List[str],
     input_metadata_paths: List[str],
     artifact_types: List[InputArtifactType],
-) -> List[InputArtifact]:
+) -> List[Any]:
```
likawind, line 35: why use Any instead of expanding the union?

src/python/aqueduct_executor/operators/param_executor/main.py
```diff
@@ -16,11 +16,14 @@ def run(spec: spec.ParamSpec) -> None:
...
             storage,
+            enums.OutputArtifactType.JSON,
             spec.output_content_path,
             spec.output_metadata_path,
             spec.val,
-            {},
-            enums.OutputArtifactType.JSON,
+            system_metadata={},
+            user_defined_metadata={
+                constants.PARAM_NAME_KEY: spec.param_name,
```
likawind, line 25: I don't fully understand the motivation passing around parameter name as an additional field in artifact metadata, as I thought the param name should be consistent with the param artifact name, and that's something we would want to pass around during execution. Happy to chat offline

likawind (changes requested):
Great! Requesting changes mainly for considering using pydantic instead of dict with fixed keys. Also having some clarification questions around the use of `user_defined_metadata`

src/python/aqueduct_executor/operators/connectors/tabular/extract.py
kenxu95: ah this actually misses dashes. Will fix.

kenxu95: It does not intend to catch spaces btw (eg. `{{ hello world }}` will not be matched)

Fanjia-Yan: @likawind the current regex `\w+` matches letters,numbers, and underscores but not dash `-` I believe. Do you think it is necessary to add that in?

src/golang/lib/job/spec.go
kenxu95: Created https://linear.app/aqueducthq/issue/ENG-1289/pass-an-artifact-interface-around-the-orchestration-engine to follow up next sprint

src/python/aqueduct_executor/operators/utils/utils.py
```diff
@@ -88,12 +104,14 @@ def _read_json_input(storage: Storage, path: str) -> Any:
 
 def write_artifacts(
     storage: Storage,
+    artifact_types: List[OutputArtifactType],
     output_paths: List[str],
     output_metadata_paths: List[str],
     contents: List[Any],
     system_metadata: Dict[str, str],
-    artifact_types: List[OutputArtifactType],
+    user_defined_metadata: Dict[str, str],
```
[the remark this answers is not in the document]
kenxu95: Created https://linear.app/aqueducthq/issue/ENG-1290/use-pydantic-object-for-artifact-metadata-in-backend-instead-of. I don't think this is in scope of this PR, since we don't the metadata path to pass information around anymore.

src/python/aqueduct_executor/operators/utils/utils.py
kenxu95: It's because parameters can inputs to these operators, and parameters can have any type as long as it is jsonable. Based on https://stackoverflow.com/questions/51291722/define-a-jsonable-type-using-mypy-pep-526, it seems like we either call it an `object` or `Any`. I chose the latter to make this library simpler, the idea being that it's the responsibility of the caller to cast/check any results to the appropriate type, and the mypy checks you're working on should help catch any issues around this.

src/python/aqueduct_executor/operators/param_executor/main.py
kenxu95: Discussed offline. We added the parameter names to the ExtractSpec itself instead of metadata.

likawind (approved):
LGTM!

PR #153, opened 2022-06-28
[SDK] Adds parameterization to SQL queries

kenxu95:
These are the API changes:
```
table_name_param = client.create_param("table_name", default="hotel_reviews")
sql_artifact = db.sql("select * from {table_name}")

sql_artifact.get(parameters={"table_name": "other table"})
```
 The parameter name can have underscores and dashes, but cannot have spaces. If there are spaces, it won't get picked up as a tag and the query will likely fail.

ENG-1258

- [ x] If this is a new feature, I have added unit tests and integration tests.
- [ x ] I have manually run the integration tests and they are passing.

### DIFF
--- a/sdk/aqueduct/integrations/sql_integration.py
+++ b/sdk/aqueduct/integrations/sql_integration.py
@@ -152,13 +152,12 @@ class RelationalDBIntegration(Integration):
             matches = re.findall(TAG_PATTERN, extract_params.query)
             for match in matches:
                 param_name = match.strip(" {}")
-
-                # If it is a built-in tag, we can ignore it for now, since the python operators will perform the expansion.
-                if param_name in BUILT_IN_EXPANSIONS:
-                    continue
-
                 param_op = self._dag.get_operator(with_name=param_name)
                 if param_op is None:
+                    # If it is a built-in tag, we can ignore it for now, since the python operators will perform the expansion.
+                    if param_name in BUILT_IN_EXPANSIONS:
+                        continue
+
                     raise InvalidUserArgumentException(
                         "There is no parameter defined with name `%s`." % param_name,
                     )
--- a/sdk/aqueduct/integrations/sql_integration.py
+++ b/sdk/aqueduct/integrations/sql_integration.py
@@ -167,7 +167,8 @@ class RelationalDBIntegration(Integration):
             param_val = json.loads(param_op.spec.param.val)
             if not isinstance(param_val, str):
                 raise InvalidUserArgumentException(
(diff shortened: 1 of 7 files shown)


likawind (commented):
Just making sure I understood - the main changes in this PR is to attach the params operator when there's an user defined params?

kenxu95:
Yep, specifically its adding the parameter artifact as input to the extract operator when it detects them.

sdk/aqueduct/integrations/sql_integration.py
```diff
@@ -127,14 +145,43 @@ def sql(
...
+                    raise InvalidUserArgumentException(
+                        "There is no parameter defined with name `%s`." % param_name,
+                    )
+
+                # Check that the parameter corresponds to a string value.
+                assert param_op.spec.param is not None
+                param_val = json.loads(param_op.spec.param.val)
+                if not isinstance(param_val, str):
+                    raise InvalidUserArgumentException(
+                        "The parameter `%s` must be defined as a string. Instead, got type %s"
```
likawind, line 171: Is this checking applied to the default value? who would perform the similar checking in `run()` ?

sdk/aqueduct/integrations/sql_integration.py
```diff
@@ -127,14 +145,43 @@ def sql(
...
+        # Find any tags that need to be expanded in the query, and add the parameters that correspond
+        # to these tags as inputs to this operator. The orchestration engine will perform the replacement at runtime.
+        sql_input_artifact_ids = []
+        if extract_params.query is not None:
+            matches = re.findall(TAG_PATTERN, extract_params.query)
+            for match in matches:
+                param_name = match.strip(" {}")
+
+                # If it is a built-in tag, we can ignore it for now, since the python operators will perform the expansion.
+                if param_name in BUILT_IN_EXPANSIONS:
```
likawind, line 157: what happens if the user is 'overriding' the built-in name? Should we validate the value if there's an user defined params op exists?

sdk/aqueduct/integrations/sql_integration.py
kenxu95: The `UpdateParametersDelta` will perform this same check. Thats the delta that changes the parameter value temporarily, and it performs the string check if the parameter is consumed by any extract operator.

sdk/aqueduct/integrations/sql_integration.py
kenxu95: You're right. The backend will first check for user-defined expansion before built-in expansions, so this line should go within the `if param_op is None` check (we should only continue if its a built-in expansion but there doesn't exist a parameter for it)

likawind (approved):
LGTM, great work!

PR #156, opened 2022-06-28
Port input in config

eunice-chan:
The issues were:
- The port is hardcoded to 5432
- The host prompt is misleading (prompt was 127.0.0.1:5432 while this input was used just for the host/ip address part and the port was hardcoded to 5432 so if the user puts in 127.0.0.1:5432, the backend tries to access 127.0.0.1:5432:5432).

Manually tested it works with this input:
image
In the Python executor, the port is set to the user's input (not 5432) in the config and the engine url is correctly set to postgresql://test:test@127.0.0.1:4321/test.

```
"connector_config": {
        "conf": {
            "database": "test",
            "host": "127.0.0.1",
            "password": "test",
            "port": "4321",
            "username": "test"
        }
    }
```

This is ENG 1277.

[N/A] If this is a new feature, I have added unit tests and integration tests.
[N/A] I have manually run the integration tests and they are passing.

### DIFF
--- a/src/python/aqueduct_executor/operators/connectors/tabular/config.py
+++ b/src/python/aqueduct_executor/operators/connectors/tabular/config.py
@@ -24,7 +24,7 @@ class PostgresConfig(models.BaseConfig):
     password: str
     database: str
     host: str
-    port: Optional[str] = "5432"
+    port: str
 
 
 class S3Config(models.BaseConfig):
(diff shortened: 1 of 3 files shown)


vsreekanti (approved):
@eunice-chan, this looks good! 

A few nitpicky things I'd like us to get in the habit of doing -- I know I'm being annoying, but I promise these things will be helpful as we grow!

* Can we make sure to include a screenshot (or a Loom video when making bigger changes) whenever we modify the UI?
* Let's get in the habit of making our PR titles more descriptive rather than less. Typically, a good rule of thumb is that the PR title should complete the sentence "This PR...". In this case, we could do something like, "Adds UI input to capture custom port for Postgres integration"

eunice-chan:
Thanks for the reminder @vsreekanti! I wonder if we can add this to the PR template--perhaps as part of the checklist.

vsreekanti:
Great idea @eunice-chan! Why don't you go ahead and submit a PR with the changes to the template?

PR #157, opened 2022-06-28
Eng 1241 Add ability to fetch an artifact from an existing flow version

Fanjia-Yan:
This PR add `get(artifact_name)` feature for `flow` object. By passing in an artifact name, the function will return the Artifact associated with the artifact_name. If not, raise ArtifactNotFound Error.

I have created a integration test:
The test create a workflow and publish it online. After the workflow has finished, I use get() to extract the Artifact provided the artifact name. Finally, I checked if the returned Artifact object has the same name as provided.

Eng 1241
- [x ] I have performed a self-review of my code.
- [x ] If this is a new feature, I have added unit tests and integration tests.
- [x ] I have manually run the integration tests and they are passing.

## Update
Many changes to this PR after last review
1. Complete the `get_workflow()` so that the `workflowresponse` also has serialized function attached to each operator. To get the serialized function, I route "/api/function/%s/export" and request by operator_id
2. Modify the previous `flow.get()` into `flow.artifact()`. In `flow.artifact()`, we first fetch the `artifact.Artifact` object with matching `artifact_name` and convert it to `generic_artifact.Artifact` so that we can call `.get()`.
3. For the integration test. I fetch the artifact using `flow.artifact(output_artifact.name())`. And check for name equality and dataframe equality between locally created artifact and fetched from server artifact.

## Update 2.0
1. Migrate `flow.artifact()` to `flowrun.artifact()`
2. Create a `export_serialized_function()` to handle the getting serialized function into the dag_response rather than putting them all in `get_workflow()`
3. As mentioned by Kenny, the operation on `operator` is crossing the the abstraction barrier. Therefore, in the `api_client`, I eliminated any operation on `operator`  and put them into `workflow_dag_response`
4. In order to prevent reuse of `artifact` from `flowrun.arifact()`, I create a boolean parameter for Artifact(TableArtifact etc.) `from_flow_run`. The parameter defaults to false but can be set to true when creating the Artifact Object. Then, in `decorator.py`, when the operators and artifacts are wrapped, I check whether any input artifact `from_flow_run` is set to true. If so return a Exception.
5. I also create a test case to test whether we catch any input artifact from flow run. I basically publish a flow and fetch the artifact and reuse it for sentiment model. When feeding in the return artifact, we are expected an Exception.

## Update 3.0
After last review, several things have been changed:
1. the process of updating operator will be taking place in `dag.py` instead of `api_client.py`
2. the action of putting serialized function into the dag will be taking place in `flow.py: construct_flow_run()` instead of `api_client.py: get_workflow()`
3. fixing some misc and updating documentations for `_from_flow_run` parameter for InputArtifact

This version looks nicer since serveral lines of codes are taken away from the api_client file :)


integration_tests/sdk/flow_test.py
```diff
@@ -169,3 +170,22 @@ def test_refresh_flow(client):
         wait_for_flow_runs(client, flow.id(), num_runs=num_initial_runs + 1)
     finally:
         client.delete_flow(flow.id())
+
+
+@pytest.mark.publish
+def test_get_flow(client):
```
vsreekanti, line 176: Tiny nit here -- shouldn't this be `test_get_artifact_from_flow` or something similar?

integration_tests/sdk/flow_test.py
```diff
@@ -169,3 +170,22 @@ def test_refresh_flow(client):
...
+        config=db.config(table=generate_table_name(), update_mode=LoadUpdateMode.REPLACE)
+    )
+    flow = client.publish_flow(
+        name=generate_new_flow_name(),
+        artifacts=[output_artifact],
+        schedule=aqueduct.hourly(),
+    )
+    wait_for_flow_runs(client, flow.id(), num_runs=1)
+    artifact_return = flow.get(get_artifact_name())
+    assert artifact_return.name == get_artifact_name()
```
vsreekanti, line 190: Is there any other metadata returned that we can check other than the name of the artifact?

sdk/aqueduct/flow.py
```diff
@@ -153,3 +154,19 @@ def describe(self) -> None:
             )
         )
         print(json.dumps(self.list_runs(), sort_keys=False, indent=4))
+
+    def get(self, artifact_name: str) -> Artifact:
```
vsreekanti, line 158: `get()` feels a little ambiguous to me. Are we getting code, data, metadata? Feels like a generic `get()` method should support all of those. I would suggest doing something specific like either `flow.artifact(name)` or `flow.get_artifact(name)`. 

@kenxu95, any thoughts?

sdk/aqueduct/flow.py
```diff
@@ -153,3 +154,19 @@ def describe(self) -> None:
...
+        """Gets the Artifact from the flow based on artifact_name"""
+        resp = self._api_client.get_workflow(self._id)
+        assert (
+            len(resp.workflow_dag_results) > 0
+        ), "Every flow must have at least one run attached to it."
+        latest_result = resp.workflow_dag_results[-1]
```
vsreekanti, line 164: Are we assuming here that when we have a handle to a `flow`, it always points to the latest version @kenxu95?

vsreekanti: I guess I'm really asking how this interacts with versioning -- can we access older versions?

vsreekanti (commented):
@Fanjia-Yan, this mostly looks good, but I left a few comments and nits inline.

Let's also please make sure to title our pull requests clearly -- GitHub defaults to pulling from the branch name, but that's a little unhelpful/unclear in this case.

Thanks!

sdk/aqueduct/flow.py
kenxu95: I think `flow.artifact(name)` is consistent with how we usually do things.

sdk/aqueduct/flow.py
kenxu95: yeah, @Fanjia-Yan  this method shouldn't be on the flow, this should be on the flow run. A flow itself is actually dag-agnostic right now, since a dag is always associated with a particular run. You can also grab a run with `flow.latest()` or `flow.fetch()`

kenxu95 (changes requested):
So checking the artifact name is fine, but it's incomplete. We should also make sure that calling `artifact.get()` returns the same value as before! An artifact fetched from a flow should be functionally indistinguishable from an artifact that was locally created.

I think thats a non-trivial follow up since you'll have to alter the `get_workflow` route to also return the serialized functions.

Fanjia-Yan:
Many changes to this PR after last review
1. Complete the `get_workflow()` so that the `workflowresponse` also has serialized function attached to each operator. To get the serialized function, I route "/api/function/%s/export" and request by operator_id
2. Modify the previous `flow.get()` into `flow.artifact()`. In `flow.artifact()`, we first fetch the `artifact.Artifact` object with matching `artifact_name` and convert it to `generic_artifact.Artifact` so that we can call `.get()`.
3. For the integration test. I fetch the artifact using `flow.artifact(output_artifact.name())`. And check for name equality and dataframe equality between locally created artifact and fetched from server artifact.

sdk/aqueduct/api_client.py
```diff
@@ -274,7 +275,16 @@ def get_workflow(self, flow_id: str) -> GetWorkflowResponse:
...
-        return GetWorkflowResponse(**resp.json())
+
+        workflow_response = GetWorkflowResponse(**resp.json())
+        for work_flow_dag in list(workflow_response.workflow_dags.values()):
+            for operator in list(work_flow_dag.operators.values()):
+                operator_url = self._construct_full_url(
+                    self.EXPORT_FUNCTION_ROUTE % str(operator.id), self.use_https
+                )
+                operator_resp = requests.get(operator_url, headers=headers)
+                operator.change_file(operator_resp.content)
```
kenxu95, line 286: instead of `change_file()` on the operator, I'd like to keep our current abstraction where the dag class is the only thing that can modify underlying operator and artifact objects. Having fewer places where things can modify the dag will make things easier to maintain + us to have fewer bugs.

Would you be able to use `dag.update_operator_spec()` instead? You can probably write a helper to construct the new operator spec from the old one + serialized contents.

sdk/aqueduct/api_client.py
```diff
@@ -274,7 +275,16 @@ def get_workflow(self, flow_id: str) -> GetWorkflowResponse:
         url = self._construct_full_url(self.GET_WORKFLOW_ROUTE_TEMPLATE % flow_id, self.use_https)
         resp = requests.get(url, headers=headers)
         utils.raise_errors(resp)
-        return GetWorkflowResponse(**resp.json())
+
+        workflow_response = GetWorkflowResponse(**resp.json())
+        for work_flow_dag in list(workflow_response.workflow_dags.values()):
+            for operator in list(work_flow_dag.operators.values()):
```
kenxu95, line 281: typo: `work_flow`

Also, for better abstraction, the API client should work at a lower layer. It shouldn't know anything about our dag operators, it should only understand backend response objects. I would recommend exposing a separate `export` function in the APIClient and moving the dag updating logic up into the caller.

sdk/aqueduct/flow.py
```diff
@@ -153,3 +158,28 @@ def describe(self) -> None:
...
+
+        if artifact_DAG is None:
+            raise ArtifactNotFoundException("The artifact name provided does not exist.")
+        if artifact_DAG.spec.table is not None:
+            return TableArtifact(flow_run._api_client, flow_run._dag, artifact_DAG.id)
+        if artifact_DAG.spec.float is not None:
+            return MetricArtifact(flow_run._api_client, flow_run._dag, artifact_DAG.id)
+        if artifact_DAG.spec.bool is not None:
+            return CheckArtifact(flow_run._api_client, flow_run._dag, artifact_DAG.id)
+        if artifact_DAG.spec.jsonable is not None:
```
kenxu95, line 182: nit: lets make all these `elif`?

sdk/aqueduct/flow.py
```diff
@@ -153,3 +158,28 @@ def describe(self) -> None:
             )
         )
         print(json.dumps(self.list_runs(), sort_keys=False, indent=4))
+
+    def artifact(
+        self, artifact_name: str
+    ) -> Union[TableArtifact, MetricArtifact, CheckArtifact, ParamArtifact, None]:
+        """Gets the Artifact from the flow based on artifact_name"""
+        flow_run = self.latest()
```
kenxu95, line 166: I think this `artifact()` method should be on the FlowRun object, not the flow. Always fetching from the latest run seems like an unnecessary restriction. A user should be able to fetch any historical run, and realize the artifact from that historical run, using latest data.

kenxu95 (changes requested):
This is awesome! I have a number of comments but thanks for getting this working.

There are two other things I thought of while reviewing:
1) Did you want to rename the generic artifact to `GenericArtifact`? We can probably just do it (here or in a separate PR)
2) A returned artifact currently cannot be used to construct a new dag right? Because the underlying dag objects are different. Could we add a defensive check in there for now, where if you attempt to use a returned artifact in new computation, we'd throw an error? This might be an interesting follow up :)

sdk/aqueduct/api_client.py
Fanjia-Yan: @kenxu95 Since `get_workflow` returns a class `WorkflowDagResponse`, the serialize function is inserted into `WorkflowDagResponse` rather than the `dag` class. Therefore I believe the update_operator_spec() should be under  `WorkflowDagResponse`?

Fanjia-Yan:
1. Migrate `flow.artifact()` to `flowrun.artifact()`
2. Create a `export_serialized_function()` to handle the getting serialized function into the dag_response rather than putting them all in `get_workflow()`
3. As mentioned by Kenny, the operation on `operator` is crossing the the abstraction barrier. Therefore, in the `api_client`, I eliminated any operation on `operator`  and put them into `workflow_dag_response`
4. In order to prevent reuse of `artifact` from `flowrun.arifact()`, I create a boolean parameter for Artifact(TableArtifact etc.) `from_flow_run`. The parameter defaults to false but can be set to true when creating the Artifact Object. Then, in `decorator.py`, when the operators and artifacts are wrapped, I check whether any input artifact `from_flow_run` is set to true. If so return a Exception.
5. I also create a test case to test whether we catch any input artifact from flow run. I basically publish a flow and fetch the artifact and reuse it for sentiment model. When feeding in the return artifact, we are expected an Exception.

sdk/aqueduct/api_client.py
```diff
@@ -325,3 +329,13 @@ def get_node_positions(
...
+
+    def export_serialized_function(self, work_flow_response: GetWorkflowResponse) -> None:
+        headers = utils.generate_auth_headers(self.api_key)
+        for work_flow_dag in list(work_flow_response.workflow_dags.values()):
+            for operator in list(work_flow_dag.operators.values()):
+                operator_url = self._construct_full_url(
+                    self.EXPORT_FUNCTION_ROUTE % str(operator.id), self.use_https
+                )
+                operator_resp = requests.get(operator_url, headers=headers)
+                work_flow_dag.update_operator_spec(operator, operator_resp)
```
kenxu95, line 341: ok this is better, but I was really saying this APIClient layer should be as thin as possible - it should literally just take in a single operator id and return the serialized function for that operator. The double-for loop belongs in the caller.

Question: does an operator always only correspond to a single workflow dag? Is there a case where an operator corresponds to multiple workflow dags on the backend. cc @cw75

sdk/aqueduct/check_artifact.py
```diff
@@ -33,14 +33,12 @@ class CheckArtifact(Artifact):
...
-        self,
-        api_client: APIClient,
-        dag: DAG,
-        artifact_id: uuid.UUID,
+        self, api_client: APIClient, dag: DAG, artifact_id: uuid.UUID, from_flow_run: bool
     ):
         self._api_client = api_client
         self._dag = dag
         self._artifact_id = artifact_id
+        self._from_flow_run = from_flow_run
```
kenxu95, line 41: Can you write a comment here describing what this variable represents? Its not super clear from the name in what scenarios it is `_from_flow_run` and what scenarios it isn't.

sdk/aqueduct/flow_run.py
```diff
@@ -64,6 +68,30 @@ def describe(self) -> None:
             assert param_op.spec.param is not None, "Artifact is not a parameter."
             print("* " + param_op.name + ": " + param_op.spec.param.val)
 
+    def artifact(
+        self, artifact_name: str
+    ) -> Union[TableArtifact, MetricArtifact, CheckArtifact, ParamArtifact, None]:
```
kenxu95, line 73: nit: should be Optional[Union[...]], get rid of the None

sdk/aqueduct/flow_run.py
```diff
@@ -64,6 +68,30 @@ def describe(self) -> None:
             assert param_op.spec.param is not None, "Artifact is not a parameter."
             print("* " + param_op.name + ": " + param_op.spec.param.val)
 
+    def artifact(
+        self, artifact_name: str
```
kenxu95, line 72: super nit: just "name" is fine. The function is called `artifact()` already.

sdk/aqueduct/flow_run.py
```diff
@@ -64,6 +68,30 @@ def describe(self) -> None:
...
+    ) -> Union[TableArtifact, MetricArtifact, CheckArtifact, ParamArtifact, None]:
+        """Gets the Artifact from the flow based on artifact_name"""
+        flow_run_dag = self._dag
+
+        for artifact_from_workflow in list(flow_run_dag.artifacts.values()):
+            if artifact_from_workflow.name == artifact_name:
+                assert isinstance(artifact_from_workflow, Artifact)
+                artifact_DAG = artifact_from_workflow
+
+        if artifact_DAG is None:
```
kenxu95, line 82: nit: uncapitalize DAG here, since we don't name like this anywhere else.

sdk/aqueduct/flow_run.py
```diff
@@ -64,6 +68,30 @@ def describe(self) -> None:
...
+    def artifact(
+        self, artifact_name: str
+    ) -> Union[TableArtifact, MetricArtifact, CheckArtifact, ParamArtifact, None]:
+        """Gets the Artifact from the flow based on artifact_name"""
+        flow_run_dag = self._dag
+
+        for artifact_from_workflow in list(flow_run_dag.artifacts.values()):
+            if artifact_from_workflow.name == artifact_name:
+                assert isinstance(artifact_from_workflow, Artifact)
+                artifact_DAG = artifact_from_workflow
```
kenxu95, line 80: this fetching of the artifact from the dag should be a method on the dag class (see dag.py). You can either augment `must_get_artifact()` or write a separate method called `get_artifact_by_name()` or something - see how all the other functions on this class do it. You'll probably have to define an `artifact_by_name` dictionary to do the lookup (similar to how `operator_by_name` works).

It should be simple here as:
```
artifact = self._dag.must_get_artifact(name)
```

sdk/aqueduct/flow_run.py
```diff
@@ -64,6 +68,30 @@ def describe(self) -> None:
...
+            raise ArtifactNotFoundException("The artifact name provided does not exist.")
+        elif artifact_DAG.spec.table is not None:
+            return TableArtifact(self._api_client, self._dag, artifact_DAG.id, True)
+        elif artifact_DAG.spec.float is not None:
+            return MetricArtifact(self._api_client, self._dag, artifact_DAG.id, True)
+        elif artifact_DAG.spec.bool is not None:
+            return CheckArtifact(self._api_client, self._dag, artifact_DAG.id, True)
+        elif artifact_DAG.spec.jsonable is not None:
+            return ParamArtifact(self._api_client, self._dag, artifact_DAG.id, True)
+
```
kenxu95, line 92: nit: better to use `get_artifact_type()` in these if-else statements, since its more readable and catches any non-valid types.

sdk/aqueduct/generic_artifact.py
```diff
@@ -9,6 +9,7 @@ class Artifact(ABC):
 
     _artifact_id: uuid.UUID
     _dag: DAG
+    _from_flow_run: bool
```
kenxu95, line 12: why is this necessary?

sdk/aqueduct/responses.py
```diff
@@ -151,6 +152,10 @@ class WorkflowDagResponse(BaseModel):
     operators: Dict[str, Operator]
     artifacts: Dict[str, Artifact]
 
+    def update_operator_spec(self, operator: Operator, resp: Response) -> None:
+        if operator in list(self.operators.values()):
+            operator.change_file(resp.content)
+
```
kenxu95, line 158: wait no, I thought we agreed on putting this write on the dag class?

sdk/aqueduct/operators.py
```diff
@@ -176,6 +176,16 @@ def file(self) -> Optional[bytes]:
...
 
+    def change_file(self, serialized_function: bytes) -> None:
+        if self.spec.function:
+            self.spec.function.file = serialized_function
+        if self.spec.metric:
+            self.spec.metric.function.file = serialized_function
+        if self.spec.check:
+            self.spec.check.function.file = serialized_function
+
+        return None
```
kenxu95, line 187: I thought we agreed to use `update_operator_spec()` in dag.py? Not this.

sdk/aqueduct/decorator.py
```diff
@@ -80,6 +80,10 @@ def wrap_spec(
     if len(input_artifacts) < 1:
         raise Exception("Transformation requires at least one input artifact.")
 
+    for artifact in input_artifacts:
+        if artifact._from_flow_run:
+            raise Exception("Artifact fetched from flow run can not be reused for computation")
```
kenxu95, line 85: This should be a `UserAction...Exception`. Look at the specific errors we have defined in `errors.py`

kenxu95 (changes requested):
It's getting better! Thanks for fixing the incompatible dag issue. There are a lot of code quality and abstraction issues still here so I still need to block. I think maybe there was miscommunication about what abstractions are present in the SDK? Especially around how we update the dag.

Fanjia-Yan:
After last review, several things have been changed:
1. the process of updating operator will be taking place in `dag.py` instead of `api_client.py`
2. the action of putting serialized function will be taking place in `flow.py: construct_flow_run()` instead of `api_client.py: get_workflow()`
3. fixing some misc and updating documentations for `_from_flow_run` parameter for InputArtifact
This version looks nicer since most codes are taken away from the api_client file :)

sdk/aqueduct/flow_run.py
```diff
@@ -64,6 +68,26 @@ def describe(self) -> None:
...
 
+    def artifact(
+        self, name: str
+    ) -> Optional[Union[TableArtifact, MetricArtifact, CheckArtifact, ParamArtifact]]:
+        """Gets the Artifact from the flow based on artifact_name"""
+        flow_run_dag = self._dag
+        artifact_from_dag = flow_run_dag.get_artifacts_by_name(name)
+
+        if artifact_from_dag is None:
+            raise ArtifactNotFoundException("The artifact name provided does not exist.")
```
kenxu95, line 79: I wonder if it might be nicer for the user if we return None if the artifact doesn't exist? Maybe they just want to check for existence of an artifact, and its a bit less annoying to check for None than it is to catch an error.

sdk/aqueduct/flow_run.py
```diff
@@ -64,6 +68,26 @@ def describe(self) -> None:
...
+        elif get_artifact_type(artifact_from_dag) is ArtifactType.TABLE:
+            return TableArtifact(self._api_client, self._dag, artifact_from_dag.id, True)
+        elif get_artifact_type(artifact_from_dag) is ArtifactType.NUMBER:
+            return MetricArtifact(self._api_client, self._dag, artifact_from_dag.id, True)
+        elif get_artifact_type(artifact_from_dag) is ArtifactType.BOOL:
+            return CheckArtifact(self._api_client, self._dag, artifact_from_dag.id, True)
+        elif get_artifact_type(artifact_from_dag) is ArtifactType.PARAM:
+            return ParamArtifact(self._api_client, self._dag, artifact_from_dag.id, True)
+
+        return None
```
kenxu95, line 89: Here I actually think we should raise an exception (aka fail very loudly), since this means something went wrong in our system. Is there an Internal error we can throw?

sdk/aqueduct/flow_run.py
kenxu95: If you add that, could you also put that in the function comment string?

sdk/aqueduct/operators.py
```diff
@@ -176,6 +176,16 @@ def file(self) -> Optional[bytes]:
...
 
+    def update_serialized_function(self, serialized_function: bytes) -> None:
+        if self.spec.function:
+            self.spec.function.file = serialized_function
+        if self.spec.metric:
+            self.spec.metric.function.file = serialized_function
+        if self.spec.check:
+            self.spec.check.function.file = serialized_function
+
+        return None
```
kenxu95, line 187: nit: no need to return None here. -> None is just a function without a return.

sdk/aqueduct/flow_run.py
```diff
@@ -64,6 +68,26 @@ def describe(self) -> None:
...
+        if artifact_from_dag is None:
+            raise ArtifactNotFoundException("The artifact name provided does not exist.")
+        elif get_artifact_type(artifact_from_dag) is ArtifactType.TABLE:
+            return TableArtifact(self._api_client, self._dag, artifact_from_dag.id, True)
+        elif get_artifact_type(artifact_from_dag) is ArtifactType.NUMBER:
+            return MetricArtifact(self._api_client, self._dag, artifact_from_dag.id, True)
+        elif get_artifact_type(artifact_from_dag) is ArtifactType.BOOL:
+            return CheckArtifact(self._api_client, self._dag, artifact_from_dag.id, True)
+        elif get_artifact_type(artifact_from_dag) is ArtifactType.PARAM:
+            return ParamArtifact(self._api_client, self._dag, artifact_from_dag.id, True)
```
kenxu95, line 87: nit: for each of the artifact constructors, "True" is not a super enlightening argument value, so I typically do eg:
```
ParamArtifact(self._api_client, self._dag, artifact_from_dag.id, from_flow_run=True)
```
especially if its one of the last few arguments.

kenxu95 (approved):
Awesome! Thanks for pushing this through 👍 Left a few stylistic/cleanup comments but looks good to me. I think @vsreekanti also needs to unblock.

vsreekanti:
Hey @Fanjia-Yan, Kenny knows this codebase much better than me, so if he's signed off on the PR, you should be good to go. 🙂 No need to wait for my approval here.

sdk/aqueduct/flow_run.py
Fanjia-Yan: @kenxu95 I think this error is really unlikely trigger as an artifact falls in either the four categories. However, if there will be an error, I think it is likely to be either "InvalidArtifactError"(does not exist yet in error.py) or "UnprocessableEntityError"(Exception raised for errors occured when the Aqueduct system fails to process certain inputs.)

Fanjia-Yan: Or maybe we can throw the general AqueductError?

sdk/aqueduct/flow_run.py
Fanjia-Yan: I think that's a good idea!

sdk/aqueduct/flow_run.py
kenxu95: Let's throw an `InternalAqueductError`. The reason that this needs to be internal is because its definitely something thats gone wrong in our system, and is not the user's fault.

This error should never happen, as you pointed out, but it's important in our code to make sure that we cover all assumptions. If any assumption breaks, we should also fail loudly so that the person that broke the assumption can fix the issue. In this case, the assumption is that the artifact is one of four types, but this can easily be changed (imagine someone adds a new type on the backend, but doesn't find this function and doesn't update it).

PR #158, opened 2022-06-29
[SDK] Adds consistent parameter validation

kenxu95:
Both get() and trigger() allow for execution of flows with different parameters. This pipes any user-defined dictionary of overriding parameters through the `check_overwriting_parameters_are_valid()` helper, which validates that the new parameter values in the dictionary can be used on the dag.

get() uses the dag of the artifact. trigger() now uses the latest run of the flow. If there isn't a run of the flow yet, it'll throw an exception.

ENG-1144

- [x ] I have performed a self-review of my code.
- [ N/A] All features on the UI continue to work correctly.

### DIFF
--- a/sdk/aqueduct/aqueduct_client.py
+++ b/sdk/aqueduct/aqueduct_client.py
@@ -370,6 +370,9 @@ class Client:
         if parameters is not None:
             flow = self.flow(flow_id)
             runs = flow.list_runs(limit=1)
+
+            # NOTE: this is a defense check against triggering runs that haven't run yet.
+            # We may want to revisit this in the future if more nuanced constraints are necessary.
             if len(runs) == 0:
                 raise InvalidUserActionException(
                     "Cannot trigger a workflow that hasn't already run at least once."
--- a/sdk/aqueduct/aqueduct_client.py
+++ b/sdk/aqueduct/aqueduct_client.py
@@ -13,7 +13,7 @@ from .dag import (
     apply_deltas_to_dag,
     SubgraphDAGDelta,
     Metadata,
-    AddOrReplaceOperatorDelta,
+    AddOrReplaceOperatorDelta, check_overwriting_parameters_are_valid,
 )
 from .enums import RelationalDBServices, ServiceType
 from .error import (
@@ -365,8 +365,9 @@ class Client:
             InternalServerError:
                 An unexpected error occurred within the Aqueduct cluster.
         """
-        # TODO(ENG-1144): If there the provided parameters dict is not valid, throw an error
-        #  earlier, before getting to execution.
+        if parameters is not None:
+            flow = self.flow(flow_id)
+            check_overwriting_parameters_are_valid(flow._latest_dag_from_flow(), parameters)
 
         serialized_params = None
         if parameters is not None:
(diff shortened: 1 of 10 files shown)


sdk/aqueduct/aqueduct_client.py
```diff
@@ -365,8 +367,14 @@ def trigger(
...
         """
-        # TODO(ENG-1144): If there the provided parameters dict is not valid, throw an error
-        #  earlier, before getting to execution.
+        if parameters is not None:
+            flow = self.flow(flow_id)
+            runs = flow.list_runs(limit=1)
+            if len(runs) == 0:
+                raise InvalidUserActionException(
+                    "Cannot trigger a workflow that hasn't already run at least once."
+                )
```
kenxu95, line 376: @saurav-c @vsreekanti I could also try and fetch the latest workflow dag here, even if there weren't any workflow runs, but the implementation of that was a little messy. It seemed logical to simply disallow `trigger()` until at least one run exists, but lemme know if thats not good.

sdk/aqueduct/dag.py
```diff
@@ -210,10 +217,20 @@ def add_artifacts(self, artifacts: List[Artifact]) -> None:
...
+        both that the operator exists, and that the spec type will be unchanged.
+        """
+        assert name in self.operator_by_name, "Operator %s does not exist." % name
+        op = self.operator_by_name[name]
+        assert get_operator_type(op) == get_operator_type_from_spec(
+            spec
+        ), "New spec has a different type."
+
+        self.operators[str(op.id)].spec = spec
+        self.operator_by_name[op.name].spec = spec
```
kenxu95, line 233: This seemed safer than the blind replace of operators we currently have. Plus, we can add additional validation on the operation.

kenxu95 (commented):
Designating @likawind as lead reviewer but tagged the other two in a question.

sdk/aqueduct/aqueduct_client.py
vsreekanti: I'm actually not 100% sure how triggering works with Airflow -- @saurav-c, is there an API hook to trigger a workflow that has been registered, or can it only be on a schedule?

I'm asking because I wonder if it makes sense to have system-specific trigger checks.

saurav-c: @vsreekanti There is an API hook to trigger an Airflow dag if its been registered.

sdk/aqueduct/dag.py
```diff
@@ -454,40 +511,17 @@ def __init__(
     def apply(self, dag: DAG) -> None:
         if self.parameters is None:
             return
-
-        if any(not isinstance(name, str) for name in self.parameters):
-            raise InvalidUserArgumentException("Parameters must be keyed by strings.")
+        check_overwriting_parameters_are_valid(dag, self.parameters)
```
likawind, line 514: maybe `validate_overwriting_parameters` ?

likawind (approved):
Great! I think this type of checking could be generalized to ensure all operator inputs are properly typed (e.g. you shouldn't pass a string params to an df function), but that's not immediate.

sdk/aqueduct/dag.py
likawind: What's the main downside of direct replacement?

kenxu95: It's just bit uglier from the callers side, since editing an operator always requirements a fetch and replace. With this we can just construct the new spec from scratch. Replacing an operator fully might be better in the long term though, not sure.

PR #159, opened 2022-06-29
[Backend] Consume error and user logs from operator executors

likawind:
This PR enables backend to clearly distinguish user vs system errors. It depends on #144 where we clearly redirect user logs and generate user errors from python executors. This PR mainly changed existing codebase to support this more structured error object, and port the object directly to client to handle. We also removed the hack that tries to distinguish user vs sys error by checking if the metadata exists.
ENG-1249


src/golang/lib/collections/shared/types.go
```diff
@@ -14,3 +26,36 @@ const (
 	PendingExecutionStatus   ExecutionStatus = "pending"
 	UnknownExecutionStatus   ExecutionStatus = "unknown"
 )
+
+type FailureType int64
+
+const (
```
vsreekanti, line 32: Same question from the UI PR about why we're separating this into another enum.

vsreekanti (commented):
I'll let someone who knows the codebase better do a more thorough review, but I dropped in a few questions about interface clarity.

src/golang/cmd/server/queries/queries.go
```diff
@@ -35,18 +34,19 @@ type ArtifactResponse struct {
...
-	ArtifactId          uuid.UUID                `db:"artifact_id" json:"artifact_id"`
-	WorkflowDagResultId uuid.UUID                `db:"workflow_dag_result_id" json:"workflow_dag_result_id"`
-	Status              shared.ExecutionStatus   `db:"status" json:"status"`
-	Name                string                   `db:"name" json:"name"`
-	Metadata            operator_result.Metadata `db:"metadata" json:"metadata"`
+	ArtifactId          uuid.UUID              `db:"artifact_id" json:"artifact_id"`
+	WorkflowDagResultId uuid.UUID              `db:"workflow_dag_result_id" json:"workflow_dag_result_id"`
+	Status              shared.ExecutionStatus `db:"status" json:"status"`
+	Name                string                 `db:"name" json:"name"`
+	Metadata            shared.ExecutionLogs   `db:"metadata" json:"metadata"`
```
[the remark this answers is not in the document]
likawind: Combining with your comment below, how about `State: ExecutionState`? cc @kenxu95

src/golang/cmd/server/handler/connect_integration.go
```diff
@@ -199,7 +198,7 @@ func ValidateConfig(
 	}
 
 	// Authentication failed, so we need to fetch the error message from storage
-	var metadata operator_result.Metadata
+	var metadata shared.ExecutionState
```
kenxu95, line 201: lets call this var  `execState` then?

src/golang/cmd/server/handler/connect_integration.go
```diff
@@ -209,5 +208,9 @@ func ValidateConfig(
 		return http.StatusInternalServerError, errors.Wrap(err, "Unable to connect integration.")
 	}
 
-	return http.StatusBadRequest, errors.Newf("Unable to authenticate credentials: %v", metadata.Error)
+	errMsg := ""
+	if metadata.Error != nil {
+		errMsg = metadata.Error.Context
```
kenxu95, line 213: Might be better to throw a different internal server error here, since `execState.Error` being nil means something very wrong has happened right? Our assumption here is that the execution has definitely failed. An empty error string may be less helpful

So you can then directly return:
```
return http.StatusBadRequest, errors.Newf("Unable to auth...", execState.Error.Context)
```

src/golang/cmd/server/handler/get_artifact_versions.go
```diff
@@ -45,9 +44,9 @@ type artifactVersion struct {
 }
 
 type CheckResult struct {
-	Name     string                   `json:"name"`
-	Status   shared.ExecutionStatus   `json:"status"`
-	Metadata operator_result.Metadata `json:"metadata"`
+	Name     string                 `json:"name"`
+	Status   shared.ExecutionStatus `json:"status"`
+	Metadata shared.ExecutionState  `json:"metadata"`
```
kenxu95, line 49: I assume changing `Metadata` to `ExecState` would require a migration? Metadata is fine for now then.

src/golang/cmd/server/handler/preview.go
```diff
@@ -192,36 +193,31 @@ func (h *PreviewHandler) Perform(ctx context.Context, interfaceArgs interface{})
...
 		err := utils.ReadFromStorage(ctx, storageConfig, path, &operatorMetadata)
 		if err != nil {
-			responses[id] = previewOperatorResponse{
-				Status: shared.FailedExecutionStatus,
-				ErrMsg: "Unable to retrieve metadata for this operator. The workflow might have failed before executing this operator.",
+			responses[id] = shared.ExecutionState{
+				Status:      shared.FailedExecutionStatus,
+				FailureType: shared.SystemFailure,
+				Error: &shared.Error{
+					Context: fmt.Sprintf("%v", err),
```
kenxu95, line 213: tip: I think `err.Error()` returns the stringified error btw.

src/golang/lib/collections/operator_result/constants.go
```diff
@@ -12,7 +12,7 @@ const (
 	WorkflowDagResultIdColumn = "workflow_dag_result_id"
 	OperatorIdColumn          = "operator_id"
 	StatusColumn              = "status"
-	MetadataColumn            = "metadata"
+	StateColumn               = "statue"
```
kenxu95, line 15: type: statue

kenxu95: can we call it `exec_state`?

src/golang/lib/workflow/orchestrator/orchestrator.go
```diff
@@ -28,16 +28,13 @@ const (
...
-func operatorExecutionError(operators map[uuid.UUID]operator.Operator, operatorId uuid.UUID) error {
-	op, ok := operators[operatorId]
-	if !ok {
-		return errors.Newf("Error during exeuction, invalid operator ID %s", operatorId)
-	}
-
-	return errors.Newf("Error during execution, operator ID %s, name %s", operatorId, op.Name)
-}
+var (
+	ErrIncorrectOperatorsScheduled = errors.New("Incorrect number of operators scheduled.")
```
kenxu95, line 32: How does one schedule an incorrect number of operators? Is this an internal error and can we add a comment here?

src/golang/lib/workflow/orchestrator/orchestrator.go
```diff
@@ -28,16 +28,13 @@ const (
...
+	ErrOpExecSystemFailure         = errors.New("Operator execution failed due to system error.")
+	ErrOpExecBlockingUserFailure   = errors.New("Operator execution failed due to user error.")
```
kenxu95, line 34: What are non-blocking user failures?

kenxu95 (changes requested):
Nice! I just have a bunch of smallish comments/questions.

src/golang/cmd/server/handler/connect_integration.go
likawind: Good catch, we should return something more meaningful if the error is nil

src/golang/cmd/server/handler/get_artifact_versions.go
likawind: in either way we would need a migration since the struct is changed. The migration is not tied to this line though, I think check results is not persisted in DB and it's just a response struct.

src/golang/lib/collections/operator_result/constants.go
likawind: oh wtf am I doing here

src/golang/lib/workflow/orchestrator/orchestrator.go
likawind: Check warning is one type

src/golang/cmd/server/handler/create_table.go
```diff
@@ -185,5 +184,9 @@ func CreateTable(ctx context.Context, args *CreateTableArgs, contentPath string,
 		return http.StatusInternalServerError, errors.Wrap(err, "Unable to create table.")
 	}
 
-	return http.StatusBadRequest, errors.Newf("Unable to create table: %v", metadata.Error)
+	errMsg := ""
+	if metadata.Error != nil {
```
kenxu95, line 188: nit: shouldn't we raise an internal exception if `metadata.Error == nil` here? It's expected that we have an error at this point right.

kenxu95 (approved):
cool! one more nit. Also there might be a few unresolved prev comments floating around?

PR #160, opened 2022-06-29
[Sdk] Distinghuish user vs system errors and produce better error message.

likawind:
This PR adds SDK changes to consume more structured error object introduced in #144 and processed by backend in #159 . Rather than try to figure out the error type from the client side, we now completely rely on the object to render the message.
 
ENG-1249

## Demo
missing table in sql:
Screen Shot 2022-06-28 at 5 15 39 PM

wrong syntax:
Screen Shot 2022-06-28 at 5 15 58 PM

bad function:
Screen Shot 2022-06-28 at 11 07 13 PM


vsreekanti (commented):
This mostly looks good to me pending changes on the other PRs, but I'll wait for someone else to review as well since I haven't spent much time in the SDK code.

sdk/aqueduct/enums.py
```diff
@@ -81,11 +81,18 @@ class RelationalDBServices(str, Enum, metaclass=MetaEnum):
...
     SUCCEEDED = "succeeded"
     FAILED = "failed"
     PENDING = "pending"
 
 
+class FailureType(Enum, metaclass=MetaEnum):
+    SUCCESS = 0
+    SYSTEM = 1
+    USER = 2
+
```
kenxu95, line 94: I had some questions about the enum values here, which I wrote about in the prereq PR reviews.

sdk/aqueduct/responses.py
```diff
@@ -4,11 +4,32 @@
...
+    def is_empty(self) -> bool:
+        return self.stdout == "" and self.stderr == ""
+
+    def __str__(self) -> str:
+        return f"""stdout:
+{self.stdout}
+--------------------------
+stderr:
+{self.stderr}
+"""
```
kenxu95, line 25: I've been using `textwrap.dedent(..)` to avoid this really weird indentation. See `FlowRun.describe()`.

sdk/aqueduct/responses.py
```diff
@@ -22,8 +43,10 @@ class OperatorResult(BaseModel):
             Only set if this operator represents a unit test on an artifact.
     """
 
-    logs: Optional[Dict[str, str]]
-    err_msg: Optional[str]
+    user_logs: Optional[Logs] = None
+    error: Optional[Error] = None
+    status: ExecutionStatus = ExecutionStatus.UNKNOWN
+    failure_type: FailureType = FailureType.SUCCESS
```
kenxu95, line 49: why do these fields need default values? Aren't they only fetched from the backend? The backend should supply all the appropriate values, and if they sometimes don't, that field should be Optional.

likawind: I think it makes sense to put `FailureType` optional as we discussed in #144 . I think it make sense to default `status` to `Unknown` which is more explicit than `status = None`

kenxu95 (approved):

PR #161, opened 2022-06-29
[UI] Handle more structured error object

likawind:
This PR updated UI component to handle more structured error object introduced in #144 and processed by backend in #159 . It doesn't make much user-facing improvement comparing to current experience, but it enabled us to distinguish and handle user vs system errors differently in the future. Additionally, we are able to remove a few checkings based on error message string and render error message completely based on the object.

ENG-1249

## Demo
Screen Shot 2022-06-28 at 10 25 03 PM

- [ ] I have manually run the integration tests and they are passing.


src/ui/common/src/utils/shared.ts
```diff
@@ -13,11 +13,18 @@ export type LoadingStatus = {
 };
 
 export enum ExecutionStatus {
+  Unknown = 'unknown',
   Succeeded = 'succeeded',
   Failed = 'failed',
   Pending = 'pending',
 }
 
+export enum FailureReason {
```
vsreekanti, line 22: Couple tiny nits here -- can we make the default the success type rather than a system error.

Is there a reason that we want to maintain this separately from the execution status? It seems a little confusing to me -- under execution status, would we rather have pending, succeeded, system error, user error, and (soon) warning? That seems more natural to me rather than splitting the context. Thoughts?

src/ui/common/src/utils/operators.ts
```diff
@@ -135,9 +135,10 @@ export function normalizeOperator(op): Operator {
 }
 
 export type GetOperatorResultResponse = {
-  status: ExecutionStatus;
-  error: string;
-  logs: { [name: string]: string };
+  code: ExecutionStatus;
```
vsreekanti, line 138: Is there a reason we renamed this from `status` to `code`? The word code in my mind is like an error code (e.g., when a process does `exit(1)` or `exit(127)`. We're just using strings here, so status seems fine, but let me know if I'm missing something.

vsreekanti (approved):
This mostly looks good, @likawind. I left a couple questions here about the interface which tie into my comments on the backend.

src/ui/common/src/utils/operators.ts
likawind: Yeah status sounds good :)

src/ui/common/src/utils/shared.ts
likawind: I've been moving back and forth on this. I do find the current implementation a bit easier to handle, the main reason is that we are duplicating this status handling in a few places, including executor, go, sdk, and frontend.

The most important use of `ExecutionStatus` enum is to determine if the execution failed by checking `status == ExecutionStatus.Failed`. If we merge the status with failure reason, that means all 4 places need to implement some helper like `isFailed(status)` and check something like `status in [ExecutionStatus.UserFailure, ExecutionStatus.SystemFailure]` which is harder to maintain if we expand on failure types.

I think to address the confusion we could use 'NotAvailable' or 'Unknown', instead of 'NoFailure'. In this way it's more clear that this field is only useful in case of failure, and success / fail check should be done based on the status field.

vsreekanti: That's a fair point. In that case, let's rename this slightly to `FailureType`, and we can leave it as is. I think we can just call the default value "Success" because it's clear that Success is a non-existent type of failure.

PR #176, opened 2022-06-30
ENG 1297: Set up build to package UI for sagemaker

vsreekanti:

SageMaker doesn't expose ports directly through it's Notebook instances but rather puts them behind `proxy/{portNumber}`. In order to make our UI work on SageMaker, we need to: 1) change our static assets to load from `/proxy/8080/dist` rather than just `/dist`; 2) change all of our API calls and links within the UI to have `/proxy/8080` prepended.

1. In order to do 1, I modified `package.json` to use the `--dist-dir` and the `--public-url` flags. `npm run build` generates out default build (now in `dist/default`), and `npm run build-sagemaker` generates a SageMaker-specific build in (in `dist/sagemaker`).
2. In order to do 2, we added `getPathPrefix` helper that looks at `window.location.pathname` and uses a regex to determine whether we're running behind `/proxy/{portNumber}` or not -- if we are, `getPathPrefix` returns that prefix, otherwise it returns an empty string.

Note that this *will not* work with any ports other than 8080 -- that is a limitation of hardcoding the `--public-url` flag at runtime.

As a follow-on to this, we need to modify the CLI to detect whether we're running on SageMaker and download the correct UI build at install time.

ENG-1297

### DIFF
--- a/src/ui/Makefile
+++ b/src/ui/Makefile
@@ -24,10 +24,11 @@ dist:
 	# don't need to specify an address when making a request.
 	cd app && echo "SERVER_ADDRESS=" > .env
 	cd app && npm run build
+	
+	# This is to ensure that the server address is /proxy/8080 when running on
+	# SageMaker. We do this because SageMaker puts everything behind a proxy that
+	# we have to sub-route everything behind.
+	cd app && echo "SERVER_ADDRESS=/proxy/8080" > .env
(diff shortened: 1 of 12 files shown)


cw75 (approved):
LGTM! @agiron123 appreciate another pair of eyes from you.

agiron123 (approved):

PR #182, opened 2022-07-01
Adds periodic integration tests

hsubbaraj-spiral:
Adds a workflow for periodic integration tests that runs across params of [python versions, concurrency]
https://linear.app/aqueducthq/issue/ENG-1300/enable-periodic-full-integration-tests-on-open-source-repo
- [n/a] I have included a small demo of the changes. For the UI, this would be a screenshot or a Loom video.
- [n/a] If this is a new feature, I have added unit tests and integration tests.
- [n/a] All features on the UI continue to work correctly.

### DIFF
--- a/.github/workflows/periodic-integration-tests.yml
+++ b/.github/workflows/periodic-integration-tests.yml
@@ -4,7 +4,6 @@ on:
   schedule:
      - cron: '0 15 * * 1-5' # Run at 7AM PST on every weekday
   workflow_dispatch:
-  pull_request:
 
 jobs:
   run-tests:
--- a/.github/workflows/periodic-integration-tests.yml
+++ b/.github/workflows/periodic-integration-tests.yml
@@ -1,0 +1,90 @@
+name: Periodic Integration Tests
+
+on:
+  schedule:
+     - cron: '0 15 * * 1-5' # Run at 7AM PST on every weekday
+  workflow_dispatch:
+
+jobs:
+  run-tests:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    strategy:
+      matrix:
+        # These are all the Python versions that we support.
+        python-version: ['3.7', '3.8', '3.9', '3.10']
+        concurrency: [1, 2, 5]
+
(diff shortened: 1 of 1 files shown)


.github/workflows/periodic-integration-tests.yml
```diff
@@ -0,0 +1,90 @@
...
+
+jobs:
+  run-tests:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    strategy:
+      matrix:
+        # These are all the Python versions that we support.
+        python-version: ['3.7', '3.8', '3.9', '3.10']
+        concurrency: [1, 2, 5]
```
kenxu95, line 16: I'm curious what insight we gain with concurrency=2 that we don't get with n=1 or n=5.

hsubbaraj-spiral: I was thinking this would give us more clarity if we have scaling issues (ie can't handle a certain number of connections). You're right that if we have race conditions the difference between 2 and 5 won't matter, both should fail inconsistently.

.github/workflows/periodic-integration-tests.yml
```diff
@@ -0,0 +1,90 @@
...
+      - name: Fetch the API key
+        run: echo "API_KEY=$(aqueduct apikey)" >> $GITHUB_ENV
+
+      - name: Run the Integration Tests
+        timeout-minutes: 10
+        working-directory: integration_tests/sdk
+        env:
+          SERVER_ADDRESS: localhost:8080
+          INTEGRATION: aqueduct_demo
+        run: pytest . -rP --publish -n ${{ matrix.concurrency }}
```
kenxu95, line 78: are we running with `--complex-models` too or does that seem unnecessary.

kenxu95: I could be convinced either way tbh

kenxu95 (approved):
LGTM but another Q

.github/workflows/periodic-integration-tests.yml
hsubbaraj-spiral: we can run complex-models as well since these are run once a day and time is not an issue.

PR #184, opened 2022-07-01
Deprecate the use of ipynbname package in our SDK

cw75:
The `ipynbname` has been causing issues whenever a notebook isn't started in the user's own terminal. This PR removes the use of this package and also clean up some deprecated logic in the packaging function. Please see the comments inline for the defensive check added.


kenxu95 (changes requested):
The change looks good to me!  But we definitely need an integration test for when the function is in a python file in a different directory, and it makes a relative import from another directory (this may already exist but would be good to check)

Would we be able to modify one of our example notebooks to include an operator defined in a separate python file (with additional dependencies in separate files too)? We can also just create a simple example test notebook and check it with the `run-notebook` action.

kenxu95 (approved):
LGTM!

PR #187, opened 2022-07-02
Direct user UI instead of pyplot dag image

Boyuan-Deng:

when the users are using sdk method, the method `publish_flow` `flow.describe()` `flow_run.describe()` will no longer return a pyplot dag image but return a url pointing to the UI

image
image
image


sdk/aqueduct/aqueduct_client.py
```diff
@@ -334,10 +334,11 @@ def publish_flow(
             retention_policy=retention_policy,
         )
 
-        if self._in_notebook_or_console_context:
-            _show_dag(self._api_client, dag)
-
         flow_id = self._api_client.register_workflow(dag).id
+
+        url = "http://" + self._api_client.aqueduct_address + "/workflow/" + str(flow_id)
```
kenxu95, line 339: we shouldn't assume that this is http. Lets expose a method on the APIClient that returns the address prefix? (`http(s)://<address>)

sdk/aqueduct/aqueduct_client.py
```diff
@@ -334,10 +334,11 @@ def publish_flow(
...
+        print("UI: ", url)
```
kenxu95, line 340: We should use the logger library here like the rest of the SDK.

sdk/aqueduct/flow.py
```diff
@@ -143,12 +144,15 @@ def describe(self) -> None:
         assert latest_metadata.schedule is not None, "A flow must have a schedule."
         assert latest_metadata.retention_policy is not None, "A flow must have a retention policy."
 
+        url = "http://" + self._api_client.aqueduct_address + "/workflow/" + self._id
```
kenxu95, line 147: we should probably just make the entire url construction a helper on utils.

sdk/aqueduct/flow_run.py
```diff
@@ -42,8 +45,9 @@ def status(self) -> ExecutionStatus:
 
     def describe(self) -> None:
         """Prints out a human-readable description of the flow run."""
-        if self._in_notebook_or_console_context:
-            _show_dag(self._api_client, self._dag)
+
+        url = "http://" + self._api_client.aqueduct_address + "/workflow/" + self._flow_id + "?"
```
kenxu95, line 49: the helper in utils will include this case historical run case too.

kenxu95 (changes requested):

sdk/aqueduct/aqueduct_client.py
Boyuan-Deng: This Print statement was used to print for sdk output 
image
I tried to use logger but it would not print to the jupyter notebook.

sdk/aqueduct/utils.py
```diff
@@ -53,6 +53,23 @@ def generate_uuid() -> uuid.UUID:
     return uuid.uuid4()
 
 
+def generate_url(
+    url_prefix: str = "", aqueduct_address: str = "", flow_id: str = "", result_id: str = ""
```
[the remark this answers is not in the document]
kenxu95: I think you don't need default values for anything except `result_id`, and it should be `Optional[str] = None` type.

sdk/aqueduct/flow.py
```diff
@@ -85,7 +85,8 @@ def _construct_flow_run(
                 continue
 
             dag.update_operator_spec(
-                param_artifact.name,  # this works because the parameter op and artifact currently share the same name.
+                # this works because the parameter op and artifact currently share the same name.
```
kenxu95, line 88: was this a linter change change? If not I feel like same line looks better :P

kenxu95 (approved):
awesome, thanks! Just suggested a few more tweaks but looks good to me

sdk/aqueduct/flow.py
Boyuan-Deng: Yes, this is a linter change.

sdk/aqueduct/utils.py
```diff
@@ -52,6 +52,29 @@ def generate_uuid() -> uuid.UUID:
     return uuid.uuid4()
 
 
+WORKFLOW_ID_URL_PARAM_TEMPLATE = "/workflow/%s"
+WORKFLOW_RESULT_ID_URL_PARAM_TEMPLATE = "?workflowDagResultId=%s"
```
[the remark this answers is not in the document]
Boyuan-Deng: @kenxu95 All changes done!. Thx for all the advices. Will learn from them and avoid repeating them in the future. Rly appreciate the details and explanation.

Boyuan-Deng:
@kenxu95 All changes done!. Thx for all the advices. Will learn from them and avoid repeating them in the future. Rly appreciate the details and explanation.

PR #194, opened 2022-07-06
Refactors S3 storage implementation to use user provided credentials

saurav-c:
This PR changes the Aqueduct server start behavior to use the storage config found in the config.yml file, instead of defaulting to file storage each time.

It also refactors all access to S3 for storage to use the AWS credentials provided by the user. Previously, the access was performed without explicitly specifying where to find the AWS credentials.

ENG 1311

See #193 for demo

### DIFF
--- a/src/golang/config/config.go
+++ b/src/golang/config/config.go
@@ -4,15 +4,20 @@ import (
 	"io/ioutil"
 	"os"
 
+	"github.com/aqueducthq/aqueduct/lib/collections/shared"
 	log "github.com/sirupsen/logrus"
 	"gopkg.in/yaml.v2"
 )
 
 type ServerConfiguration struct {
-	AqPath             string `yaml:"aqPath" json:"aq_path"`
-	EncryptionKey      string `yaml:"encryptionKey" json:"encryption_key"`
-	RetentionJobPeriod string `yaml:"retentionJobPeriod"`
-	ApiKey             string `yaml:"apiKey"`
+	AqPath             string                `yaml:"aqPath" json:"aq_path"`
+	EncryptionKey      string                `yaml:"encryptionKey" json:"encryption_key"`
+	RetentionJobPeriod string                `yaml:"retentionJobPeriod"`
+	ApiKey             string                `yaml:"apiKey"`
+	StorageConfig      *shared.StorageConfig `yaml:"storageConfig"`
+}
+
+type Storage struct {
 }
 
 func ParseServerConfiguration(confPath string) *ServerConfiguration {
--- a/src/golang/config/config.go
+++ b/src/golang/config/config.go
@@ -3,6 +3,7 @@ package config
 import (
 	"io/ioutil"
 	"os"
+	"path"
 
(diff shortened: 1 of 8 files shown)


src/golang/config/config.go
```diff
@@ -29,5 +32,17 @@ func ParseServerConfiguration(confPath string) *ServerConfiguration {
 		os.Exit(1)
 	}
 
+	if config.StorageConfig == nil {
```
likawind, line 35: Is this something necessary to make s3 working or it's a nice fix we add on the way?

src/golang/config/config.go
```diff
@@ -3,16 +3,19 @@ package config
...
 	log "github.com/sirupsen/logrus"
 	"gopkg.in/yaml.v2"
 )
 
 type ServerConfiguration struct {
-	AqPath             string `yaml:"aqPath" json:"aq_path"`
-	EncryptionKey      string `yaml:"encryptionKey" json:"encryption_key"`
-	RetentionJobPeriod string `yaml:"retentionJobPeriod"`
-	ApiKey             string `yaml:"apiKey"`
+	AqPath             string                `yaml:"aqPath" json:"aq_path"`
```
likawind, line 14: Why do we still keep `aqPath` here? Should the storage be parsed from `storageConfig` now?

src/golang/config/config.go
saurav-c: So it's necessary because we are adding a new field to the `config.yml` file. For users that don't run `aqueduct storage` that field won't exist, so we'd get a null pointer exception here without the check.

src/golang/config/config.go
saurav-c: If the storage is S3 we would not be able to determine `aqPath` from the `storageConfig`, since there would be no local filepath.

src/python/aqueduct_executor/operators/utils/storage/s3.py
```diff
@@ -14,6 +15,10 @@ class S3Storage(Storage):
     _config: S3StorageConfig
 
     def __init__(self, config: S3StorageConfig):
+        # Boto3 uses an environment variable to determine the credentials filepath and profile
+        os.environ["AWS_SHARED_CREDENTIALS_FILE"] = config.credentials_path
+        os.environ["AWS_PROFILE"] = config.credentials_profile
```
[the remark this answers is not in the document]
saurav-c: Mhm unfortunately :(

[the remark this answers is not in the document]
saurav-c: The alternative is to read the credentials file ourselves and explicitly set the access key and secret access key, but I'm not sure if there is a benefit to that over this method.

src/golang/config/config.go
likawind: Thanks! Actually I think I missed something, `aqPath` will be used in places other than storage so it does make sense to separate that from storage. This also resolved the above question.

likawind (approved):

PR #198, opened 2022-07-07
Schema migration for changing 'metadata' to 'exec_state'

likawind:
This PR adds schema migration for `operator_result` table by changing `metadata` to `exec_state` based on changes introduced in #159 . We added 3 new versions to add new col, backfill, and drop old col. After this change the schema will become 12.

ENG-1249

- [x] #161  I have included a small demo of the changes. For the UI, this would be a screenshot or a Loom video.
- [x] #168  If this is a new feature, I have added unit tests and integration tests.

### DIFF
--- a/src/golang/cmd/migrator/versions/000011_exec_state_column_backfill/queries.go
+++ b/src/golang/cmd/migrator/versions/000011_exec_state_column_backfill/queries.go
@@ -18,7 +18,7 @@ type opResultWithMetadata struct {
 type opResultWithExecState struct {
 	Id        uuid.UUID              `db:"id" json:"id"`
 	Status    shared.ExecutionStatus `db:"status" json:"status"`
-	ExecState NullExecutionState     `db:"exec_state" json:"exec_state"`
+	ExecState NullExecutionState     `db:"execution_state" json:"execution_state"`
 }
 
 func getOpResultsWithMetadata(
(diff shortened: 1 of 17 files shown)


cw75 (commented):
@likawind I was under the impression that SQLite doesn't support dropping columns, but according to [here](https://stackoverflow.com/a/5987838) the recent version does seem to support this. Have you verified that this works? Also reminder to update the schema change tracker [doc](https://www.notion.so/aqueducthq/Database-Schema-Change-Tracker-c861ce6ba45b43ad87fe7d91eee49df3) when you get a chance. Thx!

src/golang/cmd/migrator/versions/000011_exec_state_column_backfill/queries.go
```diff
@@ -0,0 +1,68 @@
...
+type opResultWithMetadata struct {
+	Id       uuid.UUID              `db:"id" json:"id"`
+	Status   shared.ExecutionStatus `db:"status" json:"status"`
+	Metadata NullMetadata           `db:"metadata" json:"metadata"`
+}
+
+type opResultWithExecState struct {
+	Id        uuid.UUID                 `db:"id" json:"id"`
+	Status    shared.ExecutionStatus    `db:"status" json:"status"`
+	ExecState shared.NullExecutionState `db:"exec_state" json:"exec_state"`
```
saurav-c, line 21: Won't this no longer work if we modify `shared.NullExecutionState` in the future? This is why we have to redefine the structs to the state they were in at the time of this migration.

likawind: Good point, fixed

likawind:
Integration test was successful. Current failure due to v0.0.4 release, will rerun after rebase

saurav-c (approved):
LGTM

PR #208, opened 2022-07-12
[Server] Create server API endpoint to fetch operators associated with integration

likawind:
This PR creates `listOperatorsOnIntegration` API endpoint to retrieve all operators for the given information. The main use cases are (copied from ENG-1324 description):
* When showing integration details, show wf / operators associated with this integration
* When deleting, block deletion and let user know there are existing workflows using the integration. Inform user if there's a previous workflow on this integration
* When editing, show all extract / load using the integration and ask user to confirm

To satisfy above requirements, the API returns a list where each item include:
* an `operator` object
* the workflow ID for this operator
* the workflow DAG ID for this operator
* an `is_active` flag noting if the workflow dag's version is 'current'.

If an operator appears in multiple DAGs, we will return the same operators associated with multiple wf_id and wf_dag_id. We will let the client to decide how to use these information.

The implementation is outlined as follows:
* fetch all operators associated with the integration ID
* fetch wf_id and wf_dag_id for each operator
* fetch latest dag id for all wf_id above
* combine the above 3 to get the final results.

We also added a small fix to replace all `->>` sql operators with `json_extract`, which we figure is more version compatible for sqlite3.

ENG-1324
ENG-1356

Validated the new endpoint in postman:
Screen Shot 2022-07-13 at 10 43 28 PM

- [ ] I have included a small demo of the changes. For the UI, this would be a screenshot or a Loom video. **N/A**
- [ ] If this is a new feature, I have added unit tests and integration tests. **Integration tests will be added in future PRs.**
- [ ] I have run the integration tests locally and they are passing. **Should be fine as long as in-PR integration tests passes.**
- [ ] I have run the linter script locally (See `python3 scripts/run_linters.py -h` for usage). **N/A no python change involved.**


src/golang/cmd/server/queries/sqlite.go
```diff
@@ -1,9 +1,66 @@
...
+		 json_extract(operator.spec, '$.type') = '%s' AND app_user.organization_id = $1;`,
+		operator.LoadType,
+	)
+
+	var response []LoadOperatorSpecResponse
+	err := db.Query(ctx, &response, query, organizationId)
+	return response, err
+}
+
+func (r *sqliteReaderImpl) GetCheckResultsByArtifactIds(
```
likawind, line 42: changes `->>` operator here https://github.com/aqueducthq/aqueduct/blob/main/src/golang/cmd/server/queries/standard.go#L102 to use `json_extract`

src/golang/cmd/server/queries/sqlite.go
```diff
@@ -1,9 +1,66 @@
...
+
 type sqliteReaderImpl struct {
 	standardReaderImpl
 }
 
 func newSqliteReader() Reader {
 	return &sqliteReaderImpl{standardReaderImpl{}}
 }
+
+func (r *sqliteReaderImpl) GetLoadOperatorSpecByOrganization(
```
likawind, line 22: changes `->>` operator here to use `json_extract`: https://github.com/aqueducthq/aqueduct/blob/main/src/golang/cmd/server/queries/standard.go#L16

src/golang/cmd/server/queries/postgres.go
```diff
@@ -1,9 +1,69 @@
...
+
 type postgresReaderImpl struct {
 	standardReaderImpl
 }
 
 func newPostgresReader() Reader {
 	return &postgresReaderImpl{standardReaderImpl{}}
 }
+
+func (r *postgresReaderImpl) GetLoadOperatorSpecByOrganization(
```
likawind, line 22: changes `->>` operator here to use `json_extract_path_text`: https://github.com/aqueducthq/aqueduct/blob/main/src/golang/cmd/server/queries/standard.go#L16

src/golang/cmd/server/queries/postgres.go
```diff
@@ -1,9 +1,69 @@
...
+		 AND app_user.organization_id = $1;`,
+		operator.LoadType,
+	)
+
+	var response []LoadOperatorSpecResponse
+	err := db.Query(ctx, &response, query, organizationId)
+	return response, err
+}
+
+func (r *postgresReaderImpl) GetCheckResultsByArtifactIds(
```
likawind, line 43: changes `->>` operator here to use `json_extract_path_text`: https://github.com/aqueducthq/aqueduct/blob/main/src/golang/cmd/server/queries/standard.go#L102

eunice-chan (approved):
Thanks for doing all the SQL query updates :)

src/golang/cmd/server/queries/standard.go
```diff
@@ -170,3 +124,130 @@ func (r *standardReaderImpl) GetWorkflowLastRun(
...
+		obj.OperatorId.String(),
+	)
+}
+
+func (r *standardReaderImpl) GetWorkflowIdsFromOperatorIds(
+	ctx context.Context,
+	operatorIds []uuid.UUID,
+	db database.Database,
+) ([]WorkflowIdsFromOperatorIdsResponse, error) {
+	// This query looks up all operators with at least one upstream
```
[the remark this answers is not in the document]
likawind: The main issue is I need to select the operator ID as well and that's either `from_id` or `to_id` depending on how the condition get matched. In this case a single `or` conditioning won't work. But I think there's some `union` trick I could try out to squeeze the query

src/golang/cmd/server/queries/queries.go
```diff
@@ -55,6 +55,12 @@ type WorkflowLastRunResponse struct {
 	LastRunAt  time.Time         `db:"last_run_at" json:"last_run_at"`
 }
 
+type WorkflowIdsFromOperatorIdsResponse struct {
```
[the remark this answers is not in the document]
likawind: We would need to operator object instead of the operator ID, so I think it's easier to just copy stuff over

src/golang/cmd/server/handler/list_operators_on_integration.go
```diff
@@ -0,0 +1,122 @@
...
+// Method: GET
+// Params: None
+// Request:
+//	Headers:
+//		`api-key`: user's API Key
+// Response:
+//	Body:
+//		serialized `listOperatorsOnIntegrationResponse`
+//
+// `listOperatorsOnIntegration` lists all operators associated with
```
saurav-c, line 26: nit: maybe consider renaming to`listOperatorsForIntegration`

src/golang/cmd/server/handler/list_operators_on_integration.go
```diff
@@ -0,0 +1,122 @@
...
+type ListOperatorsOnIntegrationHandler struct {
+	GetHandler
+
+	Database       database.Database
+	CustomReader   queries.Reader
+	OperatorReader operator.Reader
+}
+
+func (*ListOperatorsOnIntegrationHandler) Name() string {
+	return "ListWorkflowsOnIntegration"
```
saurav-c, line 54: There seems to be a name difference here

src/golang/cmd/server/queries/sqlite.go
```diff
@@ -1,9 +1,66 @@
...
+	organizationId string,
+	db database.Database,
+) ([]LoadOperatorSpecResponse, error) {
+	query := fmt.Sprintf(
+		`SELECT DISTINCT workflow_dag_edge.from_id AS artifact_id, artifact.name AS artifact_name, operator.id AS load_operator_id, workflow.name AS workflow_name, workflow.id AS workflow_id, operator.spec 
+		 FROM app_user, workflow, workflow_dag, workflow_dag_edge, operator, artifact
+		 WHERE app_user.id = workflow.user_id AND workflow.id = workflow_dag.workflow_id AND 
+		 workflow_dag.id = workflow_dag_edge.workflow_dag_id AND workflow_dag_edge.to_id = operator.id AND 
+		 artifact.id = workflow_dag_edge.from_id AND
+		 json_extract(operator.spec, '$.type') = '%s' AND app_user.organization_id = $1;`,
```
saurav-c, line 33: Nit: Instead of mixing the `%s` and `$1` syntax, can we use just the placeholders `$1` in both places here?

src/golang/cmd/server/queries/sqlite.go
```diff
@@ -1,9 +1,66 @@
...
+	if len(artifactIds) == 0 {
+		return nil, errors.New("Provided empty IDs list.")
+	}
+
+	query := fmt.Sprintf(
+		`SELECT DISTINCT workflow_dag_edge.from_id AS artifact_id, operator.name AS name, operator_result.status, 
+		 operator_result.metadata, operator_result.workflow_dag_result_id 
+		 FROM workflow_dag_edge, operator, operator_result 
+		 WHERE workflow_dag_edge.to_id = operator.id AND operator.id = operator_result.operator_id AND 
+		 workflow_dag_edge.from_id IN (%s) AND json_extract(operator.spec, '$.type') = '%s';`,
```
saurav-c, line 56: Same comment about placeholder syntax

src/golang/cmd/server/handler/list_operators_on_integration.go
```diff
@@ -0,0 +1,122 @@
...
+// `listOperatorsOnIntegration` lists all operators associated with
+// the given integraion. Together we provide the following information for
+// each associated operator:
+//  `workflow_id`: the workflow associated with this operator
+//	`workflow_dag_id`: the workflow dag associated with this operator
+//	`is_active`: whether the operator is being used in the latest version of the workflow.
+//				 This is the equivalent to whether `workflow_dag_id` is the latest for the `workflow_id`
+
+type listOperatorsOnIntegrationItem struct {
+	Operator      *operator.Operator
```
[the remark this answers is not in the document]
likawind: good catch! We should have json tag for each of  them

src/golang/lib/collections/integration/sqlite.go
```diff
@@ -82,3 +83,24 @@ func (w *sqliteWriterImpl) CreateIntegrationForUser(
...
+	getIntegrationsQuery := fmt.Sprintf(
+		"SELECT %s FROM integration WHERE json_extract(config, $1) = $2;",
+		allColumns(),
+	)
+	var integrations []Integration
+
+	// The full 'where' condition becomes
+	// `json_extract(config, '$.field_name') = 'field_value'`
+	// which matches https://www.sqlite.org/json1.html .
+	// We parametrize the extracted field_name and field_value
```
saurav-c, line 102: nit typo: parameterize

src/golang/cmd/server/queries/standard.go
```diff
@@ -170,3 +124,130 @@ func (r *standardReaderImpl) GetWorkflowLastRun(
...
+func keyWorkflowIdsFromOperatorIdsResponse(obj *WorkflowIdsFromOperatorIdsResponse) string {
+	return fmt.Sprintf(
+		"%s_%s_%s",
+		obj.WorkflowId.String(),
+		obj.WorkflowDagId.String(),
+		obj.OperatorId.String(),
+	)
+}
+
+func (r *standardReaderImpl) GetWorkflowIdsFromOperatorIds(
```
saurav-c, line 139: Can we add a database test for this

src/golang/cmd/server/handler/list_operators_on_integration.go
```diff
@@ -0,0 +1,122 @@
...
+	integrationId, err := uuid.Parse(integrationIdStr)
+	if err != nil {
+		return nil, http.StatusBadRequest, errors.Wrap(err, "Malformed integration ID.")
+	}
+
+	return integrationId, http.StatusOK, nil
+}
+
+func (h *ListOperatorsOnIntegrationHandler) Perform(ctx context.Context, interfaceArgs interface{}) (interface{}, int, error) {
+	integrationId := interfaceArgs.(uuid.UUID)
```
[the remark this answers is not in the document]
likawind: I found query like this https://github.com/aqueducthq/aqueduct/blob/main/src/golang/cmd/server/queries/standard.go#L159 very hard to follow, and here our query is probably even more complex. So I'm trying to keep things a bit more readable here  given there shouldn't be much performance difference

src/golang/cmd/server/handler/list_operators_on_integration.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/aqueducthq/aqueduct/cmd/server/queries"
+	"github.com/aqueducthq/aqueduct/cmd/server/routes"
+	"github.com/aqueducthq/aqueduct/lib/collections/operator"
+	"github.com/aqueducthq/aqueduct/lib/database"
+	"github.com/dropbox/godropbox/errors"
+	"github.com/go-chi/chi"
+	"github.com/google/uuid"
+)
+
+// Route: /integration/{integrationId}/workflows
```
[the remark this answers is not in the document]
likawind: yeah I missed a few things while rename the first draft

src/golang/cmd/server/queries/sqlite.go
likawind: We could, but I somehow found using `%s` for constant owned by us, and `$1` for user parameters is a pattern through out the db codes: https://github.com/aqueducthq/aqueduct/blob/main/src/golang/lib/collections/artifact/standard.go#L86 . I find this is a quite clean cut to distinguish the usage between the two. Esp. that some `%s` is necessary, I will add more details in the next comment down

src/golang/cmd/server/queries/sqlite.go
likawind: The first one is necessary since it's actually generating all the placeholders. We would need the secound `%s` anyway since the placeholder index will depend on `len(artifactIds)`. In this case there's not much difference putting `CheckType` directly since it's a constant controlled by us

src/golang/cmd/server/queries/sqlite.go
likawind: An alternative is to restrict all `%s` use to indexing and force all other use cases to use parameters.

saurav-c: Got it, I think we can create a task for this for now and come back to it.

likawind: https://linear.app/aqueducthq/issue/ENG-1367/make-better-use-of-query-parametrization

PR #209, opened 2022-07-13
Adds workflow_dag engine_config column

saurav-c:
This PR adds an `engine_config` column to the `workflow_dag` table. This is necessary to store important information for mapping an Aqueduct workflow to the native concepts of the 3rd-party engine the workflow is set to run on, such as Airflow.

ENG-1243

## Test
I tested this schema change by creating a workflow (and workflow_dag) before applying the schema change. Then, I applied the schema change and confirmed that the new column was created. After this, I tested that the `WorkflowDagReader` was able to successfully read the `engine_config` column.

### DIFF
--- a/src/golang/cmd/migrator/migrator/register.go
+++ b/src/golang/cmd/migrator/migrator/register.go
@@ -15,6 +15,7 @@ import (
 	_000010 "github.com/aqueducthq/aqueduct/cmd/migrator/versions/000010_add_exec_state_column"
 	_000011 "github.com/aqueducthq/aqueduct/cmd/migrator/versions/000011_exec_state_column_backfill"
 	_000012 "github.com/aqueducthq/aqueduct/cmd/migrator/versions/000012_drop_metadata_column"
+	_000013 "github.com/aqueducthq/aqueduct/cmd/migrator/versions/000013_add_workflow_dag_engine_config"
 	"github.com/aqueducthq/aqueduct/lib/database"
 )
 
@@ -76,7 +77,6 @@ func init() {
 		downPostgres: _000009.Down,
 		name:         "backfill metadata in artifact_results",
 	}
-
 	registeredMigrations[10] = &migration{
 		upPostgres: _000010.UpPostgres, upSqlite: _000010.UpSqlite,
 		downPostgres: _000010.DownPostgres,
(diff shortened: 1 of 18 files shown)


src/golang/cmd/migrator/migrator/register.go
```diff
@@ -15,6 +15,7 @@ import (
 	_000010 "github.com/aqueducthq/aqueduct/cmd/migrator/versions/000010_add_exec_state_column"
 	_000011 "github.com/aqueducthq/aqueduct/cmd/migrator/versions/000011_exec_state_column_backfill"
 	_000012 "github.com/aqueducthq/aqueduct/cmd/migrator/versions/000012_drop_metadata_column"
+	_000013 "github.com/aqueducthq/aqueduct/cmd/migrator/versions/000013_add_workflow_dag_engine_config"
```
saurav-c, line 18: TODO: Add this to the schema change tracker in Notion.

src/golang/cmd/migrator/versions/000013_add_workflow_dag_engine_config/up_postgres.go
```diff
@@ -0,0 +1,6 @@
+package _000013_add_workflow_dag_engine_config
+
+const upPostgresScript = `
+ALTER TABLE workflow_dag
+ADD COLUMN engine_config JSONB NOT NULL
+DEFAULT '{"type":"aqueduct", "aqueduct_config":{}}'::jsonb;`
```
likawind, line 6: what will happen to existing rows? Will they be back-filled with this given value?

src/golang/cmd/server/server/aqueduct_server.go
```diff
@@ -29,7 +29,7 @@ import (
 )
 
 const (
-	RequiredSchemaVersion = 8
```
likawind, line 32: wow where is this field used? I'm surprised that nothing has been broken so far

likawind (approved):
The test plan looks very solid!

src/golang/cmd/migrator/versions/000013_add_workflow_dag_engine_config/up_postgres.go
saurav-c: Yes, by adding a `DEFAULT` all existing rows are given that value for the new column. If we didn't have a `DEFAULT` (and did not have `NOT NULL`) then the values would be `NULL`.

src/golang/cmd/server/server/aqueduct_server.go
saurav-c: Well this was more important in the enterprise project. Here once we modify the `install_local` and packaging scripts, we know that the next release will be running the correct schema version.

We do have a check here https://github.com/aqueducthq/aqueduct/blob/main/src/golang/cmd/server/server/aqueduct_server.go#L118 just to confirm.

PR #210, opened 2022-07-13
Adds default values for Python connectors for Postgres.port and Extract.query_is_usable

saurav-c:
This PR adds a couple of default values that are necessary for backwards compatibility for workflows that were registered before we added these fields.

### DIFF
--- a/src/python/aqueduct_executor/operators/connectors/tabular/extract.py
+++ b/src/python/aqueduct_executor/operators/connectors/tabular/extract.py
@@ -27,16 +27,17 @@ BUILT_IN_EXPANSIONS = {
 
 
 class RelationalParams(models.BaseParams):
-    # The query cannot be used until `apply_placeholders()` is called on it. This flushes out
-    # any user-defined tags like `{{today}}`.
-    # _query_is_usable is not passed in via the spec, so we prefix it with an _ to ensure that
-    # Pydantic ignores the field during validation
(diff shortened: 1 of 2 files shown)


cw75 (approved):

cw75:
@saurav-c mypy seems failing but lgtm

saurav-c:
Yup, I'm trying to fix the places where `self.query_is_usable` is being used, since it's now an `Optional` field.

src/python/aqueduct_executor/operators/connectors/tabular/extract.py
```diff
@@ -29,7 +29,10 @@ def replace_today() -> str:
 class RelationalParams(models.BaseParams):
     # The query cannot be used until `apply_placeholders()` is called on it. This flushes out
     # any user-defined tags like `{{today}}`.
-    query_is_usable: bool = False
+    # _query_is_usable is not passed in via the spec, so we prefix it with an _ to ensure that
```
saurav-c, line 32: @kenxu95 Could you confirm that my understanding is correct here. I believe that this instance variable is just used for tracking state, but it isn't passed via the job spec ever from Go to Python?

kenxu95: yeah this never created from Golang, its just lifecycle management so that we can't use RelationalParams ever without expanding parameter values.

PR #211, opened 2022-07-13
Support multi file inputs from S3 integration

cw75:
This PR enables user's S3 integration to extract multiple files in a concatenated form, so when they have multiple files they don't need to define a bunch of extract operators. The expected value for the SDK API is commented inline. The PR also adds an authentication step when the user registers an S3 integration by trying to list the files in the bucket.

https://www.loom.com/share/b692707630a44f3bbf0ebfa107e6334b

### DIFF
--- a/sdk/aqueduct/integrations/s3_integration.py
+++ b/sdk/aqueduct/integrations/s3_integration.py
@@ -30,7 +30,7 @@ class S3Integration(Integration):
 
     def file(
         self,
-        filepath: Union[List[str], str],
+        filepaths: Union[List[str], str],
         format: S3FileFormat,
         name: Optional[str] = None,
         description: str = "",
@@ -39,8 +39,8 @@ class S3Integration(Integration):
         Retrieves a file from the S3 integration.
 
         Args:
-            filepath:
-                Filepath to retrieve from. The filepath can either be:
+            filepaths:
+                Filepath to retrieve from. The filepaths can either be:
                 1) a single string that represents a file name or a directory name. The directory
                 name must ends with a `/`. In case of a file name, we attempt to retrieve that file,
                 and in case of a directory name, we do a prefix search on the directory and retrieve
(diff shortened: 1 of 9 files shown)


sdk/aqueduct/integrations/s3_integration.py
```diff
@@ -66,7 +73,9 @@ def file(
                             extract=ExtractSpec(
                                 service=integration_info.service,
                                 integration_id=integration_info.id,
-                                parameters=S3ExtractParams(filepath=filepath, format=format),
+                                parameters=S3ExtractParams(
+                                    filepath=json.dumps(filepath), format=format
```
saurav-c, line 77: Can we document this somewhere that filepath needs to be JSON serialized

saurav-c (approved):
LGTM overall!

A couple of nits:
- we may want to rename `filepath` to `filepaths` to make it clear that we can multiple S3 paths
- did you figure out whether or not we need to specify an AWS region or does the default work in most cases?

sdk/aqueduct/integrations/s3_integration.py
cw75: documented.

cw75:
@saurav-c I changed the SDK external API to have `filepaths`, but didn't change the inner ones since that may require a data migration due to json tag name change. For the region, S3 bucket actually needs to be globally unique, and I verified that the region doesn't matter when initializing boto resource.

vsreekanti:
Thanks for the video, @cw75 -- super helpful! Good catch on the incorrect validations. This mostly look good. Just one UI nit: 
* Rather than printing a JSON blob that's a little tough to read, could we please grab the error field out of the JSON blob and then use a preformatted text block (i.e., the `<pre> </pre>`) tag to format the error message? It will make it way easier to read. I believe we have an example of doing this in the log viewer component if you need a pointer.

Let's also make sure we clearly document the prefix search/nesting thing.

cw75:
@vsreekanti Does this look fine now?
Screen Shot 2022-07-13 at 6 32 13 PM

PR #214, opened 2022-07-14
ENG-1358 Set up backend integration tests & include in Github Actions

eunice-chan:
Suite of tests for pure endpoint testing. Checking reads in a stricter fashion. 

## Testing
Successfully ran Integration Tests Github Actions workflow with these changes: https://github.com/aqueducthq/aqueduct/actions/runs/2667299271

Eng 1358
Eng 1345 (https://github.com/aqueducthq/aqueduct/pull/216): The changes being tested


eunice-chan:
Hi @kenxu95 made an initial draft of the backend integration test suite. Had some questions & would appreciate your feedback:
1. How to handle if workflows already exist? e.g. try to create workflow_1 but already exists and overwrites existing. Workflow_dag check won't work because will have extra history from the existing workflow_1. I'm thinking we can have an account specific for testing. Not sure if we support multiple users in the OSS version.
2. Could I grab the address and apikey from something? E.g. the .aqueduct config.yml file for apikey. Should I?
3. Not really sure how to add to Github Actions.

kenxu95:
1. So you should be able to assume that you start with a blank slate. So as long as the workflows you're creating in this suite don't collide with each other, you should be good. The SDK tests use uuid.New() to generate workflow names and avoid collisions, but for this suite we might be able to get away with better, human-readable names since all the tests will share the same workflows.
2. You can do something similar to what the SDK integration tests do, which is grab the apikey from an env variable. See `sdk/conftest.py` for the details of how those integration tests were set up.
3. You should be able to run the full suite with a pytest command right? We can just add that command to `.github/workflows/integration-tests.yml`

.github/workflows/integration-tests.yml
```diff
@@ -67,14 +67,22 @@ jobs:
...
           SERVER_ADDRESS: localhost:8080
           INTEGRATION: aqueduct_demo
         run: pytest . -rP --publish -n 1
 
+      - name: Run the Backend Integration Tests
+        timeout-minutes: 10
+        working-directory: integration_tests/backend
+        env:
+          SERVER_ADDRESS: localhost:8080
+          ADAPTER: http://
```
eunice-chan, line 83: For the `requests.get` url

integration_tests/backend/README.md
```diff
@@ -0,0 +1,18 @@
+# Backend Integration Tests
+
+These tests run the Aqueduct backend.
```
kenxu95, line 3: Can you write a quick sentence or two about what kinds of tests are here? And a bit about how tests are setup and cleaned up?

integration_tests/backend/conftest.py
```diff
@@ -0,0 +1,20 @@
...
+
+API_KEY_ENV_NAME = "API_KEY"
+SERVER_ADDR_ENV_NAME = "SERVER_ADDRESS"
+ADAPTER_ENV_NAME = "ADAPTER"
+
+
+def pytest_configure(config):
+    pytest.api_key = os.getenv(API_KEY_ENV_NAME)
+    pytest.server_address = os.getenv(SERVER_ADDR_ENV_NAME) or "localhost:8080"
+    pytest.adapter = os.getenv(ADAPTER_ENV_NAME) or "http://"
```
kenxu95, line 15: I think we should just error if the env variables aren't supplied. Let's not hide the test configuration if we can.

integration_tests/backend/setup/load_workflow.py
```diff
@@ -0,0 +1,23 @@
...
+    tables = []
+    for table_name, update_mode in zip(table_names, update_modes):
+        table = integration.sql(query="SELECT * FROM wine;")
+        tables.append(table)
+        table.save(integration.config(table=table_name, update_mode=update_mode))
+
+    return client.publish_flow(name=workflow_name, artifacts=tables,)
+
+
+def create_test_endpoint_GetWorkflowTables_flow(client, workflow_name, table_names, update_modes):
```
kenxu95, line 20: can we keep naming consistent? all lower case for `getworkflowtables`

integration_tests/backend/test_backend.py
```diff
@@ -0,0 +1,81 @@
...
+        cls.flows.append(cls.test_endpoint_GetWorkflowTables_flow)
+
+    @classmethod
+    def teardown_class(cls):
+        for flow in cls.flows:
+            cls.client.delete_flow(flow.id())
+
+    def test_endpoint_GetWorkflowTables(self):
+        headers = {"api-key": pytest.api_key}
+        url = f"{pytest.adapter}{pytest.server_address}/api/workflow/{self.test_endpoint_GetWorkflowTables_flow.id()}/tables"
```
kenxu95, line 44: I feel like its more readable to have a URL template, like what we do in api_client.py. That way other tests that use the same route can reuse the same template without drift.

integration_tests/backend/conftest.py
kenxu95: Why not force the user to put the full URL in the server_address? Having them in two separate fields seems more confusing.

kenxu95 (changes requested):
The test is really hard for me to read and understand, can we chat in person? Maybe we can pair a little :)

integration_tests/backend/conftest.py
eunice-chan: I can do that. For the Github Actions workflow for SDK integration tests, SERVER_ADDRESS=localhost:8080 so I thought it would be confusing if the backend integration tests require `http://` or `https://` in the server address but not the SDK integration tests.

eunice-chan:
Action items:
Abstract away adapter (see: https://github.com/aqueducthq/aqueduct/blob/main/sdk/aqueduct/api_client.py)
Run directory of workflows which is just python script that returns workflow_id

integration_tests/backend/test_backend.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def get_response(endpoint, additional_headers={}):
+    headers = {"api-key": pytest.api_key}
+    headers.update(additional_headers)
+    url = f"{pytest.server_address}{endpoint}"
+    if url[:4] != "http":
+        try:
+           r = requests.get("https://"+url, headers=headers) 
+        except:
```
kenxu95, line 18: Can we figure out what prefix to use in the  `setup_class()` method. That way we don't have to send bad requests for every `get_response()` call. You sould just pass the url to `get_response()`.

integration_tests/backend/test_backend.py
```diff
@@ -0,0 +1,73 @@
...
+            try:
+                r = requests.get("http://"+url, headers=headers) 
+            except:
+                raise Exception(f"Cannot connect to {url}")
```
kenxu95, line 22: Would it be possible to use the same logical structure that we use in the api_client.py on the SDK? That way its 1) more consistent across our tests and 2) avoiding having nested try-except clauses which is a bit confusing.

integration_tests/backend/test_backend.py
```diff
@@ -0,0 +1,73 @@
...
+    else:
+        try:
+            r = requests.get(url, headers=headers)
+        except:
+            raise Exception(f"Cannot connect to {url}")
+    return r
+
+class TestBackend:
+    GET_WORKFLOW_TABLES_TEMPLATE = "/api/workflow/%s/tables"
+    WORKFLOW_PATH = Path(__file__).parent / "setup"
```
kenxu95, line 32: wait I've never seen this before, how does the "/" not error here? The right side isn't wrapped in a string.

kenxu95: ok sg lets just leave it this way

integration_tests/backend/test_backend.py
eunice-chan: Actually, should I use the `use_http` variable directly?
image

kenxu95: hmm that might be better - how about we add a method on api_client called `ui_and_server_address()`, which will return the {prefix}{server_address} as a string that you can join with the routes? This'll be a useful internal method anyways  - see `generate_ui_url()` in the SDK's utils.py

kenxu95: I like your suggestion of keeping the logic in one place, this just wraps it a little more.

eunice-chan: - using existing `api_client._construct_full_url` in tests
- created `api_client._construct_base_url`
- `api_client._construct_full_url` uses `api_client._construct_base_url`
- made the arg `use_http` optional for`api_client._construct_full_url` and `api_client._construct_base_url` (if `use_http=None`, use `self.use_http`)
- refactor `generate_ui_url` to use `api_client._construct_full_url`
- clean url in `api_client` so `SERVER_ADDRESS=localhost:8080/` works (would error out before)

sdk/aqueduct/api_client.py
```diff
@@ -98,9 +102,16 @@ def __init__(self, api_key: str, aqueduct_address: str):
         else:
             self.use_https = self._test_connection_protocol(try_http=True, try_https=True)
 
-    def _construct_full_url(self, route_suffix: str, use_https: bool) -> str:
+    def _construct_base_url(self, use_https: Optional[bool] = None) -> str:
+        if use_https == None:
+            use_https = self.use_https
```
kenxu95, line 107: lets try to use `is None` comparisons here: https://stackoverflow.com/questions/14247373/python-none-comparison-should-i-use-is-or#:~:text=is%20is%20generally%20preferred%20when,and%20even%20on%20their%20ordering.

kenxu95: also, shouldn't we error here instead of using https? `use_https` must have been set by the constructor.

sdk/aqueduct/api_client.py
```diff
@@ -98,9 +102,16 @@ def __init__(self, api_key: str, aqueduct_address: str):
...
         protocol_prefix = self.HTTPS_PREFIX if use_https else self.HTTP_PREFIX
-        return "%s%s%s" % (protocol_prefix, self.aqueduct_address, route_suffix)
+        return "%s%s" % (protocol_prefix, self.aqueduct_address)
+
+    def _construct_full_url(self, route_suffix: str, use_https: Optional[bool] = None) -> str:
+        if use_https == None:
```
kenxu95, line 112: same with `is None` and getting rid of use_https argument.

sdk/aqueduct/api_client.py
kenxu95: We can get rid of the `use_https` argument right?

integration_tests/backend/test_backend.py
```diff
@@ -0,0 +1,76 @@
...
+    @classmethod
+    def teardown_class(cls):
+        for flow in cls.flows:
+            cls.client.delete_flow(cls.flows[flow])
+
+    @classmethod
+    def get_response_class(cls, endpoint, additional_headers={}):
+        headers = {"api-key": pytest.api_key}
+        headers.update(additional_headers)
+        url = cls.client._api_client._construct_full_url(endpoint)
```
kenxu95, line 54: oh interesting... this works for now, but it might be nice to decouple the full_url implementation and just use the `base_url`, since there will be endpoints we want to test in this suite that aren't available in api_client.

kenxu95 (changes requested):
Erroneous package-lock.json changes? I can stamp after those are fixed. Also dropped a few simplifying suggestions.

integration_tests/backend/test_backend.py
eunice-chan: `_construct_full_url` just returns `"%s%s" % (self._construct_base_url(use_https), route_suffix)` so I think re-implementing it would be duplicative,

sdk/aqueduct/api_client.py
eunice-chan: `use_https` is set to `self.use_https` except for in `_test_connection_protocol` which sets the argument directly to True or False

integration_tests/backend/test_backend.py
kenxu95: oh right its just the URL construction. Ok sounds good to me - maybe we can get rid of the leading underscore then?

sdk/aqueduct/api_client.py
kenxu95: ah I remember now, ok sg.

eunice-chan: ^ Because of that, I don't really see a need to explicitly add `use_https=self.use_https` in the function call each time, since the function has access to that value so I changed it to optional rather than required

kenxu95 (approved):
Thanks for powering through this!

PR #219, opened 2022-07-14
Rmeove unnecessary files from parcel bundle

Boyuan-Deng:

after adding the react-code-block package, a lot of unnecessary files were added to built during parcel built time. I used an external package parcel-resolver-ignore to ignore unnecessary files from building 

https://linear.app/aqueducthq/issue/ENG-1327/only-include-relevant-js-files-in-parcel-bundle

files will be built after change:
image


src/ui/app/package.json
```diff
@@ -54,9 +55,162 @@
...
     "parcel": "2.6.0",
-    "@parcel/css": "^1.10.0",
+    "parcel-resolver-ignore": "^2.1.3",
     "path-browserify": "^1.0.1",
     "process": "^0.11.10",
     "typescript": "^3.9.10"
-  }
+  },
+  "parcelIgnore": [
+    "refractor/lang/abap.*\\.js$",
```
vsreekanti, line 64: This is quite a long list and it seems brittle -- if the library adds new language support, it will need maintenance. I see two options here:
1. Can we use the resolver to provide paths for what we want to include rather than exclude?
2. Can we find a different library that doesn't do syntax highlighting in this fashion?

vsreekanti (changes requested):

src/ui/app/package.json
Boyuan-Deng: After investigation, I decided to use regular expression to exclude all languages from the refractor/lang folders except the python language. Hence, I went for the first approach " use the resolver to provide paths for what we want to include rather than exclude".

PR #226, opened 2022-07-19
Refactors python operator entrypoints

saurav-c:
This PR refactors the entrypoints to all of our Python operator implementations. The only way to invoke the Python operators was to execute the relevant Python module and pass in the spec as a command line argument. This will not work for Airflow, as the entrypoint needs to be a Python function, not a bash script. These changes will enable Airflow support (for which there will be several follow on PRs.)

The high level changes are as follows:
- `main.py` has for the most part been renamed to `execute.py`
- `main.py` contains only the code for parsing the spec from the command line argument. It then calls the `run` function in `execute.py` which does the core logic of the operator and is responsible for capturing user errors and reporting the execution state to storage as before.

ENG 1392


src/python/aqueduct_executor/operators/connectors/tabular/spec.py
```diff
@@ -44,10 +44,12 @@ def unwrap_connector_config(cls, connector_config, values):  # type: ignore
     if "connector_name" not in values:
         raise ValueError("Unknown connector name.")
 
-    values["connector_name"]
-
     if not isinstance(connector_config, dict):
         raise ValueError("connector_config is not a dictionary.")
+    
+    if "conf" not in connector_config:
```
saurav-c, line 50: This is necessary for Airflow parsing of the spec

likawind:
To make sure I understand, the `run()` will be what's shared between both airflow and our own executor? And in this case, we will pass both the `run()` to airflow, and job spec will be generated as airflow orchestrate the DAG?

saurav-c:
@likawind Yes, that's correct

saurav-c:
@ reviewers the integration tests are failing because the `start-function-executor.sh` script was modified, and the integration tests are still using the old script.

src/python/aqueduct_executor/operators/connectors/tabular/spec.py
kenxu95: Can you add a comment explaining why this is custom logic for Airflow?

kenxu95 (approved):
Looks fine to me 👍

cw75 (approved):

PR #227, opened 2022-07-20
Add ability to distinguish between warnings and error level checks on backend

kenxu95:
This is the revisit of previously closed. https://github.com/aqueducthq/aqueduct/pull/195
This is built on top of the orchestration refactor.

The approach:
- Because we want to keep our python operators generic, I introduced the concept of "Blacklisted outputs" on the function spec. For all operators except blocking checks, this field will be None. For blocking checks, we want to blacklist the output "False". That is to say, if an output matches an entry in this list, we will fail the entire workflow.
- On the read side, the python operator level is too generic to provide a very user friendly tip. So instead, it provides a generic one that is then overwritten at the `checkOp.GetExecState()` level with more specific user friendly context.

ENG-1265

- [ x] I have created a descriptive PR title. The PR title should complete the sentence "This PR...".
- [ x] If this is a new feature, I have added unit tests and integration tests.
- [ x] I have run the integration tests locally and they are passing.
- [ x] I have run the linter script locally (See `python3 scripts/run_linters.py -h` for usage).

### DIFF
--- a/integration_tests/sdk/checks_test.py
+++ b/integration_tests/sdk/checks_test.py
@@ -141,31 +141,29 @@ def test_check_wrong_number_of_inputs(client):
         check_artifact.get()
 
 
-@check(severity=CheckSeverity.ERROR)
-def success_check_return_numpy_bool(df):
-    return df["total_charges"].mean() < 2500
-
-
 def test_check_with_numpy_bool_output(client):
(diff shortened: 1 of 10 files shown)


cw75 (approved):
One nit otherwise LGTM!

PR #231, opened 2022-07-20
Cleanup all expanded function folders after execution finishes

kenxu95:
We currently don't do any deletion of the function directories that are expanded in `execute_function.py` (this is the model.py, model.pkl stuff). This means that every function execution, *even if its a preview*, leaves residue in the user's system, which is a bit silly. Once the user's function is executed, we don't need to keep the expanded directory around (we'll still have the serialized function representation if we need it - that's a separate storage path.

ENG-1375

- [x ] I have created a descriptive PR title. The PR title should complete the sentence "This PR...".
- [ x] I have run the integration tests locally and they are passing.
- [ x] I have run the linter script locally (See `python3 scripts/run_linters.py -h` for usage).

### DIFF
--- a/src/golang/lib/job/process.go
+++ b/src/golang/lib/job/process.go
@@ -326,16 +326,22 @@ func (j *ProcessJobManager) Poll(ctx context.Context, name string) (shared.Execu
 	defer func() {
 		if len(command.fnExtractStoragePath) > 0 {
 			log.Errorf("Attempting to remove %s", command.fnExtractStoragePath)
-			//if _, err = os.Stat(command.fnExtractStoragePath); os.IsExist(err) {
-			//	err = os.Remove(command.fnExtractStoragePath)
-			//	if err != nil {
-			//		log.Errorf("Unable to remove function extraction directory %s.", command.fnExtractStoragePath)
-			//	}
-			//}
+
+			// Since this is a somewhat scary recursive deleete, instead of just checking that function
+			// extract storage path exists, we check for the presence of a particular file inside this
+			// directory. This is to help make sure that we aren't deleting a folder that we aren't supposed to.
+			if _, err = os.Stat(path.Join(command.fnExtractStoragePath, "op", "model.pkl")); !os.IsNotExist(err) {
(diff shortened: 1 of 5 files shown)


src/golang/lib/job/process.go
```diff
@@ -268,6 +270,15 @@ func (j *ProcessJobManager) Launch(
...
 
+	// For function operators, we need to set the directory path to expand the function into here.
+	var fnExtractStoragePath string
+	if spec.Type() == FunctionJobType {
+		if fnSpec, ok := spec.(*FunctionSpec); ok {
+			fnExtractStoragePath = path.Join(j.conf.OperatorStorageDir, uuid.New().String())
+			fnSpec.FunctionExtractPath = fnExtractStoragePath
+		}
+	}
+
```
kenxu95, line 281: This is a bit annoying, because we need to set the field on the Command object at this level, but we also need to modify the spec if its a function. Ideally, the caller would pass in the spec with the extract field already populated, but the caller does not have access to `OperatorStorageDir` at the moment.

src/golang/lib/workflow/operator/base.go
```diff
@@ -273,15 +296,18 @@ func (bfo *baseFunctionOperator) jobSpec(fn *function.Function) job.Spec {
...
 			job.FunctionJobType,
 			bfo.jobName,
 			*bfo.storageConfig,
 			bfo.metadataPath,
 		),
 		FunctionPath: fn.StoragePath,
-		/* `FunctionExtractPath` is set by the job manager at launch time. */
+
+		// TODO(kenxu): we always use the default storage directory here.
+		FunctionExtractPath: path.Join(job.DefaultOperatorStorageDir, uuid.New().String()),
```
kenxu95, line 309: any ideas for how to grab the correct path here (we need to use the same path that the job manager is using)? The job manager is embedded very tightly with the AqServer so I don't see a great way of doing this at the moment. However, the rest of the code definitely seems like its at the right level of abstraction.

kenxu95:
This is actually just an issue on mac, and there's a much more obvious solution that I missed. PR coming out

PR #236, opened 2022-07-21
Switch to a low-overhead way of authenticating S3 integration

cw75:
Previously we were doing a list operation to check if the user has access to the bucket, which can be very costly if the bucket has lots of files. This low-overhead approach is borrowed from [here](https://stackoverflow.com/a/49817544) and I verified that it works.


saurav-c (approved):
LGTM

PR #239, opened 2022-07-22
Rename GetWorkflowTables -> GetWorkflowObjects

eunice-chan:
## Describe your changes and why you are making these changes
- Rename tables to objects because the things written to by the workflow in the integration are not necessarily tables.
- Update `integration_tests/backend/test_reads.py` to delete workflows & wait for flow runs in a similar way as in the SDK.
## Related issue number (if any)
ENG 1394 (builds on this PR)
ENG-1369 (builds on above)
ENG-1325 (builds on above)
ENG-1431 (builds on above)

## Checklist before requesting a review
- [x] I have created a descriptive PR title. The PR title should complete the sentence "This PR...".
- [x] I have performed a self-review of my code.
- [N/A] I have included a small demo of the changes. For the UI, this would be a screenshot or a Loom video.
- [N/A] If this is a new feature, I have added unit tests and integration tests.
- [x] I have run the integration tests locally and they are passing.
- [x] I have run the linter script locally (See `python3 scripts/run_linters.py -h` for usage).
- [N/A] All features on the UI continue to work correctly.
- [x] Added one of the following CI labels:
    - `run_integration_test`: Runs integration tests
    - `skip_integration_test`: Skips integration tests (Should be used when changes are ONLY documentation/UI)

PR #240, opened 2022-07-22
Eng 1394 allow getting workflow-written/saved objects from the sdk

eunice-chan:
Get all objects written by the given workflow (`.save`).
This is to make it easier to delete all touched objects/tables when deleting workflows.

For example:
If a workflow has table artifacts labeled `table_1`, `table_2`, and `table_3` and ran `table_{i}.save(name="saved_table_{i}", integration=integration)`, when we get all objects written by that workflow, we will get an object representing "saved_table_{i}" in the integration.
This makes it easy to pass all touched objects to the workflow deletion function e.g. `client.delete_flow(flow.id(), objects_to_delete=saved_table_{1-3})`.

## Demo
Output of `self.client.get_workflow_writes(self.flows["changing_saves.py"])`:
```
defaultdict(<class 'list'>, {UUID('6744ab2b-06ed-4865-89a6-faeea7064b51'): [WrittenObject(name='table_1', update_mode='append'), WrittenObject(name='table_1', update_mode='replace'), WrittenObject(name='table_2', update_mode='append')]})
```
ENG 1394
ENG-1369 (builds on above)
ENG-1325 (builds on above)
ENG-1431 (builds on above)


vsreekanti:
@eunice-chan @likawind, a few things: 
1. I'm not sure this function should be on the top level client object. It seems like we should actually have method be on a workflow object itself. 
2. Shouldn't the `WrittenObject` type have the integration the object has been written to? Can we also add a type in anticipation of the non-tabular work? 
3. We need to be careful about naming. It's confusing if we use write and save interchangeably. We should settle on a fixed vocabulary.

@kenxu95 @cw75 tagging you for the discussion around naming.

cw75:
@eunice-chan I'm having trouble understanding both the PR title and the PR description. :) Could you make them easier to read? Thx!

eunice-chan:
Thanks for your comments @vsreekanti
> 1. I'm not sure this function should be on the top level client object. It seems like we should actually have method be on a workflow object itself.
I think putting it on the workflow object is a good idea since it's getting the properties and not actually making any changes. I can update this.

> 2. Shouldn't the `WrittenObject` type have the integration the object has been written to? Can we also add a type in anticipation of the non-tabular work?
The structure being returned to the user is
```
{
integration_id: list of object(name, update_mode) in the integration
}
```
The user will be able to identify which integration that object is a part of, but I think having the integration on the object could also be useful so if they have the object by itself they can still access the integration information.

> 3. We need to be careful about naming. It's confusing if we use write and save interchangeably. We should settle on a fixed vocabulary.
I think that is a very good point. I struggled with coming up with a good name for this object. Initially I named it Table but I realized we support non-tabular data as well so I changed the terminology to be a little more inclusive (https://github.com/aqueducthq/aqueduct/pull/239). I think having consistent terminology is definitely a good idea and open to discussing this.

eunice-chan:
@cw75 Updated the title & description. Hope it clarifies things

sdk/aqueduct/api_client.py
```diff
@@ -98,6 +99,7 @@ class APIClient:
     LIST_INTEGRATIONS_ROUTE = "/api/integrations"
     LIST_TABLES_ROUTE = "/api/tables"
     GET_WORKFLOW_ROUTE_TEMPLATE = "/api/workflow/%s"
+    GET_WORKFLOW_OBJECTS_ROUTE = "/api/workflow/%s/objects"
```
kenxu95, line 102: `GET_WORKFLOW_SAVED_OBJECTS_ROUTE`?

kenxu95: and we'd name the api client method consistently.

sdk/aqueduct/aqueduct_client.py
```diff
@@ -246,6 +248,26 @@ def list_flows(self) -> List[Dict[str, str]]:
...
+
+        if all(uuid.UUID(flow_id) != workflow.id for workflow in self._api_client.list_workflows()):
+            raise InvalidUserArgumentException("Unable to find a flow with id %s" % flow_id)
+
+        workflow_writes = self._api_client.get_workflow_writes(flow_id).object_details
+        writes_mapping = defaultdict(list)
+        for item in workflow_writes:
+            written_object = WrittenObject(item.object_name, item.update_mode)
+            writes_mapping[item.integration_id].append(written_object)
+        return writes_mapping
```
kenxu95, line 269: I feel like `save` is a better term we should use throughout, since it corresponds with our `.save()` method. "writes" has a more generic connotation.

sdk/aqueduct/api_client.py
eunice-chan: Ok, sounds good. Would updating the endpoint to `/api/workflow/%s/saved_objects` make it clearer?

cw75: Also reminder to refer to this doc for endpoint naming. Typically endpoints shouldn't consist of `_` and `objects` seems fine to me.

sdk/aqueduct/api_client.py
```diff
@@ -311,6 +313,14 @@ def get_workflow(self, flow_id: str) -> GetWorkflowResponse:
         workflow_response = GetWorkflowResponse(**resp.json())
         return workflow_response
 
+    def get_workflow_writes(self, flow_id: str) -> GetWorkflowWrittenObjectsResponse:
```
likawind, line 316: referring to our naming guide https://www.notion.so/aqueducthq/SDK-External-API-Naming-ce685c0980ae4d7ebee569bf53ef2a01 this should method should be named as `list_saved_objects`, and agree that it should belongs to `flow` object

likawind (changes requested):
Good work! Generally agree with other folks on naming. Based on https://github.com/aqueducthq/aqueduct/pull/241 , we could use `saved_objects`

sdk/aqueduct/api_client.py
eunice-chan: Thanks for the helpful resource @cw75. I'll keep the endpoint as "/api/workflow/%s/objects" and modify `GET_WORKFLOW_OBJECTS_ROUTE` to `LIST_WORKFLOW_OBJECTS_ROUTE` to be consistent with the url & @likawind 's suggestion to rename the function to `list_saved_objects` as per the [naming guide](https://www.notion.so/aqueducthq/SDK-External-API-Naming-ce685c0980ae4d7ebee569bf53ef2a01).

eunice-chan:
Thanks all for your comments. Based on these, I am thinking of doing the following:
1. Renaming & moving `client.get_workflow_written(...)` --> `flow.list_saved_objects(...)`
2. In `list_saved_objects`, change the key of the returned dictionary to be the `Integration` object (or `IntegrationInfo` object) to be more intelligible for the user than just the `integration_id`. (**would like your thoughts on this**)
3. Rename `GET_WORKFLOW_OBJECTS_ROUTE` to `LIST_WORKFLOW_OBJECTS_ROUTE`
4. Update https://github.com/aqueducthq/aqueduct/pull/239 and the other PRs to use the Save/SavedObject(s) terminology rather than Write/WrittenObject(s) e.g. `saved_objects_to_delete`.

Please don't hesitate to give me any feedback. Particularly, would like to know your thoughts on item 2.

likawind (approved):
LGTM, thanks for addressing all comments!

sdk/aqueduct/integrations/integration.py
```diff
@@ -31,3 +32,13 @@ class Integration(ABC):
     """
 
     _metadata: IntegrationInfo
+
```
eunice-chan, line 35: May want to add a comment about why we do this.

PR #241, opened 2022-07-22
Eng 1369 allow users to delete workflows and associated written objects in SDK

eunice-chan:
Add the ability to delete any objects written by the workflow when deleting the workflow in the SDK.

ENG 1394
ENG-1369 (builds on above)
ENG-1325 (builds on above)
ENG-1431 (builds on above)

    - **Integration tests won't work because BE not set up. Added some extra data to be sent to the delete workflow endpoint and as a result the current endpoint fails. See the PR with the BE part for the integration tests**
    - **Integration tests won't work because BE not set up. Added some extra data to be sent to the delete workflow endpoint and as a result the current endpoint fails. See the PR with the BE part for the integration tests**
    - **Integration tests won't work because BE not set up. Added some extra data to be sent to the delete workflow endpoint and as a result the current endpoint fails. See the PR with the BE part for the integration tests**


sdk/aqueduct/aqueduct_client.py
```diff
@@ -417,12 +418,21 @@ def trigger(
         flow_id = parse_user_supplied_id(flow_id)
         self._api_client.refresh_workflow(flow_id, serialized_params)
 
-    def delete_flow(self, flow_id: Union[str, uuid.UUID]) -> None:
+    def delete_flow(
+        self,
+        flow_id: Union[str, uuid.UUID],
+        writes_to_delete: Optional[DefaultDict[uuid.UUID, List[WrittenObject]]] = None,
```
likawind, line 424: I feel we could relate to 'save' which is a concept available for the rest of sdk. For example, `saved_objects_to_delete`. cc @kenxu95 @vsreekanti

sdk/aqueduct/aqueduct_client.py
```diff
@@ -417,12 +418,21 @@ def trigger(
...
+        force: bool = False,
+    ) -> Dict[uuid.UUID, List[WritesDelete]]:
```
likawind, line 426: Should we return any objects here? I don't think if the user would actually programmatically use this returned value  esp. that it's also an `uuid->list` mapping. If it's not intended for user to use the returned value, we should return None here (and it's also generally natural that a `delete` returns `None`)

It should be sufficient that we print out any deletion failures for saved object. We may actually want to raise exceptions in these failures.

sdk/aqueduct/aqueduct_client.py
```diff
@@ -431,11 +441,31 @@ def delete_flow(self, flow_id: Union[str, uuid.UUID]) -> None:
...
+                    failed_for_integration.append(obj)
+            if len(failed_for_integration) > 0:
+                failed_deletions[str(integration)] = [
+                    res.__dict__ for res in failed_for_integration
+                ]
+                counts += len(failed_for_integration)
+        if counts > 0:
+            print("Workflow-Written Objects' Deletion Failures")
+            print(f"{counts} Failures")
+            print(json.dumps(failed_deletions, sort_keys=False, indent=4))
```
likawind, line 467: this will print the UUID as keys. I feel we would also want to print the integration name instead.

likawind (changes requested):
Overall looks good! Some comments around naming and user messaging

sdk/aqueduct/aqueduct_client.py
kenxu95: yeah thats fine with me. I agree, 'save' exposes the user to the same terminology as our API.

sdk/aqueduct/aqueduct_client.py
eunice-chan: Returning None now and raising an exception if any table failed to delete

likawind (approved):

PR #245, opened 2022-07-22
Add warning color to DAG in UI

kenxu95:
When a warning check fails to pass, we mark the artifact it outputs in a new warning color: orange.

Also, changes the color palette to something with more contrast (cc: @agiron123 as color picker) 

~Examples below:
I realize these are currently inconsistent as to whether the check operator or the check artifact are highlighted - curious as to what the right colors are in each case. The error check fails the entire workflow, so having the operator be red seems to make sense. I'm thinking I also make the warning operator orange? Or maybe in the future we should squash the check operator and artifact concepts into one. @vsreekanti~

image

ENG-1265


vsreekanti:
I think we can/should make the check artifact and the check operator both red, no? Alternately, we can make the check operator green (it executed successfully) and the check artifact red (it was bad). But my guess is that if a check fails, we should make the operator and artifact both red. @jegonzal @agiron123 any thoughts?

kenxu95:
Making both of them red is inconsistent with the way we currently do things - if an operator fails, then the artifact is never written and so stays in a pending state, where they can't see the data. Do we want to try and change this larger pattern? I have no idea how hard that would be.

And I don't know if marking only the artifact as red makes sense, because the operator needs to fail on error-severity checks to mark the entire workflow as failed. It would require some custom logic in the UI just for checks in order to pretend like the check operator passed but the artifact failed.

I feel like marking both the warning check operator and artifact as orange is both the easiest and most consistent (warning   is equivalent to success for all intents and purposes, except for the fact that a warning will show on the sidebar).

vsreekanti:
So if a check returns `False` the operator fails, the result will not be written?

kenxu95:
Only if the check has ERROR severity, yes.

vsreekanti:
Sorry, I think I'm being vague. I was asking if the `False` value would be persisted, _not_ the artifact that the check is attached to.

kenxu95:
By attached to you mean the artifact that is the input of the check operator? I think I was interpreting the question correctly then - but I just checked the code and it looks like we do persist the `False` in either case.

src/ui/common/src/components/workflows/StatusBar.tsx
```diff
@@ -371,10 +371,24 @@ export const WorkflowStatusBar: React.FC<WorkflowStatusBarProps> = ({
         type: 'tableArtifact',
       };
 
+      // Custom logic for check artifacts that display a warning message
+      // if the check did not pass, even if the check was successfully executed.
+      let isCheckArtifactAndDidNotPass = false;
+      if (artifacts[artifactId].spec.type == ArtifactType.Bool) {
+        const status = artifactResult.result?.data as CheckStatus;
+        if (status == CheckStatus.Failed) {
```
agiron123, line 379: please use === when doing equality checks in JavaScript

integration_tests/sdk/error_handling_test.py
```diff
@@ -37,7 +37,8 @@ def test_handle_bad_op_error(client):
     sql_artifact = db.sql(query=GOOD_QUERY)
     processed_artifact = bad_op(sql_artifact)
 
-    try:
-        processed_artifact.get()
-    except AqueductError as e:
-        assert TIP_OP_EXECUTION in e.message
+    client.publish_flow("hello", artifacts=[processed_artifact])
+    # try:
```
agiron123, line 41: Is there a reason why you are leaving this code commented out?

agiron123:
> I think we can/should make the check artifact and the check operator both red, no? Alternately, we can make the check operator green (it executed successfully) and the check artifact red (it was bad). But my guess is that if a check fails, we should make the operator and artifact both red. @jegonzal @agiron123 any thoughts?

@vsreekanti I think both of them being red makes more sense since the two nodes are related to one another.

agiron123:
Left a few comments on the code, but should be good to go after you resolve them @kenxu95

agiron123 (approved):
Good to go after comments are resolved.

src/ui/common/src/components/workflows/nodes/Node.tsx
```diff
@@ -50,36 +51,60 @@ export const Node: React.FC<Props> = ({
...
-      textColor = 'green.800';
+      backgroundColor = selected
+        ? theme.palette.DarkSuccessMain50
+        : theme.palette.DarkSuccessMain;
+      hoverColor = theme.palette.DarkSuccessMain75;
       break;
     case CheckStatus.Failed:
+      // We use yellow to denote a check warning.
+      const checkOpID = artifactIDToInputOpID(data.nodeId);
+      if (
```
agiron123, line 79: ===

kenxu95:
Ok I just read a bit more of the frontend code and there's quite a bit of entanglement here due to how much we treat checks as a special case through out codebase. For me to implement the best solution I need to know:

1) What are the principles behind the color coding of failed operators? If we decide to do red-red here, but other non-check operators stay red-gray on failure, is that inconsistency ok? Or should I simply keep it consistent and do red-gray for everything. Or should I try and change everything to red-red.

2) For warning checks, should I keep it orange-orange, with the principle that orange is really the same as green, except that it comes with a warning message on the sidebar?

3) When should the user see the value "False" when the select the check artifact? Should they always see it, or only when the check has error severity WARNING. If they always see it, we would once again be inconsistent with how non-check operators work. When you select a non-check artifact of a failed operator, you currently get:
image

vsreekanti:
@kenxu95, let me know if you have any more questions about our conversations on Friday or if you feel like you have enough direction to wrap this up!

kenxu95:
oh sry meant to update this @vsreekanti I think I have enough direction. Chatted with Wei about this today and decided that a lot of this is better implemented on the backend, since the UI logic was getting super confusing. The backend stuff was done today and I'm currently polishing this up.

agiron123 (approved):

PR #251, opened 2022-07-25
Package lock changes to bump @aqueducthq/common to version 0.0.10.

agiron123:
## Describe your changes and why you are making these changes
After publishing a release, we need to also publish the latest changes to the @aqueducthq/common package for usage in our enterprise project.

I've gone ahead and ran ```npm publish``` to publish the latest version, v0.0.10. 

This PR updates package-lock.json files to reflect the latest version.

## Checklist before requesting a review
- [x] I have created a descriptive PR title. The PR title should complete the sentence "This PR...".
- [x] I have performed a self-review of my code.
- [x] I have included a small demo of the changes. For the UI, this would be a screenshot or a Loom video.
- [x] If this is a new feature, I have added unit tests and integration tests.
- [x] I have run the integration tests locally and they are passing.
- [x] I have run the linter script locally (See `python3 scripts/run_linters.py -h` for usage).
- [x] All features on the UI continue to work correctly.
- [x] Added one of the following CI labels:
    - `run_integration_test`: Runs integration tests
    - `skip_integration_test`: Skips integration tests (Should be used when changes are ONLY documentation/UI)

PR #254, opened 2022-07-26
[UI] Show workflows associated with an integration in integration details page

likawind:
This PR adds a 'workflow' section in integration details page to show all workflows for the integration. Specifically, we show an accordion list of workflows, and for each item we show all operators in the workflow on this integration.

We added a few components:
* An operator table for showing list of oeprators
* An workflow accordion to show an expandable accordion together with the above operator table
* An 'OperatorsOnIntegration' to show the list of above workflows.

ENG-759

Update:
* gray out operators in older version
* use buttons for expand / hide old operators, as well as workflow link
https://www.loom.com/share/e4fbdd3c3ec44cfb8f19901a1f459e9a

First version demo:
https://www.loom.com/share/e9d1b3781da540a4b80832d5c82652ab

### DIFF
--- a/src/ui/common/package-lock.json
+++ b/src/ui/common/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@aqueducthq/common",
-  "version": "0.0.8",
+  "version": "0.0.9",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@aqueducthq/common",
-      "version": "0.0.8",
+      "version": "0.0.9",
       "hasInstallScript": true,
       "devDependencies": {
         "@babel/preset-react": "^7.17.12",
--- a/src/ui/common/package-lock.json
+++ b/src/ui/common/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@aqueducthq/common",
-  "version": "0.0.9",
+  "version": "0.0.10",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@aqueducthq/common",
-      "version": "0.0.9",
+      "version": "0.0.10",
       "hasInstallScript": true,
       "devDependencies": {
         "@babel/preset-react": "^7.17.12",
--- a/src/ui/common/src/components/integrations/operatorsOnIntegration.tsx
+++ b/src/ui/common/src/components/integrations/operatorsOnIntegration.tsx
@@ -2,7 +2,7 @@ import Alert from '@mui/material/Alert';
 import AlertTitle from '@mui/material/AlertTitle';
 import Box from '@mui/material/Box';
 import CircularProgress from '@mui/material/CircularProgress';
-import React from 'react';
+import React, { useState } from 'react';
 import { useSelector } from 'react-redux';
 
 import { OperatorsForIntegrationItem } from '../../reducers/integrationOperators';
@@ -15,11 +15,10 @@ const OperatorsOnIntegration: React.FC = () => {
   const listWorkflowState = useSelector(
     (state: RootState) => state.listWorkflowReducer
   );
-  const operatorsState = useSelector(
-    (state: RootState) => {
-        return state.integrationOperatorsReducer
-    }
-  );
+  const operatorsState = useSelector((state: RootState) => {
+    return state.integrationOperatorsReducer;
+  });
+  const [expandedWf, setExpandedWf] = useState('');
 
   if (
     isLoading(operatorsState.loadingStatus) ||
@@ -69,6 +68,14 @@ const OperatorsOnIntegration: React.FC = () => {
     <Box>
       {Object.entries(operatorsByWorkflow).map(([wfId, item]) => (
         <WorkflowAccordion
+          expanded={expandedWf === wfId}
+          handleExpand={() => {
+            if (expandedWf === wfId) {
+              setExpandedWf('');
+              return;
+            }
+            setExpandedWf(wfId);
+          }}
           key={wfId}
           workflow={item.workflow}
           operators={item.operators}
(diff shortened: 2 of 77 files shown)


agiron123:
Nice work! When we show the older versions of workflows, would it be possible to show the time that the version was run? I'm noticing that there are older versions with the same name, and it's not so easy to distinguish which version is which.

src/ui/common/package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "@aqueducthq/common",
-  "version": "0.0.8",
+  "version": "0.0.9",
```
agiron123, line 3: Latest version here should be 0.0.10, I have another related PR that will probably go in before this one, so you will be able to just rebase and get latest changes.

https://github.com/aqueducthq/aqueduct/pull/251

src/ui/common/src/components/integrations/operatorsOnIntegration.tsx
```diff
@@ -0,0 +1,88 @@
...
+import WorkflowAccordion from '../workflows/accordion';
+
+const OperatorsOnIntegration: React.FC = () => {
+  const listWorkflowState = useSelector(
+    (state: RootState) => state.listWorkflowReducer
+  );
+  const operatorsState = useSelector((state: RootState) => {
+    return state.integrationOperatorsReducer;
+  });
+  const [expandedWf, setExpandedWf] = useState('');
```
agiron123, line 21: For this useState, can you add a type for expanedWf please?

Will look something like this:
```
const [expandedWf, setExpandedWf] = useState<string>();
```

agiron123 (approved):
Nice work! Left some small comments but you can resolve before merging.

Let's coordinate a time to update the @aqueducthq/common package offline. We'll need to bump up the version number since you have made some changes to it so our enterprise edition can get these changes as well.

vsreekanti:
Thanks for the demo @likawind! A few things:
* I think it's a little confusing that the accordion expands and is clickable to go to another page. I'd prefer if clicking anywhere in the accordion causes it to expand and then there's a "Go to workflow details" button at the bottom that links to the workflow details page.
* @agiron123, any wisdom on whether table columns should be all left-aligned, right-aligned, or mixed (as in the current impl.)?
* Tiny nit: Let's remove the divider below the last entry in the table. 
* For the historical artifacts, let's change the text to be "Show operators from previous versions" and make it a button rather than plaintext. 
  * I'm open to suggestions on the exact implementation, but it's confusing to see three operators with the same name and mo differentiation. I'd like to do one or both of: (1) grey out the historical operators; (2) add a timestamp for when the operator version was created. 

@agiron123, feel free to add or disagree with any of the design suggestions here.

likawind:
Thanks @agiron123 and @vsreekanti 
- for showing old operators, I've updated the screenshot to show older ones in lighter color. Adding a timestamp would also work, I would do that in a separate follow-up as it also require some backend changes.
- I borrowed table alignment from MUI's demo: https://mui.com/material-ui/react-table/#main-content . They generally put 'index' column to be left-aligned and the 'body' columns to be right-aligned. Which I found it makes sense

likawind:
@agiron123 we have a few UI changes in the future, do you think if we should bump the package version after all changes are in-place?

agiron123:
@likawind yeah we should wait a little for changes to be in place. 

We can go ahead and bump the number now though so we can continue our development process using
```npm run link``` and ```npm link@aqueducthq/common``` when installing from app.

We just need to bump the version number to get the latest changes from the link. (I'm sure there's probably another way to do so, but I always have trouble unless I just bump the version up)

likawind:
@agiron123 If I understand you would suggest to bump to v10 since you already did so (I believe I could do so by rebase), and we should wait for other future changes to bump to v11?

likawind:
@vsreekanti @agiron123 this is the updated look:
https://www.loom.com/share/e4fbdd3c3ec44cfb8f19901a1f459e9a

vsreekanti:
This looks great -- thanks @likawind!

agiron123:
I think this looks very nice too!

One thing I'd love to do in our future redesign work is give better visibility into our operators, which we now have a nice API for. 

I think putting them on some sidebar in the Workflows page would be a nice start :)

agiron123:
> Thanks for the demo @likawind! A few things:
> 
> * I think it's a little confusing that the accordion expands and is clickable to go to another page. I'd prefer if clicking anywhere in the accordion causes it to expand and then there's a "Go to workflow details" button at the bottom that links to the workflow details page.
> * @agiron123, any wisdom on whether table columns should be all left-aligned, right-aligned, or mixed (as in the current impl.)?
> * Tiny nit: Let's remove the divider below the last entry in the table.
> * For the historical artifacts, let's change the text to be "Show operators from previous versions" and make it a button rather than plaintext.
>   
>   * I'm open to suggestions on the exact implementation, but it's confusing to see three operators with the same name and mo differentiation. I'd like to do one or both of: (1) grey out the historical operators; (2) add a timestamp for when the operator version was created.
> 
> @agiron123, feel free to add or disagree with any of the design suggestions here.

I'm definitely in favor of centering things. We don't do this currently in our app, but I'd like us to start doing so in our redesign.

agiron123 (approved):

likawind:
@vsreekanti @agiron123 here's how centering every column looks like:
Screen Shot 2022-07-27 at 12 05 55 PM

I don't have preference about the 'body' part, but centering the index makes the alignment looks a bit off. I think we should at least left-align the index

likawind:
if we left-align index and center-align body:

Screen Shot 2022-07-27 at 1 23 46 PM

likawind:
I think we've using left-align for data table, so we could also be consistent with that:
Screen Shot 2022-07-27 at 2 55 56 PM

PR #257, opened 2022-07-26
Laying the foundation for multi type support

cw75:
This PR lays our foundation for supporting various types of artifacts. Specifically, the PR:
1. Outlines the enum definition of all artifact types that we want to support.
2. Implements type-specific serialization logic for both the backend and the sdk.
3. Modifies the Golang side's type structs to work with multiple artifact types.
4. Implements a skeleton for the `UntypesArtifact` on the sdk side.

This PR by itself won't work with our existing tests. Instead we will use this as a starting point to add:
1. [Eager execution](https://www.notion.so/aqueducthq/SDK-API-design-for-eager-lazy-execution-71e80bfcb475469495a40ed2283552a9).
2. Caching.
3. Type enforcement.

I'm configuring the PR to be merged to a branch `non_tabular_support` that I created off of `main`, and once everything is working we can merge this to `main`.


sdk/aqueduct/deserialize.py
```diff
@@ -0,0 +1,32 @@
...
+
+def read_pickle_content(content: bytes) -> Any:
+    return pickle.loads(content)
+
+
+def read_image_content(content: bytes) -> Image.Image:
+    return Image.open(io.BytesIO(content))
+
+
+def read_standard_content(content: bytes) -> str:
```
kenxu95, line 27: as a user I feel like I'd be confused about what a "standard" type is. Isn't this just `read_string_content`

sdk/aqueduct/enums.py
```diff
@@ -128,8 +128,24 @@ class DisplayNodeType(str, Enum, metaclass=MetaEnum):
...
+    PICKLABLE = "picklable"
+
+class SerializationType(str, Enum, metaclass=MetaEnum):
+    TABULAR = "tabular"
+    JSON = "json"
+    PICKLE = "pickle"
+    IMAGE = "image"
+    STANDARD = "standard"
+    BYTES = "bytes"
+
```
kenxu95, line 151: It's not clear to me as a reader what the relationship between artifact type and serialization type is. Is it a many-to-many mapping? Is there only one possible serialization for each type? Can we make this explicit to a reader with a map of [ArtifactType] -> List[SerializationType] denoting what serializations we support for each type? In addition to being useful documentation, we can use this map as validation before sending stuff to the backend too.

sdk/aqueduct/enums.py
```diff
@@ -128,8 +128,24 @@ class DisplayNodeType(str, Enum, metaclass=MetaEnum):
...
+    BYTES = "bytes"
+    IMAGE = "image"
+    PICKLABLE = "picklable"
+
+class SerializationType(str, Enum, metaclass=MetaEnum):
+    TABULAR = "tabular"
+    JSON = "json"
+    PICKLE = "pickle"
+    IMAGE = "image"
+    STANDARD = "standard"
```
kenxu95, line 149: same possible confusion about "standard"

sdk/aqueduct/integrations/s3_integration.py
```diff
@@ -53,7 +55,7 @@ def file(
                 Description of the query.
```
kenxu95, line 55: Can we update the docstring here? Specifically the requirement between artifact_type and format.

sdk/aqueduct/responses.py
```diff
@@ -51,39 +51,9 @@ class OperatorResult(BaseModel):
...
-class ParamArtifactResult(BaseModel):
-    val: str
-
-
 class ArtifactResult(BaseModel):
-    table: Optional[TableArtifactResult]
-    metric: Optional[MetricArtifactResult]
-    check: Optional[CheckArtifactResult]
-    param: Optional[ParamArtifactResult]
+    serialization_type: str
```
kenxu95, line 55: shouldn't this be an enum?

sdk/aqueduct/utils.py
```diff
@@ -125,7 +125,7 @@ def retention_policy_from_latest_runs(k_latest_runs: int) -> RetentionPolicy:
 REQUIREMENTS_FILE = "requirements.txt"
 BLACKLISTED_REQUIREMENTS = "aqueduct"
 
-UserFunction = Callable[..., pd.DataFrame]
+UserFunction = Callable[..., Any]
```
kenxu95, line 128: Unrelated but just a thought: for the metric and check functions, if they return a non-traditional value like np.bool_, will we cast their return values into a bool for them? Or will we use the raw type - I think we cast atm.

src/golang/cmd/server/handler/preview.go
```diff
@@ -50,28 +49,9 @@ type previewArgs struct {
...
-type previewBoolArtifactResponse struct {
-	Passed bool `json:"passed"`
-}
-
-type previewParamArtifactResponse struct {
-	Val string `json:"val"`
-}
-
-type previewTableArtifactResponse struct {
-	TableSchema []map[string]string `json:"table_schema"`
```
kenxu95, line 66: just to confirm, we're intentionally getting rid of TableSchema here right?

src/golang/cmd/server/handler/preview.go
```diff
@@ -211,11 +191,21 @@ func (h *PreviewHandler) Perform(ctx context.Context, interfaceArgs interface{})
...
 	for _, artf := range workflowDag.Artifacts() {
 		if artf.Computed(ctx) {
-			artifactResp, err := convertToPreviewArtifactResponse(ctx, artf)
+			artifact_metadata, err := artf.GetMetadata(ctx)
 			if err != nil {
 				return errorRespPtr, http.StatusInternalServerError, err
 			}
-			artifactResults[artf.ID()] = *artifactResp
+
+			log.Info(artifact_metadata.SerializationType)
```
kenxu95, line 199: can we use the Logger.logger object here? It's configured by the Client constructor.

kenxu95 (commented):
I've only gotten through the SDK part, will do a backend review after lunch.

Why do you say the existing tests won't pass? I'm not sure what SDK semantics have changed so much that we can no longer pass tests.

src/golang/lib/collections/artifact/types.go
```diff
@@ -25,129 +19,21 @@ import (
...
+	BytesType     Type = "bytes"
+	ImageType     Type = "image"
+	PicklableType Type = "picklable"
 )
 
-type specUnion struct {
-	Type  Type         `json:"type"`
-	Table *table.Table `json:"table,omitempty"`
-
-	// TODO(ENG-1119): The float artifact currently also represents integers.
```
kenxu95, line 38: Should we cancel ENG-1119 then?

src/golang/lib/collections/operator/connector/extract_params.go
```diff
@@ -39,8 +39,10 @@ type SalesforceExtractParams struct {
 }
 
 type S3ExtractParams struct {
-	Filepath string `json:"filepath"`
-	Format   string `json:"format"`
+	Filepath     string `json:"filepath"`
+	ArtifactType string `json:"artifact_type"`
+	Format       string `json:"format,omitempty"`
+	Merge        bool   `json:"merge,omitempty"`
```
kenxu95, line 45: Sort of unrelated, but I wonder hard would it be to connect our backend to a demo S3 bucket? It doesn't have to be actually in S3, we can write our own local version of the integration. This would make it possible to test S3 integrations from the SDK integration tests right? This seems like valuable coverage.

src/python/aqueduct_executor/operators/connectors/tabular/connector.py
```diff
@@ -1,11 +1,12 @@
...
-from typing import List
+from typing import Any, List
 
 import pandas as pd
 from aqueduct_executor.operators.connectors.tabular import extract, load
+from aqueduct_executor.operators.utils.enums import ArtifactType
 
 
-class TabularConnector(ABC):
+class StorageConnector(ABC):
```
kenxu95, line 9: Is "storage connector" the terminology we're going with for all our non-github integrations? It's a little confusing for me as a developer since we already use the term "storage" a lot in the backend to refer to our result stores (eg. we write output contents to "storage" at output_content_path)

src/python/aqueduct_executor/operators/connectors/tabular/s3.py
```diff
@@ -29,58 +36,121 @@ def authenticate(self) -> None:
...
+                return pd.read_json(buf)
+            elif params.format == common.S3TabularFormat.PARQUET:
+                return pd.read_parquet(buf)
+            raise Exception("Unknown S3 file format %s." % params.format)
+        elif params.artifact_type == ArtifactType.JSON:
+            # This assumes that the encoding is "utf-8". May worth considering letting the user
+            # specify custom encoding in the future.
+            json_data = data.decode(_DEFAULT_JSON_ENCODING)
+            # Make sure the data is a valid json object.
+            json.loads(json_data)
```
kenxu95, line 58: we should probably raise a nicer error message if the data is not json - make it really clear to them that the data in their S3 is not in json format.

src/python/aqueduct_executor/operators/connectors/tabular/s3.py
```diff
@@ -29,58 +36,121 @@ def authenticate(self) -> None:
...
+            if params.artifact_type == ArtifactType.STRING:
+                assert(isinstance(unpickled_data, str))
+            elif params.artifact_type == ArtifactType.BOOL:
+                assert(isinstance(unpickled_data, bool) or isinstance(unpickled_data, np.bool_))
+            elif params.artifact_type == ArtifactType.NUMERIC:
+                assert(isinstance(unpickled_data, int) or isinstance(unpickled_data, float) or isinstance(unpickled_data, np.number))
+            elif params.artifact_type == ArtifactType.DICT:
+                assert(isinstance(unpickled_data, dict))
+            elif params.artifact_type == ArtifactType.TUPLE:
+                assert(isinstance(unpickled_data, tuple))
```
kenxu95, line 81: I don't think these should be assertions, since they rely on user inputs, so they're not validating internal consistency without our system. Can we raise an exception with a useful error message?

src/python/aqueduct_executor/operators/function_executor/execute.py
```diff
@@ -66,11 +71,41 @@ def _import_invoke_method(spec: FunctionSpec) -> Callable[..., DataFrame]:
...
+    elif isinstance(result, str):
+        # We first check if the result is a valid JSON string.
+        try:
+            json.loads(result)
+            return ArtifactType.JSON
+        except:
+            return ArtifactType.STRING
+    elif isinstance(result, bool) or isinstance(result, np.bool_):
+        return ArtifactType.BOOL
+    elif isinstance(result, int) or isinstance(result, float) or isinstance(result, np.number):
```
kenxu95, line 90: These if statements are shared with the extraction code right? Can we put them in a shared helper to prevent divergence?

sdk/aqueduct/deserialize.py
cw75: Yeah we've been using this serialization mode for JSON and string types, so calling the serialization method "string" does seem more appropriate.

sdk/aqueduct/enums.py
cw75: Each artifact type corresponds to exactly one serialization method. We aren't sending these to the backend right? It's the other way around: the backend is sending the preview response to the SDK, with the type info as a hint so that SDK knows which serialization method to use. The map you suggested is still useful because we can replace the `if-else` in `_deserialize_content` function of the untyped artifact class with this map.

sdk/aqueduct/enums.py
cw75: yeah let's rename to "string"

sdk/aqueduct/responses.py
cw75: good catchl

sdk/aqueduct/utils.py
cw75: casting seems fine to me.

src/golang/cmd/server/handler/preview.go
cw75: This info doesn't seem to be used by the SDK, so I got rid of this.

src/golang/cmd/server/handler/preview.go
cw75: oops sorry I should have removed this. It's more of a debug logging.

src/golang/lib/collections/artifact/types.go
cw75: Yeah I can this task should be deprecated.

src/golang/lib/collections/operator/connector/extract_params.go
cw75: Yes I like this! Could you create a task?

src/python/aqueduct_executor/operators/connectors/tabular/connector.py
cw75: That was my intention. `TabularConnector` no longer make sense because in case of S3 the file doesn't necessarily need to be tabular any more. Any better naming ideas?

src/python/aqueduct_executor/operators/connectors/tabular/s3.py
cw75: Let me check what the error message is if we don't do anything. If it's confusing I can add a catch to raise a nicer one.

src/python/aqueduct_executor/operators/connectors/tabular/s3.py
cw75: Good catch, will fix!

src/python/aqueduct_executor/operators/function_executor/execute.py
cw75: I remember there's a bit of a difference in the use case, but will try to see if we can consolidate them.

cw75:
@kenxu95 Responded to your comments and they all make sense. For the tests, I was mainly concerned about those that use syntax candies like `head`. Because now we return an `UntypedArtifact`, it doesn't have access to these functions any more.

sdk/aqueduct/enums.py
kenxu95: right, and I think it would help a lot as a "documentation in code" especially. Can we also have something similar on the go side?

src/golang/lib/collections/operator/connector/extract_params.go
kenxu95: https://linear.app/aqueducthq/issue/ENG-1469/setup-a-demo-s3-bucket-for-our-integration-tests

src/python/aqueduct_executor/operators/connectors/tabular/connector.py
kenxu95: `DataConnector` actually seems a little more specific for me (since I would think code != data) and this step-sides the "storage" naming collision. Essentially "storage" can be something we own (eg. our S3, our filesystem) whereas "data" is always coming from the user's data integrations.

src/python/aqueduct_executor/operators/function_executor/execute.py
cw75: They look similar but don't serve the same goal, plus the one in the connector side is only a subset of the one in the function executor, so I think having them separated makes more sense now.

PR #259, opened 2022-07-26
Fix a bug where we couldn't execute imported functions

likawind:
This PR fixes the bug that we couldn't execute imported functions like the following:
```
from external import func

@aqueduct.op(file_dependencies=["external.py"])
def copy_field(df):
    return func(df)

copy_field(inputs).get()
```

Specifically, we fixed the `sys.path` which we used to import user functions to make sure the dependencies also properly appears in `sys.path`. We also added a few debugging logs for the extracted operator files.

ENG-1451

Verified the code snippet above works after fix
- [na] If this is a new feature, I have added unit tests and integration tests.
- [na] (in-PR tests should be sufficient) I have run the integration tests locally and they are passing.

### DIFF
--- a/src/python/aqueduct_executor/operators/function_executor/execute.py
+++ b/src/python/aqueduct_executor/operators/function_executor/execute.py
@@ -57,7 +57,7 @@ def _import_invoke_method(spec: FunctionSpec) -> Callable[..., DataFrame]:
     work_dir = os.path.join(fn_path, OP_DIR)
     print(f"listdir(workdir): {os.listdir(work_dir)}")
     print(f"listdir(fn_path): {os.listdir(fn_path)}")
-    
+
     # this ensures any file manipulation happens with respect to work_dir
     os.chdir(work_dir)
     # adds work_dir to sys.path to support relative imports from work_dir
--- a/src/python/aqueduct_executor/operators/function_executor/execute.py
+++ b/src/python/aqueduct_executor/operators/function_executor/execute.py
@@ -55,10 +55,8 @@ def _import_invoke_method(spec: FunctionSpec) -> Callable[..., DataFrame]:
 
     # work_dir should be `<storage>/operators/<id>/op`
     work_dir = os.path.join(fn_path, OP_DIR)
-    print("listdir(workdir)")
-    print(os.listdir(work_dir))
-    print("listdir(fn_path)")
-    print(os.listdir(fn_path))
+    print(f"listdir(workdir): {os.listdir(work_dir)}")
+    print(f"listdir(fn_path): {os.listdir(fn_path)}")
     
     # this ensures any file manipulation happens with respect to work_dir
     os.chdir(work_dir)
--- a/src/python/aqueduct_executor/operators/function_executor/execute.py
+++ b/src/python/aqueduct_executor/operators/function_executor/execute.py
@@ -40,17 +40,32 @@ def _get_py_import_path(spec: FunctionSpec) -> str:
 
     if file_path.startswith("/"):
         file_path = file_path[1:]
-    return ".".join([OP_DIR, file_path.replace("/", ".")])
+    return file_path.replace("/", ".")
 
 
 def _import_invoke_method(spec: FunctionSpec) -> Callable[..., DataFrame]:
+    # `_import_invoke_method` imports the model object.
+    # it assumes the operator has been extracted to `<storage>/operators/<id>/op`
+    # and imports the route from the above path.
+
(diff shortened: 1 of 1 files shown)


src/python/aqueduct_executor/operators/function_executor/execute.py
```diff
@@ -40,17 +40,32 @@ def _get_py_import_path(spec: FunctionSpec) -> str:
...
+    # and imports the route from the above path.
+
+    # fn_path should be `<storage>/operators/<id>`
     fn_path = spec.function_extract_path
-    os.chdir(os.path.join(fn_path, OP_DIR))
-    sys.path.append(fn_path)
+
+    # work_dir should be `<storage>/operators/<id>/op`
+    work_dir = os.path.join(fn_path, OP_DIR)
+    print("listdir workdir")
```
cw75, line 56: Let's be a bit more descriptive here: "Listing files under the current working directory." The reason being I feel others may have a hard time understanding what `listdir workdir` means.

src/python/aqueduct_executor/operators/function_executor/execute.py
```diff
@@ -40,17 +40,32 @@ def _get_py_import_path(spec: FunctionSpec) -> str:
...
     if file_path.startswith("/"):
         file_path = file_path[1:]
-    return ".".join([OP_DIR, file_path.replace("/", ".")])
+    return file_path.replace("/", ".")
 
 
 def _import_invoke_method(spec: FunctionSpec) -> Callable[..., DataFrame]:
+    # `_import_invoke_method` imports the model object.
+    # it assumes the operator has been extracted to `<storage>/operators/<id>/op`
+    # and imports the route from the above path.
```
kenxu95, line 49: what does "route" mean here?

nit: can we use the """ format for docstrings? I looks like its the standard and also looks way better on my IDE :D https://peps.python.org/pep-0257/

src/python/aqueduct_executor/operators/function_executor/execute.py
```diff
@@ -40,17 +40,32 @@ def _get_py_import_path(spec: FunctionSpec) -> str:
...
     fn_path = spec.function_extract_path
-    os.chdir(os.path.join(fn_path, OP_DIR))
-    sys.path.append(fn_path)
+
+    # work_dir should be `<storage>/operators/<id>/op`
+    work_dir = os.path.join(fn_path, OP_DIR)
+    print("listdir workdir")
+    print(os.listdir(work_dir))
+    print("listdir fn path")
+    print(os.listdir(fn_path))
```
cw75, line 59: It seems that the purpose of `fn_path` is just to construct the `work_dir` which is what we need, so I don't see a strong motivation for us to print out `fn_path` anymore.

src/python/aqueduct_executor/operators/function_executor/execute.py
```diff
@@ -40,17 +40,32 @@ def _get_py_import_path(spec: FunctionSpec) -> str:
...
+    os.chdir(work_dir)
+    sys.path.append(work_dir)
```
kenxu95, line 61: Can we comment this line? I think its important that a reader knows what the purpose of this specific line is, and what failure case can occur when its not included.

src/python/aqueduct_executor/operators/function_executor/execute.py
kenxu95, line 59: I have a tiny tiny readability preference to merge the prefix and the list into one line ala `"listdir(fn_path): %s" % os.listdir(fn_path)`

src/python/aqueduct_executor/operators/function_executor/execute.py
```diff
@@ -40,17 +40,32 @@ def _get_py_import_path(spec: FunctionSpec) -> str:
...
+
     import_path = _get_py_import_path(spec)
+    print(f"import_path: {import_path}")
```
cw75, line 64: I'd suggest using a full sentence for clarity: f"The model import path is: {import_path}."

cw75 (approved):
A few nits otherwise LGTM!

kenxu95 (approved):
yay! Left a few comments

src/python/aqueduct_executor/operators/function_executor/execute.py
Fanjia-Yan: I agree. I think it is kind of confusing what listdir and workdir represent, might need some documentation or clarification there.

Fanjia-Yan (approved):
Looks good to me! Thanks for the fix.

PR #265, opened 2022-07-28
Support operator with no input

cw75:
One way to enable this is to make module-level variables for the `dag` and `api_client` objects. The following change seems to work.

### DIFF
--- a/sdk/aqueduct/api_client.py
+++ b/sdk/aqueduct/api_client.py
@@ -186,7 +186,6 @@ class APIClient:
                 self._test_url(url)
                 return False
             except Exception as e:
-                print(e)
                 Logger.logger.info(
                     "Testing if connection is HTTP fails with:\n{}: {}".format(type(e).__name__, e)
                 )
--- a/sdk/aqueduct/api_client.py
+++ b/sdk/aqueduct/api_client.py
@@ -3,6 +3,7 @@ import json
 from typing import IO, Any, Dict, List, Optional, Tuple
 
 import requests
+from aqueduct._version import __version__
 from aqueduct.dag import DAG
 from aqueduct.enums import ExecutionStatus
 from aqueduct.error import (
@@ -23,14 +24,13 @@ from aqueduct.responses import (
 )
 
 from aqueduct import utils
-from aqueduct._version import __version__
-
 
 # Auth headers
 API_KEY_HEADER = "api-key"
 
 # Client version header
 CLIENT_VERSION = "sdk-client-version"
+
 
 def _handle_preview_resp(preview_resp: PreviewResponse, dag: DAG) -> None:
     """
--- a/sdk/aqueduct/api_client.py
+++ b/sdk/aqueduct/api_client.py
@@ -25,6 +25,8 @@ from aqueduct.responses import (
 from aqueduct import utils
 
 
+__GLOBAL_API_CLIENT__ = None
+
 def _handle_preview_resp(preview_resp: PreviewResponse, dag: DAG) -> None:
     """
     Prints all the logs generated during preview, in BFS order.
--- a/sdk/aqueduct/api_client.py
+++ b/sdk/aqueduct/api_client.py
@@ -23,7 +23,14 @@ from aqueduct.responses import (
 )
 
 from aqueduct import utils
+from aqueduct._version import __version__
 
+
+# Auth headers
+API_KEY_HEADER = "api-key"
+
+# Client version header
+CLIENT_VERSION = "sdk-client-version"
 
 def _handle_preview_resp(preview_resp: PreviewResponse, dag: DAG) -> None:
     """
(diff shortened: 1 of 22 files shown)


sdk/aqueduct/aqueduct_client.py
```diff
@@ -88,10 +90,12 @@ def __init__(
         """
         logging.basicConfig(level=log_level)
         self._api_client = APIClient(api_key, aqueduct_address)
+        api_client.__GLOBAL_API_CLIENT__ = self._api_client
         self._connected_integrations: Dict[
             str, IntegrationInfo
         ] = self._api_client.list_integrations()
         self._dag = DAG(metadata=Metadata())
+        dag.__GLOBAL_DAG__ = self._dag
```
kenxu95, line 98: the global api client is fine with me. But what happens if we user constructs multiple aqueduct clients? Do we want to wipe out the previous dag completely? I'd rather us not overwrite the global dag if it already exists, so we'd have two clients using the same underlying dag, which is weird but seems less destructive.

sdk/aqueduct/decorator.py
```diff
@@ -71,17 +73,19 @@ def wrap_spec(
...
 
-    dag = input_artifacts[0]._dag
-    api_client = input_artifacts[0]._api_client
+    if len(input_artifacts) == 0:
+        print("This operator takes no artifact input.")
+        dag = dag_module.__GLOBAL_DAG__
+        api_client = api_client_module.__GLOBAL_API_CLIENT__
+    else:
+        dag = input_artifacts[0]._dag
+        api_client = input_artifacts[0]._api_client
```
kenxu95, line 88: I'm not sure we should even keep the else statement once we have global variables.

sdk/aqueduct/decorator.py
```diff
@@ -71,17 +73,19 @@ def wrap_spec(
...
             raise InvalidUserActionException(
                 "Artifact fetched from flow run can not be reused for computation"
             )
 
-    dag = input_artifacts[0]._dag
-    api_client = input_artifacts[0]._api_client
+    if len(input_artifacts) == 0:
+        print("This operator takes no artifact input.")
+        dag = dag_module.__GLOBAL_DAG__
+        api_client = api_client_module.__GLOBAL_API_CLIENT__
```
kenxu95, line 85: There's another assumption we're sort of breaking here. Previously, all starting operators have to be constructed through the aqueduct client, so the client would have to be instantiated before running operators. Now, a user may expect their decorated function to run even without the client instantiated, and will be confused as to why the client needs to be constructed even if it's not directly used.

I don't think we can avoid this dependency right now, but what if we communicated this better by:
- initialize the global dag to `DAG(metadata=Metadata())` instead of None.
- initialize the global api client without the api_key and server_address. We can add a `configure(api_key, server_address)` method to set these fields up. If we attempt to use the api_client without configuring it, the user will get a very informative error about how they need to set the api_key and server_address first, currently by instantiating the client. Later on, when we have the concept of a `aq.config(...)` global config object, we can configure the api_client directly there.

cw75:
@kenxu95 addressed comments. I had to add dag variable reset logic to integration tests for some of them to pass. I think we had an assumption that when the client gets destroyed the dag var will be reset, which is no longer true, so I had to manually reset. I noticed that when I ran the integration test locally, "test_check_failure_with_varying_severity" is failing with a preview error, but it's not showing here. Not sure why.

cw75:
@kenxu95 added an integration test.

sdk/aqueduct/api_client.py
```diff
@@ -107,14 +107,17 @@ class APIClient:
...
+
+    def configure(self, api_key: str, aqueduct_address: str) -> None:
         self.api_key = api_key
         self.aqueduct_address = aqueduct_address
 
         # Clean URL
         if self.aqueduct_address.endswith("/"):
             self.aqueduct_address = self.aqueduct_address[:-1]
 
+        self.configured = True
```
kenxu95, line 120: Should configured be set at the end of this function? If we fail to validate the api_client, should we allow it to be "configured"?

We can get rid of the logic below, since that is actually solved by the constructor not taking in any arguments.
```
# If a dummy client is initialized, don't perform validation.
        if self.api_key == "" and self.aqueduct_address == "":
```

sdk/aqueduct/api_client.py
```diff
@@ -132,13 +135,22 @@ def __init__(self, api_key: str, aqueduct_address: str):
         else:
             self.use_https = self._test_connection_protocol(try_http=True, try_https=True)
 
+    def _check_config(self) -> None:
+        if not self.configured:
+            raise Exception(
+                "API client has not been configured, please complete the configuration \
+                by initializing an Aqueduct client with the api key and the server address."
+            )
```
kenxu95, line 143: Instead of having to label every single helper in this class, I think all we need to check configs for is:
`construct_base_url()`, `construct_full_url()`, `url_prefix()`, and `generate_auth_headers()`. The reason I think this makes more sense is because each of these helpers directly uses the methods we've set in `configure()`.

It would also be great to move `generate_auth_headers()` into the APIClient, I'm not sure why these were in utils to begin with.

sdk/aqueduct/tests/metric_test.py
```diff
@@ -28,14 +29,13 @@ def test_metric():
     op_id = generate_uuid()
     artifact_id = generate_uuid()
     artifact_name = "artifact"
-    api_client = APIClient("", "")
+    api_client.__GLOBAL_API_CLIENT__.configure("", "")
```
kenxu95, line 32: I don't think we need this line anymore right.

sdk/aqueduct/tests/utils.py
```diff
@@ -149,8 +147,8 @@ def default_table_artifact(
         inputs=[],
         outputs=[artifact_id],
     )
-    dag = _construct_dag(
+    dag_module.__GLOBAL_DAG__ = _construct_dag(
```
kenxu95, line 150: hmm because `_construct_operator()` is used by multiple test cases, I wonder if its safer to just use the local dag for the artifacts. These unit tests are not supposed to use decorators or the aqueduct client methods anyways, since its supposed to validate lower-level constraints.

kenxu95 (changes requested):
I have a few questions/suggestions but other than that it looks really good!

cw75:

sdk/aqueduct/api_client.py
```diff
@@ -166,6 +186,7 @@ def _test_connection_protocol(self, try_http: bool, try_https: bool) -> bool:
                 self._test_url(url)
                 return False
             except Exception as e:
+                print(e)
```
kenxu95, line 189: I think we can probably just do `logger.error()` here, which already includes the exception.

kenxu95 (approved):

PR #267, opened 2022-07-28
[ENG-1354, 1415] Fixes bottom sidesheet and ReactFlow alignment issues

vsreekanti:

This PR fixes two bugs around the layout of the ReactFlow canvas and the bottom sidesheet:
* The fix for ENG-1354 is around calling `fitView` at the right time. Because we were previously calling `fitView` on `componentDidMount`, it didn't actually work on the first load because the nodes and edges were loaded in _after_ the component was mounted. This PR changes the `fitView` effect so that it also is called when the node/edge positions are changed, which causes the view to be fit correctly (after a small lag) on first load.
* This PR also fixes ENG-1415 simplifies absolute positioning of the bottom side sheet so that it properly aligns with the ReactFlow viewport (see screenshot below): 

(Ignore the miscolored artifacts; that's because my local backend isn't updated.)
image

### DIFF
--- a/src/ui/common/src/components/layouts/sidebar/AqueductSidebar.tsx
+++ b/src/ui/common/src/components/layouts/sidebar/AqueductSidebar.tsx
@@ -151,9 +151,8 @@ export const AqueductSidebar: React.FC<Props> = ({
     // open bottom to top
     bottomAligned: {
       position: 'absolute',
-      right: bottomSideSheetOffset,
+      left: MenuSidebarOffset,
       bottom: 0,
-      mx: `${BottomSidebarMarginInPx}px`,
       transition: AllTransition,
       width: bottomSideSheetWidth,
       height: `${BottomSidebarHeightInPx}px`,
--- a/src/ui/common/src/components/workflows/ReactFlowCanvas.tsx
+++ b/src/ui/common/src/components/workflows/ReactFlowCanvas.tsx
@@ -1,4 +1,3 @@
-import { CircularProgress } from '@mui/material';
 import React, { useEffect } from 'react';
 import ReactFlow, {
   Node as ReactFlowNode,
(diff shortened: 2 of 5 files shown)


src/ui/common/src/components/workflows/ReactFlowCanvas.tsx
```diff
@@ -19,23 +18,23 @@ type ReactFlowCanvasProps = {
...
+  );
+
+  const { fitView, viewportInitialized } = useReactFlow();
   useEffect(() => {
     fitView();
-  }, []);
+  }, [dagPositionState]);
 
   useEffect(() => {
     // NOTE(vikram): There's a timeout here because there seems to be a
```
agiron123, line 40: timeout still needed?

src/ui/common/src/components/layouts/sidebar/AqueductSidebar.tsx
```diff
@@ -151,9 +151,8 @@ export const AqueductSidebar: React.FC<Props> = ({
     // open bottom to top
     bottomAligned: {
       position: 'absolute',
-      right: bottomSideSheetOffset,
+      left: MenuSidebarOffset,
```
agiron123, line 154: Not super urgent, but we'll need to add this to the EnterpriseMenuSidebar as well. we use different components on enterprise and OSS

agiron123 (approved):

agiron123:
Looks good, remember to also update the version number of @aqueducthq/common here as well. 

We can say something like ```v0.0.11-rc.0``` so as not to have too many version numbers.

Here's how the folks at emotion do it:
image

src/ui/common/src/components/workflows/ReactFlowCanvas.tsx
vsreekanti: I think this is still necessary. This has to do with a race condition between the CSS transition for the `div` and when the `fitView` call is made.

PR #266, opened 2022-07-28
Eng 1325 api route delete workflow associated

eunice-chan:
## Describe your changes and why you are making these changes
backend functionality of what is being surfaced in the SDK functionality here https://github.com/aqueducthq/aqueduct/pull/241.
Expand workflow deletion to include deletion of saved objects touched by the workflow.

## Related issue number (if any)
ENG 1394
ENG-1369 (builds on above)
ENG-1325 (builds on above)
ENG-1431 (builds on above)

## Checklist before requesting a review
- [x] I have created a descriptive PR title. The PR title should complete the sentence "This PR...".
- [x] I have performed a self-review of my code.
- [N/A] I have included a small demo of the changes. For the UI, this would be a screenshot or a Loom video.
- [x] If this is a new feature, I have added unit tests and integration tests.
- [x] I have run the integration tests locally and they are passing.
- [x] I have run the linter script locally (See `python3 scripts/run_linters.py -h` for usage).
- [N/A] All features on the UI continue to work correctly.
- Delete workflow functionality doesn't work anymore because it expects a json input with `external_delete` and `force`. Will be fixed once I make the UI delete workflow PR.
- [x] Added one of the following CI labels:
    - `run_integration_test`: Runs integration tests
    - `skip_integration_test`: Skips integration tests (Should be used when changes are ONLY documentation/UI)

PR #268, opened 2022-07-28
Eng 1431 testing delete saved objects + api endpoint to get saved objects associated with an integration

eunice-chan:
Testing https://github.com/aqueducthq/aqueduct/pull/266 functionality + adding an API endpoint to check the objects are indeed removed in the integration.

ENG 1394
ENG-1369 (builds on above)
ENG-1325 (builds on above)
ENG-1431 (builds on above)


likawind (commented):
@eunice-chan I'm a bit confused on the purpose of this handler. It seems that you are trying to verify if objects are properly deleted form the workflow? There are a few things I'm not sure:
* I think you already include failed deletion as a part of workflow deletion. Why do we need a separate handler here?
* if the checking is only associated with the workflow + integration, why do we only provide an integration ID in the parameter? Seems that it will retrieve objects from other workflows as well?

likawind:
After staring at it more I assume you want it for testing? In this case I'm curious if we could just use the 'discover' route

eunice-chan:
@likawind The `discover` route doesn't include tables that are created via workflows so it specifically filters out the tables I am interested in. However, I am able to reuse most of the existing code because `discover` gets all tables before filtering out tables that were created by a workflow.

likawind (approved):
I see, do you think if we could achieve this by adding a flag for discover route specifying whether we want to apply additional filters? On the other hand I think we can focus on main project delivery. So let's merge as-is to get features done. We can keep discuss on this and fix separately if necessary.

eunice-chan:
@likawind That's a good idea! I added a Linear task for this and will revisit once the core functionality is done.

PR #273, opened 2022-07-29
Refactor refresh_workflow() handler to use Orchestrator

Fanjia-Yan:
This PR refrator the `refresh_workflow()` method. The `refresh_workflow()` method takes in `WorkflowID` as an argument. We pass the `WorkflowID` into `ReadLatestWorkflowDagFromDatabase()` to get the `workflowdag` from the database. Then, we instantiate a new `WorkflowDag` base on the previous result and passes in as a parameter for creating `AqEngine`. With the `AqEngine`, we call `AqEngine.ExecuteWorkflow` to trigger the workflow again.

Note:
The linter did not fully passes because there are some style issue from the AqEngine.go
The integration test for "test_trigger_workflow" passes, but the entire integration test did not pass.

https://linear.app/aqueducthq/issue/ENG-1383/refactor-refresh-workflow-handler-to-use-orchestrator


src/golang/cmd/server/handler/refresh_workflow.go
```diff
@@ -79,44 +103,63 @@ func (h *RefreshWorkflowHandler) Prepare(r *http.Request) (interface{}, int, err
...
-
-	jobSpec := job.NewWorkflowSpec(
-		workflowObject.Name,
-		workflowObject.Id.String(),
-		h.Database.Config(),
-		h.Vault.Config(),
-		h.JobManager.Config(),
-		h.GithubManager.Config(),
-		args.Parameters,
+	workflowDag, err := dag_utils.NewWorkflowDag(
```
hsubbaraj-spiral, line 123: Right now we aren't using the `args.Parameters` that are passed in. We'll need to add the user-defined parameters to the new `dbWorkflowDag` that we pass in here. Something like this:
```
// Overwrite the "default" values in the operator spec for this workflowDag.
	// Because this workflowDag is never written to the database, we will not contaminate
	// the default in the db.
	if ex.Parameters != nil {
		for name, newVal := range ex.Parameters {
			op := dbWorkflowDag.GetOperatorByName(name)
			if op == nil {
				continue
			}

			if !op.Spec.IsParam() {
				return errors.Newf("Cannot set parameters on a non-parameter operator %s", name)
			}
			dbWorkflowDag.Operators[op.Id].Spec.Param().Val = newVal
		}
	}
```
from https://github.com/aqueducthq/aqueduct/blob/main/src/golang/cmd/executor/executor/workflow.go#L89 should help you do this.

hsubbaraj-spiral (changes requested):
This looks good! I left a comment about dealing with parameters, but we should be good to merge this in after the changes!

src/golang/cmd/server/handler/refresh_workflow.go
```diff
@@ -2,17 +2,29 @@ package handler
...
+	"github.com/aqueducthq/aqueduct/lib/collections/workflow_dag_edge"
+	"github.com/aqueducthq/aqueduct/lib/collections/workflow_dag_result"
 	aq_context "github.com/aqueducthq/aqueduct/lib/context"
 	"github.com/aqueducthq/aqueduct/lib/database"
 	"github.com/aqueducthq/aqueduct/lib/job"
 	"github.com/aqueducthq/aqueduct/lib/vault"
+	dag_utils "github.com/aqueducthq/aqueduct/lib/workflow/dag"
+	"github.com/aqueducthq/aqueduct/lib/workflow/engine"
 	"github.com/aqueducthq/aqueduct/lib/workflow/operator/connector/github"
+	utils "github.com/aqueducthq/aqueduct/lib/workflow/utils"
```
saurav-c, line 27: Nit: You can remove the package alias at the front since the package name is the same.

src/golang/cmd/server/handler/refresh_workflow.go
Fanjia-Yan: Have Updated this change!

hsubbaraj-spiral (approved):

PR #274, opened 2022-07-29
Re-introduce requirements inference in the SDK

kenxu95:
This should clean up our requirements story across op, metric, and checks. Now, the user can supply the following for requirements:
- A path to the requirements.txt file to use. (string)
- A list of pip requirements specifiers (eg, ["transfomers"]) (list of strings)
If neither of those are provided, we'll look for "requirements.txt" in the same directory as the decorated function.
If all of those fail, we'll use inference with `pip freeze`. 

Adds this inference at the top level, so a user can do `aqueduct.infer_requirements()` and get a list of pip requirements specifiers back.

Also adds some argument validation to the decorator.

Adds a special requirements test suite to our integration tests, which will be triggered only with the `--requirements` cmd line flag.

ENG-1217


hsubbaraj-spiral (approved):
These changes look good to me. I appreciate the detailed test cases + comments, they made understanding what each test covers very clear. 

Do we have a sense of how large/how many unnecessary packages we are introducing in the `requirements.txt` by using `pip freeze`? This may not matter in the Aqueduct Engine setting, but when we use other execution engines where we have to download packages to each function pod, this might slow execution. Maybe we can make a note of this in our documentation (push users to provide their own requirements.txt). Or maybe even disallow inferring requirements when not using Aqueduct Engine.

vsreekanti:
I think this looks good to me @kenxu95. @hsubbaraj-spiral, your concerns about package size is valid, but it's hard to say without knowing what the user has installed on their machine. It could be a few MB or 10s of GB. I think we should go ahead and merge this for now and then optimize performance once we hear complaints about this in the future.

PR #275, opened 2022-07-29
Use direct API call rather than separate process when setting up backend tests

likawind:
Quote from the task description:
Now we use an external scripts to set up backend tests. This is a bit error probing esp. that we rely on stderr to decide if an error occurred, which will have false-negative when the process runs with warnings. Also we need to build additional error handling to expose any exceptions if the process failed.

Instead, we can just set up the workflows using direct SDK API calls which is much debuggable and readable.

This change allow us to have more explicit control over what to use during workflow setup (e.g. we now return `flow.id()`, and can return additional information in the future). It also removes the error-probing behavior relying on stderr to detect failures, and parse stdout to get flow ids. Code is also a lot more cleaner after this change.

ENG-1474

- [na] If this is a new feature, I have added unit tests and integration tests.


eunice-chan (approved):
I think that looks good. I updated the integration tests to return flow.id() & number of runs. I can update my tests before merging once I start merging my changes to main.

.github/workflows/integration-tests.yml
```diff
@@ -89,6 +89,7 @@ jobs:
         working-directory: integration_tests/backend
         env:
           SERVER_ADDRESS: localhost:8080
+          INTEGRATION: aqueduct_demo
```
likawind, line 92: adding this for the next PR #276 . It seems that this need to be merged first before taking effect

PR #276, opened 2022-07-29
[1/4][Server] Eng 1037 implement a test connection button

likawind:
This PR adds the backend implementation for testing an existing integration's connection. Specifically, we slightly modified `ValidateConfig` helper in `ConnectIntegration` handler so that in the new handler, we read the config for existing integration and use that config to test-connect.

The handler returns nothing, it's sufficient for caller to rely on http status and error message.

Integration test for this endpoint is included. We add slight modification so that we can provide integration name externally as env var, and the test methods can be aware of the integration ID being used.

ENG-1037


.github/workflows/integration-tests.yml
```diff
@@ -80,6 +80,7 @@ jobs:
         working-directory: integration_tests/backend
         env:
           SERVER_ADDRESS: localhost:8080
+          INTEGRATION: aqueduct_demo
```
likawind, line 83: @kenxu95 the test failed complaining about missing env var: https://github.com/aqueducthq/aqueduct/runs/7586075455?check_suite_focus=true . Should this change be merged prior to this PR?

kenxu95: Hmm I thought we would be checking out the current branch so things would work ( actions/checkout@v2 is called at the top) @hsubbaraj-spiral is this true?

yeah if it always uses the main branch

integration_tests/backend/README.md
```diff
@@ -16,7 +16,7 @@ At the very end, the tests **require** you to print the flow id (e.g. `print(flo
 ## Usage
 
 Running all the tests in this repo:
-`API_KEY=<your api key> SERVER_ADDRESS=<your server's address> pytest . -rP`
+`API_KEY=<your api key> SERVER_ADDRESS=<your server's address> INTEGRATION=<integraiton name> pytest . -rP`
```
saurav-c, line 19: nit: typo in `<integration name>`

src/golang/cmd/server/handler/test_integration.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func (h *TestIntegrationHandler) Perform(ctx context.Context, interfaceArgs interface{}) (interface{}, int, error) {
+	args := interfaceArgs.(*TestIntegrationArgs)
+	id := args.IntegrationId
+
+	emptyResp := TestIntegrationResponse{}
+
+	integrationObject, err := h.IntegrationReader.GetIntegration(ctx, id, h.Database)
+	if err == database.ErrNoRows {
```
saurav-c, line 64: Why do we not return an error response for all cases where `err != nil` here?

saurav-c (approved):
LGTM, Left a couple of minor comments that once resolved should be good to go

src/golang/cmd/server/handler/test_integration.go
likawind: Nice catch! We should indeed return an 500 for other types of errors

integration_tests/backend/test_backend.py
```diff
@@ -7,11 +7,14 @@
...
 class TestBackend:
     GET_WORKFLOW_TABLES_TEMPLATE = "/api/workflow/%s/tables"
+    GET_TEST_INTEGRATION_TEMPLATE = "/api/integration/%s/test"
 
     @classmethod
     def setup_class(cls):
         cls.client = aqueduct.Client(pytest.api_key, pytest.server_address)
-        cls.flows = {"changing_saves": setup_changing_saves(cls.client)}
+        integration = cls.client.integration(name=pytest.integration)
+        cls.integration = integration
```
eunice-chan, line 16: I think here we can directly:
```
cls.integration = cls.client.integration(name=pytest.integration)

**Review by eunice-chan — APPROVED** [review]

File: src/golang/cmd/server/handler/test_integration.go [hunk 5]
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/aqueducthq/aqueduct/lib/database"
+	"github.com/aqueducthq/aqueduct/lib/job"
+	"github.com/aqueducthq/aqueduct/lib/vault"
+	"github.com/aqueducthq/aqueduct/lib/workflow/operator/connector/auth"
+	"github.com/dropbox/godropbox/errors"
+	"github.com/go-chi/chi/v5"
+	"github.com/google/uuid"
+)
+
+// TestIntegrationHandler tries to connect to an existing integration.
```
eunice-chan, line 20: We may want to go through all the handlers, including this, and add a comment with the expected input/output like in https://github.com/aqueducthq/aqueduct/blob/1cb2e166caf6793573e1f8206bfa285416413674/src/golang/cmd/server/handler/preview_table.go#L30-L40.

eunice-chan (commented):
Dev documentation

PR #277, opened 2022-07-29
Improve and logging in our SDK

kenxu95:
Main purpose was to address https://linear.app/aqueducthq/issue/ENG-1403/remove-warnings-about-overriding-an-existing-operator. I also took the chance to clean up our logging a little bit:
- `Logger.logger.info(...)` seemed pretty redundant. I don't see why we can't use `logger().info(..)`. We're not supposed to call functions from the root logger like `logging.info(..)`.
- Instead of adding a verbose flag, we can have the user configure the client with the correct logging level, the same way Ray does it https://docs.ray.io/en/latest/ray-core/package-ref.html#ray-remote. 

I'm also fine with us exposing a `verbose` argument and mapping that to `log.INFO`, whereas it's regularly `log.WARNING`. What do you think? I'd like to also be able to set log level to be DEBUG whenever I want, and I wasn't sure if that level should be included in verbose mode.

ENG-1217

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -68,7 +68,7 @@ jobs:
         run: echo "API_KEY=$(aqueduct apikey)" >> $GITHUB_ENV
 
       - name: Run the SDK Integration Tests
-        timeout-minutes: 10
+        timeout-minutes: 15
         working-directory: integration_tests/sdk
         env:
           SERVER_ADDRESS: localhost:8080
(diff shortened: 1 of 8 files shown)


vsreekanti (commented):
@kenxu95, so we only print warnings and above now, and the bit about overriding a function is now an "info" level log. Is that right?

kenxu95:
@vsreekanti yep thats correct! It sounded from the task description that we didn't want this to be a warning anymore.

vsreekanti:
Yep, that was my intention! Just wanted to make sure I understood correctly. This set up is good with me!

vsreekanti (approved):

PR #280, opened 2022-08-01
[ENG-1362, 1450] Fixes bug where user got unhelpful error message when Aqueduct server was off

vsreekanti:
When the user tries to connect an integration but the Aqueduct server is off, they get a very confusing error message that simply says "Failed to fetch". This PR fixes that by explicitly checking when a fetch failed to happen. It also improves the integration connection modal by making it wider, so that stack traces display more legibly.

image

This PR also fixes a tiny bug that I found where the list of integrations on the UI isn't refreshed after you connect a new integration. Turns out that was already logged as ENG-1450, but I only discovered that afterwards. 🙂

ENG-1362, ENG-1450

### DIFF
--- a/src/ui/common/src/utils/integrations.ts
+++ b/src/ui/common/src/utils/integrations.ts
@@ -236,10 +236,14 @@ export async function connectIntegration(
       throw new Error(message.error);
     }
   } catch (err) {
-    if (err instanceof TypeError) { // This happens when we fail to fetch.
-      throw new Error("Unable to connect to the Aqueduct server. Please double check that the Aqueduct server is running and accessible.")
-    } else { // This should never happen.
-      throw err
+    if (err instanceof TypeError) {
+      // This happens when we fail to fetch.
+      throw new Error(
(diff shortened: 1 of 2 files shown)


src/ui/common/src/utils/integrations.ts
```diff
@@ -219,19 +219,32 @@ export async function connectIntegration(
...
+      method: 'POST',
+      headers: {
+        'api-key': user.apiKey,
+        'integration-name': name,
+        'integration-service': service,
+        'integration-config': JSON.stringify(config),
+      },
+    });
+
+    if (!res.ok) {
```
agiron123, line 234: Just making sure I understand the API's response correctly here, when an integration connects successfully do we just return a 200 OK and no response body?

agiron123 (approved):

src/ui/common/src/utils/integrations.ts
vsreekanti: Yep!

PR #284, opened 2022-08-01
Eng 1422 Integrations List Page Improvements

agiron123:
This PR adds some styling improvements to the grid on the Integrations List Page.

Loom demo can be found here: https://www.loom.com/share/5b632beb968a4ebbb9de721023e88218

https://linear.app/aqueducthq/issue/ENG-1422/integrations-list-page-improvements

### DIFF
--- a/src/ui/app/package.json
+++ b/src/ui/app/package.json
@@ -3,7 +3,7 @@
   "author": "Aqueduct, Inc. <hello@aqueducthq.com>",
   "version": "0.0.7",
   "scripts": {
-    "start": "parcel --no-cache --no-hmr index.html",
+    "start": "parcel --no-cache index.html",
     "build": "parcel build --public-url /dist --dist-dir dist/default index.html",
     "build-sagemaker": "parcel build --public-url /proxy/8080/dist --dist-dir dist/sagemaker index.html",
     "lint": "eslint '*/**/*.{js,ts,tsx}' --format table",
--- a/src/ui/common/package-lock.json
+++ b/src/ui/common/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@aqueducthq/common",
-  "version": "0.0.12",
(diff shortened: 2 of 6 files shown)


vsreekanti:
This looks great, @agiron123! A few questions/thoughts:

* Should we enable a global feature flag on the UI for breadcrumbs? It might be confusing to have breadcrumbs incrementally on some pages and not on others, so we can just turn them on once they're all implemented. 
    * Relatedly, "All" seems confusing -- shouldn't it just be `Home / Integrations`? 
* Is there a reason you went for a max width of 3 panes? Seems like we could do a bit more on wider screens like yours, but that's not a strong opinion.
*

agiron123:
> This looks great, @agiron123! A few questions/thoughts:
> 
> * Should we enable a global feature flag on the UI for breadcrumbs? It might be confusing to have breadcrumbs incrementally on some pages and not on others, so we can just turn them on once they're all implemented.
> 

@vsreekanti I think we should just comment out the breadcrumbs out for now and then add them all at once. Made a task to track this work here: https://linear.app/aqueducthq/issue/ENG-1488/add-breadcrumbs-to-the-app
  
>   * Relatedly, "All" seems confusing -- shouldn't it just be `Home / Integrations`?

Sure, I can update to just show Home / Integrations

> * Is there a reason you went for a max width of 3 panes? Seems like we could do a bit more on wider screens like yours, but that's not a strong opinion.

I went for three instead of four as there are nine integrations at the moment.
There's a lot of room, yes. But I think this is something that we can address in all layouts in another task. Ideally, I'd like to have all the layouts be centered and have a 1200px max-width. I've seen other websites do more relative units like 75rem, but we run on a scale of units of 8, so 1200 would be 150 units in this case.

vsreekanti (approved):
LGTM! :shipit:

PR #290, opened 2022-08-03
Eng 1370 UI allow users to delete workflow saved objects

eunice-chan:
Allow users to delete workflow and associated saved objects in the UI.

## Demo
Demo of info popup:
On hover, gives details about why there are multiple update modes.

https://user-images.githubusercontent.com/30596854/182651220-98aae124-b925-4ed2-8d79-5395bec7b3a8.mov

Demo of workflow deletion flow:
1. Shows if table to delete exists and table selected for deletion, successfully deletes & shows green checkmark next to table in results page.
2. Shows if table to delete doesn't exist and table selected for deletion, fails to delete & shows red cross next to table in results page.
3. Creates the table again and shows table can still be deleted if it exists.

https://user-images.githubusercontent.com/30596854/182651184-7acedd04-d9a7-43fb-9233-58714cc93e37.mov

Eng 1370


src/ui/app/package.json
```diff
@@ -10,11 +10,15 @@
     "lint:fix": "eslint '*/**/*.{js,ts,tsx}' --format table --fix"
   },
   "dependencies": {
+<<<<<<< HEAD
```
likawind, line 13: note to clean up these

src/ui/common/src/components/workflows/WorkflowSettings.tsx
```diff
@@ -278,25 +335,45 @@ const WorkflowSettings: React.FC<WorkflowSettingsProps> = ({
 
   const handleDeleteWorkflow = (event) => {
     event.preventDefault();
+
     setIsDeleting(true);
 
+    const data = { force: true };
```
likawind, line 341: what is this 'data' for?

src/ui/common/src/reducers/workflow.ts
```diff
@@ -46,6 +48,7 @@ export type OperatorResult = {
 
 export type WorkflowState = {
   loadingStatus: LoadingStatus;
+  savedObjects: { [id: string]: SavedObject[] };
```
likawind, line 51: since the request-response is handled separately than workflow load, let's add a separate `loadingStatus` to handle the status of these objects

likawind (changes requested):
Looks great! Really like the dialog box! A few suggestions other than that:
- it would be really nice to have a video / screenshots for the full feature. I'm mainly curious what things looks like / move around after the deletion is processing / succeeded / failed
- I feel we could start being more disciplined on using redux store to handle request-response in general. Specifically for this PR:
  - we could add the `loadingStatus` for saved object loading
  - we should use redux to handle deletion state as well
 @agiron123 I'm also curious about how you think

src/ui/app/package.json
eunice-chan: Oops I didn't have my merge conflict fixes. Thanks for pointing them out!

src/ui/common/src/components/workflows/WorkflowSettings.tsx
eunice-chan: `data` is what is sent to the request. There are two items:
1. `external_delete`: the saved objects to remove
2. `force`: whether to force deletion even if some saved objects have append update modes.

`force` is toggled in the SDK with the force argument to the function but here, we present the integrations and update modes to the user and they enter the workflow name to show that they understand and agree the whole thing is deleted so `force=True` when deleting from the UI.

eunice-chan:
@likawind Thanks for your comments!

> Once the workflow is deleted, I feel we should inform the user both the success of workflow deletion and saved objects deletion. The content could also be more tightly aligned.

Can you explain what you mean by tightly aligned?


> It would be great to have a screenshot / video for failure case

0:00 - 0:22 Is an example of the success case
0:30 - 0:56 is an example of the failure case

https://user-images.githubusercontent.com/30596854/182651184-7acedd04-d9a7-43fb-9233-58714cc93e37.mov


> I feel we could start being more disciplined on using redux store to handle request-response in general. Specifically for this PR:
> we could add the loadingStatus for saved object loading
> we should use redux to handle deletion state as well

I'll look into that!

src/ui/common/src/components/workflows/WorkflowSettings.tsx
likawind: Sounds good!

likawind:
> Can you explain what you mean by tightly aligned?
I feel the icon (checkmark and cross) can be closer together?

It's great we are informing user the failures. I feel we should inform something like 'workflow successfully delete, but we failed to delete some objects', probably as a part of the modal body. Then we could show  the list of objects that are failed to delete.

eunice-chan:
Updated look of Saved Object Results page:
image

Updated UI to surface errors in endpoint call:

https://user-images.githubusercontent.com/30596854/182973022-e7b5ec0f-f6af-4e8b-92b7-7f3ade6886b3.mov

vsreekanti:
@eunice-chan, this looks awesome! There's a few things I think we can improve that will make the experience a little smoother:
* The text fields on the settings modal seems to be running over the edge of the modal. It should be a pretty easy fix to get them to stay within the bounds of the modal.
* On the delete dialog, can you remove the line break before "Please select the saved objects..." 
* Can you show what it looks like if we have multiple saved objects? I think it might make more sense to show the object name first and the integration second. 
* How come update mode shows "replace, append"? Can't the table only have one of those modes? 
* The last sentence is a little confusingly worded -- "The workflow and all 1 selected object(s) regardless of update mode will be completely removed." If I select an object on the checklist above, why do we need this warning? I'm explicitly choosing to delete the workflow, no? 
    * I'm guessing this has to do with our previous conversation about append mode deletion. I would maybe say that explicitly  -- **if you're deleting an append mode table, we can't guarantee we're only deleting Aqueduct data** or something like that. 
* The error shown at the end of the video above is confusing. What does it mean the organization doesn't own the workflow? The user has no notion of an organization, and we only have one organization. How does that error ever arise? 
* The title of the final modal is a little confusing. I would make the title `<workflow_name> successfully deleted.` and then include the information you have there. 
    * Ideally, we'd also show the error associated with the deletion. Is that possible? 
* I think the above will fix this, but it should be very clear to the user that *if saved object deletion fails, the workflow is still deleted*. That's not clear in the current UI, so maybe we can also add this to the first dialog when they select objects to delete? 
* The word delete being in red in the text is a little confusing. I'd prefer we bold it rather than making it red.

Let me know if you have any questions!

eunice-chan:
@vsreekanti Thanks for your feedback!


> The text fields on the settings modal seems to be running over the edge of the modal. It should be a pretty easy fix to get them to stay within the bounds of the modal.

Updated:
image


> On the delete dialog, can you remove the line break before "Please select the saved objects..."
> Can you show what it looks like if we have multiple saved objects? I think it might make more sense to show the object name first and the integration second.

Updated + show with multiple saved objects:
image


> How come update mode shows "replace, append"? Can't the table only have one of those modes?

We show multiple update modes if the update mode associated with the integration - table combination differs among save operators in different runs (and also if one saves to the same integration - table with different update modes in the same workflow run).

We explain it as a tooltip. Open to suggestions on making it clearer: 
image


> The last sentence is a little confusingly worded -- "The workflow and all 1 selected object(s) regardless of update mode will be completely removed." If I select an object on the checklist above, why do we need this warning? I'm explicitly choosing to delete the workflow, no?
> I'm guessing this has to do with our previous conversation about append mode deletion. I would maybe say that explicitly -- if you're deleting an append mode table, we can't guarantee we're only deleting Aqueduct data or something like that.

Yes, the last sentence was to make sure the user is aware of the implications of deleting objects.

**[Followup]**
Would the following be more clear: Are you sure you want to **delete** [workflow_name]? This action is not reversible. The workflow and all **[selected_count]** selected object(s) will be **completely removed**. Please note we cannot guarantee we are only deleting Aqueduct-written data when removing external objects.


> The error shown at the end of the video above is confusing. What does it mean the organization doesn't own the workflow? The user has no notion of an organization, and we only have one organization. How does that error ever arise?

The error was to show generally how errors would look if an error is encountered while the endpoint was running that isn't related to deleting saved objects.

This specific error results from when the user does not own or belong to the organization that owns integration. I can make it show up here by deleting the integration and attempting to delete it again. It's not possible in the current UI without fiddling around in the backend because after the workflow is deleted, you can only press Close which redirects to the workflows page.


> The title of the final modal is a little confusing. I would make the title <workflow_name> successfully deleted. and then include the information you have there.

Updated:
image


> Ideally, we'd also show the error associated with the deletion. Is that possible?

Currently we just return true/false. I've added this to the Nice-To-Have and work on the design + implementation of this once the other features are done.


> I think the above will fix this, but it should be very clear to the user that if saved object deletion fails, the workflow is still deleted. That's not clear in the current UI, so maybe we can also add this to the first dialog when they select objects to delete?

**[Followup]**
Will update the first line for clarity. Let me know your thoughts.

Current:
The following objects had been saved by [workflow_name] and can be removed when deleting the workflow. Please select the saved objects you wish to delete

Proposed:
The following objects had been saved by [workflow_name] and can be removed when deleting the workflow. **This is done best-effort.** The workflow will be deleted even if the object deletions failed. Please select the saved objects you wish to delete


> The word delete being in red in the text is a little confusing. I'd prefer we bold it rather than making it red.

Updated:
image

vsreekanti:
* re: fixing the width of the text fields: looks great! 
* re: multiple saved objects: also looks great! Do we want to consider switching the table name (putting it first and in bold) and then putting the integration name second? Right now, it sort of looks to me like we're selecting the option to delete `aqueduct_demo` twice. 
* re: multiple update modes, got it. Maybe let's do something like this: By default, just show the current update mode, and if there was a historically different update mode, you can tack on something like "(previously used <different mode>)". 
    * I think it's going to be a pretty uncommon use case that the user saves the same integration-table combination multiple times within a workflow. In fact, this almost sounds like a bug that we might want to disallow in the future to me.
* re: the final text on the delete modal, the new message sounds great to me! 
* re: making the error message a nice-to-have, that sounds great. 
* re: best effort deletion, I would tweak the wording to be something like this: "**Aqueduct cannot guarantee that the underlying data will be successfully deleted. The workflow will be deleted regardless.**"

Other than that, this looks great -- thanks @eunice-chan!

eunice-chan:
@vsreekanti Thanks for your feedback. Would like your thoughts on the following:

> re: multiple saved objects: also looks great! Do we want to consider switching the table name (putting it first and in bold) and then putting the integration name second? Right now, it sort of looks to me like we're selecting the option to delete aqueduct_demo twice.

I see. My reasoning was GROUP_NAME: ITEM_IN_GROUP made sense to me intuitively. But I can see how that could be confusing esp since bold makes it seem like the integration is the part the user should pay attention to. Mocked up some alternatives. Let me know what you think:
image

> re: multiple update modes, got it. Maybe let's do something like this: By default, just show the current update mode, and if there was a historically different update mode, you can tack on something like "(previously used )".

That's a good point. Currently the backend just returns a list of update modes. I added it to the nice-to-have and will revisit it in a future task.

> I think it's going to be a pretty uncommon use case that the user saves the same integration-table combination multiple times within a workflow. In fact, this almost sounds like a bug that we might want to disallow in the future to me.

I could see a use case where someone might REPLACE a table with an empty table and APPEND to that table later in the workflow. APPENDing to a table multiple times also seems reasonable although I can't think of a specific use case right now. It seems a bit hacky, but I don't think we would want to disallow it yet.

> re: the final text on the delete modal, the new message sounds great to me!
> re: best effort deletion, I would tweak the wording to be something like this: "Aqueduct cannot guarantee that the underlying data will be successfully deleted. The workflow will be deleted regardless."

Here is how it looks put together! It seems a bit wordy. 
image

vsreekanti:
* re: integration name vs. item name, I like your first option. Showing the integration first makes sense, and putting it in square brackets makes it clear it's a category, and bolding the table name is clear to me.
* re: multiple update modes, makes sense. Let's keep it in for now, and we can return to it later
* You're right that it's wordy. Here's a slightly shortened version: 
```The following objects have been saved by <workflow name> and can be deleted along with the workflow: 
<list of objects>
Deleting workflow <workflow name> and the associated <number of selected objs> objects is not reversible. Please note that we cannot guarantee this will only delete data created by Aqueduct. The workflow will be deleted even if the underlying objects are not successfully deleted.
```

eunice-chan:
Additional screenshot of what the delete page looks like when there is no saved objects associated with the workflow:

image

src/ui/common/src/reducers/workflow.ts
```diff
@@ -301,6 +403,39 @@ export const workflowSlice = createSlice({
       };
     });
```
likawind, line 405: Let's add the case for `pending` even if we are not explicitly render this

vsreekanti:
The screenshots look great to me @eunice-chan! No more changes from my end. Great work! :shipit:

src/ui/common/src/reducers/workflow.ts
```diff
@@ -326,6 +466,23 @@ export const workflowSlice = createSlice({
         err: payload as string,
       };
     });
+
```
likawind, line 469: missed this in previous review, let's add the pending case here as well?

likawind (approved):
Great work! It took some iteration to get the UI sorted but definitely appreciate the efforts! Besides one small redux nit I think the feature is ready to go!

PR #291, opened 2022-08-03
[2/4] Add button group to integration details

likawind:
## Describe your changes and why you are making these changes
This PR adds a button group to integration details page to organize all integration-specific actions. Each action is hoverable with tooltips. This change makes the UI cleaner with incoming edit / delete integration changes. Full demo will be available in the final PR.

## Related issue number (if any)
ENG-1037

## Checklist before requesting a review
Demo in #293 and integration test in #276 
- [x] I have created a descriptive PR title. The PR title should complete the sentence "This PR...".
- [x] I have performed a self-review of my code.
- [x] I have included a small demo of the changes. For the UI, this would be a screenshot or a Loom video.
- [x] If this is a new feature, I have added unit tests and integration tests.
- [x] I have run the integration tests locally and they are passing.
- [x] I have run the linter script locally (See `python3 scripts/run_linters.py -h` for usage).
- [x] All features on the UI continue to work correctly.
- [x] Added one of the following CI labels:
    - `run_integration_test`: Runs integration tests
    - `skip_integration_test`: Skips integration tests (Should be used when changes are ONLY documentation/UI)

PR #293, opened 2022-08-03
[4/4] Enables integration's test connection button

likawind:
This PR enables test-connect integration button to allow user to test-connect an existing integration. The button rendering is introduced in #291. In this PR we mainly enable the button by setting up all request-response to endpoint introduced in #276 . We also added an additional dialog box to render connection information.

ENG-1037.

Demo: https://www.loom.com/share/78b6190260924173ba13c1287ee42bcc
Demo of failure:
Screen Shot 2022-08-04 at 12 18 38 AM

- [x] If this is a new feature, I have added unit tests and integration tests. Tests are introduced in #276 


src/ui/common/src/components/integrations/dialogs/testConnectDialog.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  return (
+    <Dialog open={true} onClose={onCloseDialog} fullWidth maxWidth="lg">
+      <DialogTitle>
+        <Typography variant="h5">{'Testing Connection'}</Typography>
+      </DialogTitle>
+      <DialogContent>{dialogBody}</DialogContent>
+      <DialogActions>
+        {isLoading(status) && (
+          <Button color="secondary" autoFocus onClick={onCloseDialog}>
```
eunice-chan, line 53: If the status is loading, this is the only button being displayed so I think the color should be `primary`.

src/ui/common/src/components/integrations/dialogs/testConnectDialog.tsx
```diff
@@ -0,0 +1,81 @@
...
+            Cancel
+          </Button>
+        )}
+        {isFailed(status) && (
+          <Button
+            autoFocus
```
eunice-chan, line 59: I think we can remove `autoFocus` because the confirm button also has `autofocus`.

eunice-chan (approved):
Looks good! Thanks for the demo on Loom.

vsreekanti:
Thanks for recording the demos, @likawind! A few things on the UI front. @agiron123, I'd love your thoughts here as well:
* I wonder if the beaker and CSV icons are a little cryptic for people to follow. What do you think about having some sort of settings dropdown that has options for testing the connection, editing, deleting, and "Add CSV" can be its own button? 
* I'm not sure that a modal is the best interaction mechanism for testing the connection. I wonder if it might be better to just show a Snackbar when the test starts and one when it succeeds/fails?
* If we keep the modal, "Confirm" is also a slightly confusing word to use on this because I'm not sure what exactly I'm confirming. 
* I think the success message should be slightly more specific: "Connection to <integration name> was successful!" 
* In the error case, it's not clear what exactly `fail` means from the user perspective. It would be nice to show a more detailed error message here. 
* The `File an Issue` button is nice, but it's not clear what the user would say when they filed an issue. Also, unless it's a 500/internal error, the user should just be given the context to debug themselves. 

My proposal is to maybe do the following: 
* Show a toast for testing started. 
* Show a toast for successful testing. 
* If the last connection test was successful, show a green check mark next to the integration name. 
* If the connection test failed, show a stack trace on below the header and above the preview dropdown. 

@agiron123, thoughts?

agiron123:
@likawind @vsreekanti 

First off, nice work @likawind !

I have a couple suggestions on the UI and I've created a Figma mock here:
https://www.figma.com/file/zK2bQgVjHgmUrcwb6qnhM9/UI-V2.1-Screens?node-id=2028%3A30706

> I wonder if the beaker and CSV icons are a little cryptic for people to follow. What do you think about having some sort of settings dropdown that has options for testing the connection, editing, deleting, and "Add CSV" can be its own button?
I'm not sure that a modal is the best interaction mechanism for testing the connection. I wonder if it might be better to just show a Snackbar when the test starts and one when it succeeds/fails?

I agree here about the icons. I think we can just put them as options behind an options dropdown.

> If we keep the modal, "Confirm" is also a slightly confusing word to use on this because I'm not sure what exactly I'm confirming.

Yes, totally agree here. Instead of a modal for success, we should show a success toast that disappears after a few seconds. Happy to help find you an example of this in the codebase if you need. As mentioned in Figma mocks, after connection is successful, update the timestamp for the "Connected on" text.

> I think the success message should be slightly more specific: "Connection to was successful!"
In the error case, it's not clear what exactly fail means from the user perspective. It would be nice to show a more detailed error message here.

I think the connection failed modal that is being shown is fine. The button at the bottom of the connection failed modal should say "Dismiss"

The File an Issue button is nice, but it's not clear what the user would say when they filed an issue. Also, unless it's a 500/internal error, the user should just be given the context to debug themselves.

>My proposal is to maybe do the following:
>Show a toast for testing started.
>Show a toast for successful testing.
>If the last connection test was successful, show a green check mark next to the integration name.
>If the connection test failed, show a stack trace on below the header and above the preview dropdown.

All these are fine with me. If we want to do less work here, I think we can just keep the error modal and change the button at the bottom to say "Dismiss" and remove file an issue button. For extra credit, we can add a button to copy the contents of the error message / stack trace to clipboard.

likawind:
@vsreekanti @agiron123 I kinda combined both of your comments about the button group, lmk how that looks:
Screen Shot 2022-08-03 at 3 38 47 PM

agiron123:
> @vsreekanti @agiron123 I kinda combined both of your comments about the button group, lmk how that looks: Screen Shot 2022-08-03 at 3 38 47 PM

Sounds good to me!

vsreekanti:
Yep, this looks great @likawind!

likawind:
@vsreekanti @agiron123 updated! One thing to clarify is when testing the failure, we hack the error message to be a short one. In practice we will render whatever message from the backend which should contain all contexts.

I changed 'connected on' to 'created on' since 'connected' could confuse user with test-connect concept. Another thing is I didn't add the check for success case since the I couldn't find a good place to fit it. As we are already using toasts I feel it's fine to keep the success information within a limited lifecycle.

vsreekanti:
@likawind, this looks great! A few tiny nits:
* Can we put the Upload CSV button to the left of the "Options" dropdown? 
* In the toasts, can we change the text to be: "Attempting to connect to <integration name>..." and "Successfully connected to <integration name>!" 
* If there's a full stack trace where `failed_hack` currently it, it might be difficult to read. Could we put the stack trace in a `<pre>` and on a separate line? That will significantly improve readability. If you have a moment and can share what this might look like with a full stack trace, that would be awesome.

src/ui/common/src/components/integrations/options.tsx
```diff
@@ -0,0 +1,62 @@
...
+          onClick={() => {
+            setAnchorEl(null);
+            onTestConnection();
+          }}
+        >
+          <FontAwesomeIcon color="gray.800" icon={faFlask} />
+          <Typography color="gray.800" variant="body2" sx={{ marginLeft: 1 }}>
+            Test Connection
+          </Typography>
+        </MenuItem>
```
likawind, line 56: @eunice-chan we can add delete button here as another `<MenuItem/>`

likawind:
@vsreekanti updated:
Screen Shot 2022-08-04 at 4 18 01 PM

likawind:
Since @eunice-chan has been blocked on this, @vsreekanti @agiron123 I will merge with the current state. If there's additional suggestions I'm happy address them in separate future PRs

vsreekanti:
I think this is good to go, @likawind!

PR #302, opened 2022-08-05
Use virtual table to improve render efficiency

cw75:
This PR replaces our `DataTable` component with the React virtualized table that only renders the rows visible on the browser, which significantly improves the rendering performance. See the demo below:

https://www.loom.com/share/f0eab6f015994c36be39cc13e074fd3f

@agiron123 we may need to add the `--force` flag for the UI build test to pass, due to the peer dependency issue.


scripts/install_local.py
```diff
@@ -118,14 +118,14 @@ def execute_command(args, cwd=None):
         UI_APP_PATH = UI_PATH + "/app"
         
         print("Updating UI files...")
+        execute_command(["rm", "-rf", "node_modules"], cwd=join(cwd, UI_COMMON_PATH))
```
vsreekanti, line 121: Hm, is this something we want to do every time? That means that running `install_local` is going to be a pretty expensive operation.

scripts/install_local.py
```diff
@@ -118,14 +118,14 @@ def execute_command(args, cwd=None):
...
         execute_command(["rm", "-rf", ".parcel-cache"], cwd=join(cwd, UI_COMMON_PATH))
         execute_command(["rm", "-rf", "dist"], cwd=join(cwd, UI_COMMON_PATH))
         execute_command(["npm", "install"], cwd=join(cwd, UI_COMMON_PATH))
-        execute_command(["npm", "run", "build"], cwd=join(cwd, UI_COMMON_PATH))
```
vsreekanti, line 124: I'm also not sure if this script will work if we're not actually building the `common` lib.

src/ui/common/src/components/tables/virtualizedTable.tsx
```diff
@@ -0,0 +1,252 @@
...
+import * as React from 'react';
+import {
+  AutoSizer,
+  Column,
+  Table,
+  TableCellRenderer,
+  TableHeaderProps,
+} from 'react-virtualized';
+
+const classes = {
```
vsreekanti, line 14: Are these not exposed as constants from `react-virtualized`? I know MUI does this. It'd make it more maintainable for the future (i.e., if these class names change).

src/ui/common/src/components/tables/virtualizedTable.tsx
```diff
@@ -0,0 +1,252 @@
+import { Box, Typography } from '@mui/material';
+import { styled, Theme } from '@mui/material/styles';
+import TableCell from '@mui/material/TableCell';
+import clsx from 'clsx';
+import * as React from 'react';
```
vsreekanti, line 5: Nit: We've been doing `import React from "react";` by default. Slightly cleaner to read.

src/ui/common/src/components/tables/virtualizedTable.tsx
```diff
@@ -0,0 +1,252 @@
...
+  noClick: 'ReactVirtualizedDemo-noClick',
+};
+
+const columnWidthMultiplier = 13;
+
+const styles = ({ theme }: { theme: Theme }) =>
+  ({
+    // temporary right-to-left patch, waiting for
+    // https://github.com/bvaughn/react-virtualized/issues/454
+    '& .ReactVirtualized__Table__headerRow': {
```
vsreekanti, line 28: Should we at least make these class names `const`s for the time being as well?

src/ui/common/src/components/tables/virtualizedTable.tsx
```diff
@@ -0,0 +1,252 @@
...
+  columns: readonly ColumnData[];
+  headerHeight?: number;
+  minColumnWidth?: number;
+  onRowClick?: () => void;
+  rowCount: number;
+  rowGetter: (row: Row) => any;
+  rowHeight?: number;
+}
+
+class MuiVirtualizedTable extends React.PureComponent<MuiVirtualizedTableProps> {
```
vsreekanti, line 99: Is there a reason we're doing a class component here as opposed to a functional component? Also, what is a `PureComponent`? @agiron123, I know you'd like us to use FCs everywhere, so feel free to chime in.

vsreekanti, line 99: Style nit: The fact that the table is virtualized seems like an implementation detail, so I think we should still call this a `DataTable`. 

Also, I think the general style we've been trying to follow is that both the class names and file names should be PascalCased. So here, I think we should call both `DataTable`.

src/ui/common/src/components/integrations/integrationObjectPreview.tsx
```diff
@@ -3,40 +3,62 @@ import Box from '@mui/material/Box';
...
+    content = (
+      <Box sx={{ display: 'flex', flexDirection: 'row', mt: 3 }}>
+        <CircularProgress size={30} />
+        <Typography sx={{ ml: 2 }}>
+          Loading object <b>{objectName}</b>...
+        </Typography>
+      </Box>
+    );
+  }
+  if (isFailed(object.status)) {
```
vsreekanti, line 27: Style nit: These blocks could be spaced a little better.

src/ui/common/src/components/integrations/integrationObjectPreview.tsx
```diff
@@ -3,40 +3,62 @@ import Box from '@mui/material/Box';
...
+    content = (
+      <Alert style={{ marginTop: '10px' }} severity="error">
+        Object <b>{objectName}</b> failed to load. Try refreshing the page.{' '}
```
vsreekanti, line 30: Do we actually expect that refreshing the page will help anything here? If not, we shouldn't say that. 🙂

src/ui/common/src/components/integrations/integrationObjectPreview.tsx
```diff
@@ -3,40 +3,62 @@ import Box from '@mui/material/Box';
...
+        <br />
+        Error: {object.status.err}
```
vsreekanti, line 32: Let's get in the habit of putting our error messages in `<pre>` blocks so that they format in a readable fashion.

src/ui/common/src/components/integrations/integrationObjectPreview.tsx
```diff
@@ -3,40 +3,62 @@ import Box from '@mui/material/Box';
...
+          Loading object <b>{objectName}</b>...
+        </Typography>
+      </Box>
+    );
+  }
+  if (isFailed(object.status)) {
+    content = (
+      <Alert style={{ marginTop: '10px' }} severity="error">
+        Object <b>{objectName}</b> failed to load. Try refreshing the page.{' '}
+        <br />
```
vsreekanti, line 31: Rather than adding a `<br />` here, lets instead put the above title in an `<AlertTitle>` component and then have the error message be the body of the error.

vsreekanti (commented):
This is *awesome*, @cw75! The demo looks really good. I know this wasn't easy, so thanks for soldiering through the vagaries of third-party React libraries. I've left a few style nits in line below. 

The only thing I'm not sure about is whether we want to (or even can?) make the new table component a functional component. @agiron123 might know better than me here, so I'll let him comment before we move forward. Other than that, I think this is 💯.

scripts/install_local.py
cw75: @agiron123

scripts/install_local.py
cw75: @agiron123

src/ui/common/src/components/integrations/integrationObjectPreview.tsx
cw75: @likawind

src/ui/common/src/components/tables/virtualizedTable.tsx
cw75: I don't think so after poking around a bit. The example code on the web also have these class name strings hard-coded..

src/ui/common/src/components/tables/virtualizedTable.tsx
cw75: At the end of this file we have: `const DataTable = styled(<ClassName>)(styles);`. To prevent name collision I'm naming the class `PreStyledDataTable`. Let me know if you have better ideas. @agiron123 @vsreekanti

cw75: I believe this is from a code template we copied over. @agiron123 I'll let you figure out if we should do sth else here. :)

cw75:
Addressed comments. @likawind I think some of @vsreekanti 's comments are on your changes, so would be nice if you can take a quick look as well.

src/ui/common/src/components/tables/virtualizedTable.tsx
vsreekanti: That's fine with me!

vsreekanti:
Awesome! I'll wait for @agiron123 to weigh in on the functional component thing, and we should be good to go.

scripts/install_local.py
agiron123: @vsreekanti @cw75 ```npm install``` when run inside common actually builds common at the end. This is a redundant command that will save us a couple minutes on build time.

scripts/install_local.py
agiron123: Technically correct, it will take longer. This ensures that people have a clean build and that things should work properly.

We can also add a flag to allow people to choose whether or not they want to do a clean build. Personally I'm of the camp of doing clean builds every time as there's less moving parts and things to worry about.

In the future I'd like to look into some other build tools to use. I'm pretty sure we can get much faster build times by adopting a different build tool for common.

src/ui/common/src/components/tables/virtualizedTable.tsx
agiron123: styles is a const, so I think it's fine to leave them as is.

My suggestion here would be to move the styles into a file of their own.
so we'd have virtualizedTable.tsx and virtualizedTable.styles.tsx.

Then the styles can be exported and shared amongst other components. I'd also be sure to export the styles from common (src/index.ts) as well. If we haven't done so already, the VirtualTable component should be in there as well.

src/ui/common/src/components/tables/virtualizedTable.tsx
agiron123: @vsreekanti as @cw75 mentioned, this was part of the example code given by MUI. All of the other examples in MUI use functional components. Pure components allow us to have more control over when to re-render a particular component when it's parents update.

See this blog post here for a quick rundown: 
https://dev.to/ashwin_mothilal/difference-between-pure-component-and-functional-component-51mi

If we wish to use a functional component, we can probably do so using the useMemo() hook. I don't really have an opinion either way, we can just leave as is now since things are working quite well. I'd rather use what the example did on MUI, since they probably know more about how MUI and react-virtualized work together than we do :).

agiron123: @vsreekanti @cw75 I'd rather leave this one named VirtualizedTable and have DataTable be the parent class that all other tables inherit from. (more of an OOP analogy here, but you get the point)

WIth that being said, I have a table view that I am working on right now that will have pagination. 

After that goes in, we can combine all three into one table component and control which table component is used based on the props that we pass in.

let's leave as is for now and then we can revisit and build out a nicer DataTable implementation afterwards.

src/ui/app/package.json
```diff
@@ -10,7 +10,7 @@
     "lint:fix": "eslint '*/**/*.{js,ts,tsx}' --format table --fix"
   },
   "dependencies": {
-    "@aqueducthq/common": "^0.0.12",
+    "@aqueducthq/common": "0.0.13-rc0",
```
agiron123, line 13: please npm publish after this PR gets merged in.

src/ui/common/src/components/integrations/integrationObjectPreview.tsx
agiron123: As a general rule, we really shouldn't be using <br /> and should be using css margins/paddings instead.

src/ui/common/src/components/tables/virtualizedTable.tsx
agiron123: I think this just came from the example code that we brought over, but should be an easy fix to stay consistent.

src/ui/common/src/components/tables/virtualizedTable.tsx
```diff
@@ -0,0 +1,252 @@
...
+    );
+  };
+
+  render() {
+    const { columns, rowHeight, headerHeight, minColumnWidth, ...tableProps } =
+      this.props;
+
+    let MIN_TABLE_WIDTH = 0;
+    columns.forEach((column) => {
+      if (column.columnWidth == null) {
```
agiron123: ===

agiron123: However, JS is truthy, so we can do this instead:
```
if (!column.columnWidth)
```

agiron123 (approved):
Good to merge once comments are resolved.

src/ui/common/src/components/integrations/integrationObjectPreview.tsx
```diff
@@ -1,42 +1,66 @@
...
+          Loading object <b>{objectName}</b>...
+        </Typography>
+      </Box>
+    );
+  }
+
+  if (isFailed(object.status)) {
+    content = (
+      <Alert style={{ marginTop: '10px' }} severity="error">
+        <AlertTitle>Object <b>{objectName}</b> failed to load. Try refreshing the page.</AlertTitle>
```
likawind, line 31: Yeah we should remove the page refreshing thing

PR #304, opened 2022-08-05
Add `verbose` mode so that by default only error logs are shown in the terminal

cw75:
This PR adds `verbose` option to our CLI. By default, `verbose` is set to `False`, meaning that only error logs are shown in the terminal. If the user sets `verbose` to `True`, then all logs are shown in the terminal.

Regardless of `verbose`, all logs are always stored into a log file. The log file is managed by a library that performs log rotation to prevent the logs from growing infinitely.

The above only applies to logging done via `logrus`. Our route logger doesn't have this flexibility, so all REST events are still logged in the terminal regardless of `verbose`.

Without `verbose`:
Screen Shot 2022-08-05 at 9 18 07 AM

With `verbose`:
Screen Shot 2022-08-05 at 9 18 31 AM

### DIFF
--- a/src/golang/cmd/server/handler/archive_notification.go
+++ b/src/golang/cmd/server/handler/archive_notification.go
@@ -17,8 +17,10 @@ import (
 // Method: POST
 // Params: notificationId
 // Request:
+//
 //	Headers:
 //		`api-key`: user's API Key
+//
 // Response: none
 type ArchiveNotificationHandler struct {
 	PostHandler
--- a/src/golang/cmd/server/main.go
+++ b/src/golang/cmd/server/main.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"flag"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 
@@ -40,25 +39,29 @@ func main() {
 		ForceColors:  true,
 	})
 
-	if !*verbose {
-		log.SetOutput(ioutil.Discard) // Send all logs to nowhere by default
+	// Always store all logs to a log file.
+	// With lumberjack.Logger we can do log rotation to prevent it from growing infinitely.
+	log.SetOutput(&lumberjack.Logger{
+		Filename:   serverLogPath,
+		MaxSize:    100, // megabytes
+		MaxBackups: 3,
+		MaxAge:     28, // days
+	})
 
-		log.AddHook(&writer.Hook{ // Send logs with level higher than error to stderr
-			Writer: os.Stderr,
-			LogLevels: []log.Level{
-				log.PanicLevel,
-				log.FatalLevel,
-				log.ErrorLevel,
-			},
-		})
-		log.AddHook(&writer.Hook{ // Send info, debug, warning logs to the log file
-			// With lumberjack.Logger we can do log rotation to prevent it from growing infinitely.
-			Writer: &lumberjack.Logger{
-				Filename:   serverLogPath,
-				MaxSize:    100, // megabytes
-				MaxBackups: 3,
-				MaxAge:     28, // days
-			},
+	// Send logs with level higher than error to stderr.
+	log.AddHook(&writer.Hook{
+		Writer: os.Stderr,
+		LogLevels: []log.Level{
+			log.PanicLevel,
+			log.FatalLevel,
+			log.ErrorLevel,
+		},
+	})
+
+	if *verbose {
+		// If verbose, also send info, debug, warning logs to stdout.
+		log.AddHook(&writer.Hook{
+			Writer: os.Stdout,
 			LogLevels: []log.Level{
 				log.InfoLevel,
 				log.DebugLevel,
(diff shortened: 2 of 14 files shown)


src/golang/cmd/server/handler/archive_notification.go
```diff
@@ -17,8 +17,10 @@ import (
 // Method: POST
 // Params: notificationId
 // Request:
+//
```
vsreekanti, line 20: Seems like a bunch of these files weren't previously linted? Or linter rules changed somehow?

src/golang/cmd/server/main.go
```diff
@@ -17,8 +19,10 @@ var (
 		"",
 		"The path to .yml config file",
 	)
-	expose = flag.Bool("expose", false, "Whether you want to expose the server to the public.")
-	port   = flag.Int("port", connection.ServerInternalPort, "The port that the server listens to.")
+	expose        = flag.Bool("expose", false, "Whether the server will be exposed to the public.")
+	verbose       = flag.Bool("verbose", false, "Whether all logs will be shown in the terminal.")
+	port          = flag.Int("port", connection.ServerInternalPort, "The port that the server listens to.")
+	serverLogPath = filepath.Join(os.Getenv("HOME"), ".aqueduct", "server", "logs", "server_log")
```
vsreekanti, line 25: Nit: Let's make this page `~/.aqueduct/logs/server`. We might eventually want to unify our logging here (e.g., add operator logs separately as well).

src/golang/cmd/server/main.go
```diff
@@ -30,7 +34,41 @@ func main() {
...
+		ForceColors:  true,
+	})
+
+	// Always store all logs to a log file.
+	// With lumberjack.Logger we can do log rotation to prevent it from growing infinitely.
+	log.SetOutput(&lumberjack.Logger{
+		Filename:   serverLogPath,
+		MaxSize:    100, // megabytes
+		MaxBackups: 3,
+		MaxAge:     28, // days
```
vsreekanti, line 48: Any reason why 28 and not 30?

src/golang/cmd/server/main.go
```diff
@@ -30,7 +34,41 @@ func main() {
...
+	})
+
+	// Send logs with level higher than error to stderr.
+	log.AddHook(&writer.Hook{
+		Writer: os.Stderr,
+		LogLevels: []log.Level{
+			log.PanicLevel,
+			log.FatalLevel,
+			log.ErrorLevel,
```
vsreekanti, line 57: Let's do warning here by default?

src/python/bin/aqueduct
```diff
@@ -519,6 +513,13 @@ if __name__ == "__main__":
         action="store_true",
         help="Use this option to expose the server to the public.",
     )
+    start_args.add_argument(
+        "--verbose",
+        default=False,
+        action="store_true",
+        help="""Use this option to see all logs in the terminal. By default, only error logs are
```
vsreekanti, line 520: English nit: This sounds a little funny for a help message. I'd say something like , "If set to true, all log messages will be printed on stdout. By default, only error and warning-level logs are routed to stdout. All log messages can be found at $HOME/.aqueduct/logs/server".

src/golang/lib/logging/response.go
```diff
@@ -51,22 +52,37 @@ func LogRoute(
...
-		"Status":        status,
-		"Code":          statusCode,
-		"Component":     component,
-		"Route":         routeName,
-		"UserId":        ctx.Value(aq_context.UserIdKey),
-		"UserRequestId": ctx.Value(aq_context.UserRequestIdKey),
-		"Error":         errMsg,
-	}).Info()
+	if status == errorStatus {
+		log.WithFields(log.Fields{
```
vsreekanti, line 60: Style nit: Is this code deduplictable? i.e., we can do `logMsg := log.WithFields(...)` and then do `.Error()` or `.Info()` based on the status?

vsreekanti (approved):
This looks good to me, @cw75! I added a few tiny wording and style nits in line, but they should be quick to fix. It would also be nice if you could add a screenshot for what the new default and the `--verbose` behavior look like. Thanks!

The only thing I wasn't sure about was that there seem to be a bunch of linting changes in this PR. Did we change our linter rules somehow? Just want to make sure we stay on the same page about that. 

I'm good with this PR, but my backend knowledge is quite spotty, so might be worth having someone with more detailed knowledge take a look here.

src/golang/cmd/server/main.go
cw75: @vsreekanti the operator log files are already in `~/.aqueduct/server/logs/<OPERATOR_NAME>`. That's why I put the server log in this same directory as well. Do you still wanna change this to the new location? If so for consistency we also want to move the operator logs to the new directory, which may be tricky because existing workflow's logs were in the old directory and we may need to move them.

src/golang/cmd/server/main.go
cw75: The example code had 28. Maybe because Feb only have 28 days. :) I think it's fine though.

src/golang/cmd/server/handler/archive_notification.go
cw75: I think the golang linter version got updated in our Git action, and the updated version caused these changes. I was originally using the old version which didn't pass the linter test.

cw75:
@vsreekanti uploaded screenshot and addressed comments. I changed the log file name to `.../server` instead.

src/golang/cmd/server/main.go
vsreekanti: Got it, let's rename the file to `server` and then leave this directory.

vsreekanti:
Awesome, no more complaints from me!

PR #305, opened 2022-08-05
Eng 1326 api route to delete integration

eunice-chan:
Allow the user to delete integration. Directly tested endpoint.

ENG-1326


src/golang/cmd/server/handler/delete_integration.go
```diff
@@ -0,0 +1,113 @@
...
+
+	ok, err := h.IntegrationReader.ValidateIntegrationOwnership(
+		r.Context(),
+		integrationId,
+		aqContext.OrganizationId,
+		h.Database,
+	)
+	if err != nil {
+		return nil, http.StatusInternalServerError, errors.Wrap(err, "Unexpected error during integration ownership validation.")
+	}
```
likawind, line 68: nit: space after bracket after `if` block

src/golang/cmd/server/handler/delete_integration.go
```diff
@@ -0,0 +1,113 @@
...
+	integrationIdStr := chi.URLParam(r, routes.IntegrationIdUrlParam)
+	integrationId, err := uuid.Parse(integrationIdStr)
+	if err != nil {
+		return nil, http.StatusBadRequest, errors.Wrap(err, "Malformed integration ID.")
+	}
+
+	ok, err := h.IntegrationReader.ValidateIntegrationOwnership(
+		r.Context(),
+		integrationId,
+		aqContext.OrganizationId,
```
likawind, line 63: We should probably do this as a low-pri follow up. I think for deletion and if the integration is 'user only' type, we should restrict the permission to the user rather than the org.

src/golang/cmd/server/handler/delete_integration.go
```diff
@@ -0,0 +1,113 @@
...
+		return emptyResp, http.StatusInternalServerError, errors.Wrap(err, "Unable to delete integration.")
+	}
+	defer database.TxnRollbackIgnoreErr(ctx, txn)
+
+	err = h.IntegrationWriter.DeleteIntegration(ctx, args.integrationId, txn)
+	if err != nil {
+		return emptyResp, http.StatusInternalServerError, errors.Wrap(err, "Unexpected error occurred while deleting integration.")
+	}
+
+	if err := txn.Commit(ctx); err != nil {
```
likawind, line 104: I think the main motivation of using txn here is to make the change consistent with vault deletion. So we should actually commit after `Vault.Delete` being successful.

likawind (changes requested):
Great work! I think the main blocker is the use of transaction. Let's not worry about the ownership for now but file a task instead

src/golang/cmd/server/handler/delete_integration.go
eunice-chan: I see, so update `ValidateIntegrationOwnership` to check the user or the user's organization owns the integration rather than just check the organization owns the integration? I think I can do that in this PR

eunice-chan: Updated to check `SELECT COUNT(*) AS count FROM integration WHERE id = $1 AND (organization_id = $2 OR user_id = $3);`. Not entirely sure if this is what you are talking about, so will leave this comment open in case I misunderstood you.

likawind: Based on https://github.com/aqueducthq/aqueduct/blob/main/src/golang/cmd/server/handler/connect_integration.go#L128 It seems that the integration is `userOnly` if `user_id` exists. So I think the logic could be
* if user only doesn't exist, check if org ID match
* otherwise check if user id match
I would let you decide if you still want to do it in this PR as it may require some set up to verify the correctness

likawind (approved):
Great work!

PR #307, opened 2022-08-06
Fix a bug where non-preview operator zip files are mistakenly deleted

cw75:
If an operator failed during a regular workflow run, we had the logic where all of its downstream operators' zip files were deleted by mistake. This PR fixes the issue by passing a `isPreview` bool when constructing the operators, so that it relies on this value to determine whether the zip file should be GCed. We also deprecate the use of `shouldPersistResults `, as it's redundant.

I noticed that for some reason, integration tests for some Python versions are timing out, but I was able to successfully run all of them locally. Maybe there's something wrong with the GitHub worker or we should increase the timeout?

### DIFF
--- a/src/golang/lib/workflow/engine/aq_engine.go
+++ b/src/golang/lib/workflow/engine/aq_engine.go
@@ -289,10 +289,10 @@ func (eng *aqEngine) ExecuteWorkflow(
 	)
 	if err != nil {
 		workflowRunMetadata.Status = shared.FailedExecutionStatus
+		log.Errorf("Error when executing workflow: %v", err)
 	} else {
 		workflowRunMetadata.Status = shared.SucceededExecutionStatus
 	}
-	log.Errorf("Error when executing workflow: %v", err)
 }
 
 func (eng *aqEngine) PreviewWorkflow(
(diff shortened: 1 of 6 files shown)


src/golang/lib/workflow/engine/aq_engine.go
```diff
@@ -673,11 +672,9 @@ func (eng *aqEngine) execute(
 			}
 
 			// From here on we can assume that the operator has terminated.
-			if shouldPersistResults {
```
likawind, line 676: should this depends on `isPreview` ?

cw75: If I understand correctly, the value of `shouldPersistResults` is solely dependent on whether the request is a preview, so I replaced this with `isPreview`. @kenxu95 can you confirm?

likawind: I might misunderstood - after reading more code in context, is the behavior now that we always persist in `execute()` and only clean-up in `preview`?

cw75: I think so. In preview we never persist anything, but in `execute` (regular workflow execution) we do want to persist stuff.

kenxu95: yeah they are the same thing, but I would like us to be consistent - either we always call it `isPreview` or always `shouldPersistResults`.

kenxu95 (approved):

kenxu95 (commented):
@cw75 can we add an integration test for this? Publishing a broken flow twice should cover it right?

saurav-c (approved):

PR #320, opened 2022-08-10
[1/3] Eng 1323 Provide api ndpoint to edit integration

likawind:
This PR adds a route to support editing configs of an existing integration. The route performs the following steps:
* parse new configs. We assume a field needs to be updated if the string is not empty. There's a small issue that a user couldn't update a field to empty string, but I don't think it's an immediate concern (cc @vsreekanti ).
* fetch current integration
* merge configs
* test merged configs
* write merged configs back to metadata and vault

ENG-1323
ENG-170

We will add integration tests as a separate PR.
- [x] See #322  I have included a small demo of the changes. For the UI, this would be a screenshot or a Loom video.
- [x] (We will include integration test as follow-up. Tracked in ENG-1520) If this is a new feature, I have added unit tests and integration tests.


src/golang/cmd/server/handler/edit_integration.go
```diff
@@ -0,0 +1,318 @@
...
+		if v == "" {
+			continue // no update occurs
+		}
+
+		_, isReadonlyField := readOnlyFields[k]
+		curValue, existsInCurConfig := curConfigToUpdate[k]
+		if isReadonlyField && existsInCurConfig && curValue != v {
+			// Throw if:
+			// * field is read-only, and
+			// * field both exists in cur and new, and
```
saurav-c, line 117: Is there a reason why a field would be read-only, but not exist in the current config?

src/golang/cmd/server/handler/edit_integration.go
```diff
@@ -0,0 +1,318 @@
...
+		h.Database,
+		h.Vault,
+	); err != nil {
+		return emptyResp, statusCode, err
+	}
+
+	return emptyResp, http.StatusOK, nil
+}
+
+// ConnectIntegration connects a new integration specified by `args`. It returns a status code for the request
```
saurav-c, line 262: nit: update comment

src/golang/cmd/server/handler/edit_integration.go
```diff
@@ -0,0 +1,318 @@
...
+	}
+
+	hasPermission, err := h.IntegrationReader.ValidateIntegrationOwnership(
+		r.Context(),
+		integrationId,
+		aqContext.OrganizationId,
+		h.Database,
+	)
+	if err != nil {
+		return nil, http.StatusBadRequest, errors.Wrap(err, "Error validating integraiton ownership.")
```
saurav-c, line 155: nit: typo

src/golang/cmd/server/handler/edit_integration.go
```diff
@@ -0,0 +1,318 @@
...
+	}, http.StatusOK, nil
+}
+
+func (h *EditIntegrationHandler) Perform(ctx context.Context, interfaceArgs interface{}) (interface{}, int, error) {
+	args := interfaceArgs.(*EditIntegrationArgs)
+	id := args.IntegrationId
+
+	emptyResp := EditIntegrationResponse{}
+
+	integrationObject, err := h.IntegrationReader.GetIntegration(ctx, id, h.Database)
```
saurav-c, line 185: I think you should wrap all of the database operations in this method in a transaction. Otherwise it's possible the name gets updated but config update fails, etc.

src/golang/cmd/server/handler/edit_integration.go
```diff
@@ -0,0 +1,318 @@
...
+		changedFields[integration.ConfigColumn] = (*postgres_utils.Config)(&publicConfig)
+	}
+
+	txn, err := db.BeginTx(ctx)
+	if err != nil {
+		return http.StatusInternalServerError, errors.Wrap(err, "Unable to update integration.")
+	}
+	defer database.TxnRollbackIgnoreErr(ctx, txn)
+
+	// This is a user-specific integration
```
saurav-c, line 290: Wrong comment?

saurav-c (approved):
LGTM, left some comments!

src/golang/cmd/server/handler/edit_integration.go
likawind: @saurav-c when double checking the implementation here I think we are already putting the entire `update` in one transaction so that we only commit transaction when vault update succeeded. There's no other DB writes outside the transaction

PR #321, opened 2022-08-10
[2/3] Allow user to edit integrations in integration management.

likawind:
This PR adds changes to enable user to edit integration. The main changes include:
* set up redux store to handle integration connection. We also migrated add new integration use case to use redux store so that the state management is consistent.
* Updated integration dialog modal to allow passing in an `existing integration` externally so that the dialog can fill current value as placeholder.
* Improved state management of dialog modal by removing redundant state variables. See snowflake as an example.

ENG-170

We will upload demos in the follow-up PR.
- [x] See #322  I have included a small demo of the changes. For the UI, this would be a screenshot or a Loom video.


src/ui/common/src/components/integrations/dialogs/snowflakeDialog.tsx
```diff
@@ -16,27 +14,11 @@ const Placeholders: SnowflakeConfig = {
 };
 
 type Props = {
-  setDialogConfig: (config: IntegrationConfig) => void;
+  onUpdateField: (field: keyof SnowflakeConfig, value: Stringifiable) => void;
+  value?: SnowflakeConfig;
 };
 
-export const SnowflakeDialog: React.FC<Props> = ({ setDialogConfig }) => {
-  const [accountIdentifier, setAccountIdentifier] = useState<string>(null);
```
likawind, line 23: @agiron123 lmk if this make sense to you. Basically we directly pass external state as 'value' and use `onUpdateField` to set the external state, and in this way this component becomes stateless and we removed the original state copying

src/ui/common/src/components/integrations/dialogs/snowflakeDialog.tsx
```diff
@@ -16,27 +14,11 @@ const Placeholders: SnowflakeConfig = {
 };
 
 type Props = {
-  setDialogConfig: (config: IntegrationConfig) => void;
+  onUpdateField: (field: keyof SnowflakeConfig, value: Stringifiable) => void;
```
likawind, line 17: `keyof` ensures the `field` value is restricted to `SnowflakeConfig`'s fields

src/ui/common/src/components/integrations/dialogs/dialog.tsx
```diff
@@ -38,149 +40,58 @@ import { RedshiftDialog } from './redshiftDialog';
...
+  );
 
-  const handleSuccessToastClose = () => {
-    setShowSuccessToast(false);
-  };
+  const connectStatus = editMode ? editStatus : connectNewStatus;
+  const disableConnect =
+    !editMode &&
+    (!isConfigComplete(config) || name === '' || name === aqueductDemoName);
+  const setConfigField = (field: string, value: string) =>
```
likawind, line 81: this is basically `config[field] = value` in the language of state update

src/ui/common/src/components/integrations/dialogs/dialog.tsx
```diff
@@ -40,147 +49,58 @@ import { RedshiftDialog } from './redshiftDialog';
...
-        src={
-          'https://spiral-public-assets-bucket.s3.us-east-2.amazonaws.com/webapp/pages/integrations/csv-outline.png'
-        }
-      />
-    </Box>
+  const editMode = !!integrationToEdit;
+  const dispatch: AppDispatch = useDispatch();
+  const [config, setConfig] = useState<IntegrationConfig>(
+    editMode
+      ? { ...integrationToEdit.config } // make a copy to avoid accessing a state object
```
agiron123, line 71: Why exactly are we doing this instead of just integrationToEdit.config?

I can see that integrationToEdit is a prop above, but i'm not sure yet how this is being used as a component.

(This should be fine, just curious for my own edification)

src/ui/common/src/components/integrations/dialogs/dialog.tsx
agiron123: Yep, nice use of the spread operator here too!

src/ui/common/src/components/integrations/dialogs/dialog.tsx
```diff
@@ -330,5 +277,10 @@ export function isConfigComplete(
     return true;
   }
 
-  return Object.values(config).every((x) => x === undefined || (x && x !== ''));
+  return (
```
agiron123, line 280: This code is fun :) From looking at it, I think that it is checking to see if each of the config objects have some key set?

Perhaps we should add a small comment here for future people who come across this code about it's intended behavior.

src/ui/common/src/components/integrations/dialogs/snowflakeDialog.tsx
agiron123: Makes sense to me. Nice react pattern of "lifting state up" being used here :)

src/ui/common/src/reducers/integration.ts
```diff
@@ -225,10 +229,110 @@ export const handleTestConnectIntegration = createAsyncThunk<
...
+      apiKey: string;
+      service: Service;
+      name: string;
+      config: IntegrationConfig;
+    },
+    thunkAPI
+  ) => {
+    const { apiKey, service, name, config } = args;
+    Object.keys(config).forEach((k) => {
+      if (config[k] === undefined) {
```
agiron123, line 253: is config[k] allowed to be null? I'm not sure what all values are allowed for config keys, but might also be worth trying the truthy check: 
```
if (!config[k])
```

Again, depends on the values that config[k] is allowed to be. (not 100% sure here off the top of my head)

src/ui/common/src/utils/shared.ts
```diff
@@ -12,11 +12,12 @@ export type LoadingStatus = {
   err: string;
 };
 
+export function isInitial(status: LoadingStatus) {
```
agiron123, line 15: Maybe it's just me, but these functions feel a little redundant to me.

Perhaps we can just remove them?

agiron123 (approved):
Nice work! Left a couple comments here, but nothing here's a show stopper preventing this from shipping.

Resolve comments when ready and merge when ready. 🚢

src/ui/common/src/components/integrations/dialogs/dialog.tsx
likawind: exactly..but yes I agree we should have some comments here

src/ui/common/src/components/integrations/dialogs/dialog.tsx
likawind: Yeah it's a good question. This `integrationToEdit.config` is from the redux store for the current integration. The integration detail page renders this value.

In this way we need a copy of this state for the user to mess up during editing (so that the page won't re-render the value user's trying to update). Otherwise any input will directly get reflected on the current integration

src/ui/common/src/reducers/integration.ts
likawind: This code is copied from somewhere but I think what you suggested is correct

src/ui/common/src/utils/shared.ts
likawind: hmmmm okay I somehow feel they look a bit cleaner..The caller only write `isSuccess(someComplexReduxState.status) || isFailure(someComplexReduxState.status)` which is slightly better look / readable than `someComplexReduxState.status.loading === LoadingStatusEnum.Success || someComplexReduxState.status.loading === LoadingStatusEnum.Failure`

Happy to clean them up in another PR if you feel the original way is a better practice :)

src/ui/common/src/components/integrations/dialogs/dialog.tsx
likawind: actually should be sufficient to just do `.every(x => !!x)`

PR #334, opened 2022-08-11
Eng 1438 UI support to connect to S3 using credential file.

likawind:
This PR adds UI support for connecting to S3 using credential files. We support both specifying a path and read in live, and uploading the content of the file. The main changes involves adding UI components so that the users could select how to auth, and fill those new fields properly.

ENG-1436 ENG-1437 ENG-1438

Demo: https://www.loom.com/share/259db52d71074a0fad74603996fae7cb

The alert box is too close to detail card. Also fixed:
Screen Shot 2022-08-11 at 10 54 44 AM

### DIFF
--- a/src/ui/common/src/components/integrations/dialogs/s3Dialog.tsx
+++ b/src/ui/common/src/components/integrations/dialogs/s3Dialog.tsx
@@ -1,13 +1,27 @@
 import Box from '@mui/material/Box';
+import Tab from '@mui/material/Tab';
+import Tabs from '@mui/material/Tabs';
+import Typography from '@mui/material/Typography';
 import React, { useEffect, useState } from 'react';
 
-import { IntegrationConfig, S3Config } from '../../../utils/integrations';
+import {
+  FileData,
+  IntegrationConfig,
+  S3Config,
+  S3CredentialType,
+} from '../../../utils/integrations';
+import { readCredentialsFile } from './bigqueryDialog';
+import { IntegrationFileUploadField } from './IntegrationFileUploadField';
 import { IntegrationTextInputField } from './IntegrationTextInputField';
 
 const Placeholders: S3Config = {
+  type: S3CredentialType.AccessKey,
   bucket: 'aqueduct',
   access_key_id: '',
   secret_access_key: '',
+  config_file_path: '',
+  config_file_content: '',
+  config_file_profile: '',
 };
 
 type Props = {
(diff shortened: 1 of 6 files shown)


src/ui/common/src/components/integrations/dialogs/s3Dialog.tsx
```diff
@@ -61,4 +98,111 @@ export const S3Dialog: React.FC<Props> = ({ setDialogConfig }) => {
...
+      />
+
+      {configProfileInput}
+    </Box>
+  );
+
+  const configUploadTab = (
+    <Box>
+      <Typography variant="body2" color="gray.700">
+        Upload your aws credentials file and provide the name of the profile you
```
vsreekanti, line 130: capitalize AWS

src/ui/common/src/components/integrations/dialogs/s3Dialog.tsx
```diff
@@ -61,4 +98,111 @@ export const S3Dialog: React.FC<Props> = ({ setDialogConfig }) => {
...
+        would like to use. Typically, it&apos;s <code>~/.aws/credentials</code>.
```
vsreekanti, line 131: Since we want a profile here as well, let's use similar text to what I suggested above.

src/ui/common/src/components/integrations/dialogs/s3Dialog.tsx
```diff
@@ -61,4 +98,111 @@ export const S3Dialog: React.FC<Props> = ({ setDialogConfig }) => {
...
+        {/* uncomment once integration edit is ready:
+      <br/>
+      <br/>
+        Once connected, you would need to re-upload the file to update the credentials.
+      */}
+      </Typography>
+      <IntegrationFileUploadField
+        label={'AWS Credentials File*'}
+        description={'upload your credentials file here.'}
```
vsreekanti, line 140: Let's capitalize the U in upload to be consistent.

vsreekanti (approved):

vsreekanti:
Thanks @likawind! Some UI nits:
* We have a Tabs implementation that is adapted to match the look & feel of our UI. Let's use that instead of the default Tabs implementation. See [here](https://github.com/aqueducthq/aqueduct/blob/main/src/ui/common/src/components/primitives/Tabs.styles.ts). 
* What code highlighting are you using for the file preview? It seems different from the one we're using elsewhere, but it'd be nice to be uniform. 
* Once you upload the file, I think we have the name of the file after the colon -- that's a little confusing. I think we can just remove it, or we should use the full path for clarity.

likawind:
@vsreekanti where else we are using code block? Interestingly I'm using the one we have: https://github.com/aqueducthq/aqueduct/blob/main/src/ui/common/src/components/layouts/codeblock.tsx and there's no other reference to this component

likawind:
Updated according to most comments. A few small adjustments:
* I used 'you also need to specify profile' as for now this field is not optional
* I keep the message about file content updates in path mode
Screen Shot 2022-08-11 at 12 22 50 PM
Screen Shot 2022-08-11 at 12 22 35 PM

vsreekanti:
This looks good! Is there any reason the messages are on three different lines? I would just make it one paragraph.

vsreekanti:
re: codeblocks, I thought we were using it on the account page, but maybe that has been deprecated.

saurav-c:
@likawind One issue we will have to think about is whether all of these upload methods are compatible with 3rd party engines. For example, with Airflow the access key and file upload methods will work since we can directly upload those to the Airflow server, but the filepath approach will not work.

likawind:
@saurav-c 100% agree, I'm thinking about give some error message if the user want to register / trigger an wf on a 3rd party engine if the auth method is not compatible.

Another way is to upload the credentials at execution time esp. if we detect the execution is remote. There would be some security concerns. I have some ideas around pass around a vault pointer and let executors to read vault directly (I can share more on this offline). Either approach is probably out of scope of this PR and we could address later depending on priority, esp. after we have some new engine in-place.

Do you have other ideas for easy fix? I saw we are using file path for for airflow workflows but I didn't find how we let a remote airflow cluster access them.

PR #337, opened 2022-08-11
[ENG-1504] Fixes bug where CSV upload modal errors were incorrectly displayed

vsreekanti:

Previously, when we got an error message from the backend, we were parsing it with `response.text()` rather than `response.json()`, so the error message body was wrapped in many `title: {some blob}` layers. Fixing that parsing allows us to return the actual error message from the backend and then display it in a `<pre>` block here: 

image

ENG-1504


eunice-chan (approved):
Thanks for the change! Looks great

PR #340, opened 2022-08-12
Eng 1507 add stack trace to failed object deletion in workflow deletion [BE/3]

eunice-chan:
**skipping integration tests because it is currently failing integration tests. This is due to the fact the SDK side expects a `succeeded` field but that gets changed to `exec_state` and I update the SDK expectation in the next PR (Component https://github.com/aqueducthq/aqueduct/pull/342)**
## Describe your changes and why you are making these changes
Add stack traces / error messages to the failed object deletions as suggested in https://github.com/aqueducthq/aqueduct/pull/290.

BE Component https://github.com/aqueducthq/aqueduct/pull/340
UI Component https://github.com/aqueducthq/aqueduct/pull/341
SDK Component https://github.com/aqueducthq/aqueduct/pull/342

## Related issue number (if any)
ENG 1370
ENG 1507

## Checklist before requesting a review
- [x] I have created a descriptive PR title. The PR title should complete the sentence "This PR...".
- [x] I have performed a self-review of my code.
- [N/A] I have included a small demo of the changes. For the UI, this would be a screenshot or a Loom video.
- [N/A] If this is a new feature, I have added unit tests and integration tests.
- [x] I have run the integration tests locally and they are passing.
- [x] I have run the linter script locally (See `python3 scripts/run_linters.py -h` for usage).
- [x] All features on the UI continue to work correctly.
- [x] Added one of the following CI labels:
    - `run_integration_test`: Runs integration tests
    - `skip_integration_test`: Skips integration tests (Should be used when changes are ONLY documentation/UI)

PR #336, opened 2022-08-11
Add exception to traceback message

eunice-chan:
## Describe your changes and why you are making these changes
The context currently only displays the trace but not the actual exception class & message which is not as helpful as it could be. Since the stack trace in Python shows both, this results in the trace looking cut-off. I believe this is the same issue encountered in https://github.com/aqueducthq/aqueduct/pull/248 being described in ENG-1445.

Before:
```
  File "/Users/eunice/Desktop/aqueduct/src/python/aqueduct_executor/operators/connectors/tabular/relational.py", line 47, in delete
    sql_table = metadata.tables[table]
```
After:
```
  File "/Users/eunice/Desktop/aqueduct/src/python/aqueduct_executor/operators/connectors/tabular/relational.py", line 47, in delete
    sql_table = metadata.tables[table]
KeyError: 'delete_table'
```

## Related issue number (if any)
ENG-1445

## Checklist before requesting a review
- [x ] I have created a descriptive PR title. The PR title should complete the sentence "This PR...".
- [x] I have performed a self-review of my code.
- [x] I have included a small demo of the changes. For the UI, this would be a screenshot or a Loom video.
- [N/A] If this is a new feature, I have added unit tests and integration tests.
- [N/A] I have run the integration tests locally and they are passing.
- [x] I have run the linter script locally (See `python3 scripts/run_linters.py -h` for usage).
- [N/A] All features on the UI continue to work correctly.
- [x] Added one of the following CI labels:
    - `run_integration_test`: Runs integration tests
    - `skip_integration_test`: Skips integration tests (Should be used when changes are ONLY documentation/UI)

PR #341, opened 2022-08-12
Eng 1507 add stack trace to failed object deletion in workflow deletion [SDK/3]

eunice-chan:
Add stack traces / error messages to the failed object deletions as suggested in https://github.com/aqueducthq/aqueduct/pull/290.

BE Component https://github.com/aqueducthq/aqueduct/pull/340
UI Component https://github.com/aqueducthq/aqueduct/pull/341
SDK Component https://github.com/aqueducthq/aqueduct/pull/342

If we fail to delete the objects, an exception is thrown with this message (taken from the integration pytest):
```
E           Exception: Failed to delete 1 saved objects.
E           Failures
E           [aqueduct_demo] delete_table
E           >   File "/Users/eunice/Desktop/aqueduct/src/python/aqueduct_executor/operators/connectors/tabular/relational.py", line 47, in delete
E           >       sql_table = metadata.tables[table]
E           Sorry, we've run into an unexpected error! Please create bug report in github: https://github.com/aqueducthq/aqueduct/issues/new?assignees=&labels=bug&template=bug_report.md&title=%5BBUG%5D . We will get back to you as soon as we can.
```

ENG 1370
ENG 1507

### DIFF
--- a/integration_tests/sdk/utils.py
+++ b/integration_tests/sdk/utils.py
@@ -193,7 +193,7 @@ def delete_flow(client: aqueduct.Client, workflow_id: uuid.UUID) -> None:
     try:
         client.delete_flow(str(workflow_id))
     except Exception as e:
-        print("Error deleting workflow %s with exception %s" % (workflow_id, e))
+        print("Error deleting workflow %s with exception: %s" % (workflow_id, e))
     else:
         print("Successfully deleted workflow %s" % (workflow_id))
 
--- a/sdk/aqueduct/aqueduct_client.py
+++ b/sdk/aqueduct/aqueduct_client.py
@@ -38,7 +38,7 @@ from .integrations.sql_integration import RelationalDBIntegration
 from .logger import logger
 from .operators import Operator, OperatorSpec, ParamSpec, serialize_parameter_value
 from .param_artifact import ParamArtifact
-from .responses import SavedObjectDelete, SavedObjectUpdate
+from .responses import Error, SavedObjectDelete, SavedObjectUpdate
 from .utils import (
     _infer_requirements,
     generate_ui_url,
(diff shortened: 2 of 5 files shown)


sdk/aqueduct/aqueduct_client.py
```diff
@@ -454,18 +454,16 @@ def delete_flow(
         for integration in resp.saved_object_deletion_results:
             failed_for_integration = []
             for obj in resp.saved_object_deletion_results[integration]:
-                if not obj.succeeded:
+                if not obj.exec_state.status == ExecutionStatus.SUCCEEDED:
                     failed_for_integration.append(obj)
             if len(failed_for_integration) > 0:
                 failed_deletions[str(integration)] = [
-                    res.__dict__ for res in failed_for_integration
+                    json.loads(res.json()) for res in failed_for_integration
```
eunice-chan, line 461: [class].json() gives me the json as a string which gets escaped when I do `json.dumps`. However, I cannot serialize it directly with `json.dumps` so I serialize with `.json()` then `json.loads()` to get the format that is serializable by `json.dumps`.

integration_tests/sdk/utils.py
```diff
@@ -193,7 +193,7 @@ def delete_flow(client: aqueduct.Client, workflow_id: uuid.UUID) -> None:
     try:
         client.delete_flow(str(workflow_id))
     except Exception as e:
-        print("Error deleting workflow %s with exception %s" % (workflow_id, e))
+        print("Error deleting workflow %s with exception: %s" % (workflow_id, e))
```
likawind, line 196: try to use `%v` for errors, this might give you more information

sdk/aqueduct/aqueduct_client.py
```diff
@@ -449,22 +449,19 @@ def delete_flow(
...
-                    failed_for_integration.append(obj)
-            if len(failed_for_integration) > 0:
-                failed_deletions[str(integration)] = [
-                    res.__dict__ for res in failed_for_integration
-                ]
-                counts += len(failed_for_integration)
-        if counts > 0:
-            failures = json.dumps(failed_deletions, sort_keys=False, indent=4)
+                if obj.exec_state.status == ExecutionStatus.FAILED:
+                    assert isinstance(obj.exec_state.error, Error)
```
likawind, line 456: why do we need an `assert` here? I remembered the field is just `Optional[Error]`? If the type mismatch the serialization might just doesn't work. If it's only for detecting if the field is not `None`, we should just use `if  not obj.exec_state.error`

likawind (approved):
Mostly looks good! Added some minor nits

integration_tests/sdk/utils.py
eunice-chan: Was looking at Python formatting reference and I don't think `%v` exists in Python. https://python-reference.readthedocs.io/en/latest/docs/str/formatting.html

Ran a small example:
```
try:
    1 / 0
except Exception as e:
    print("%s" % e)
    print("%r" % e)
    print("%v" % e)
```
Output:
```
division by zero
ZeroDivisionError('division by zero')
Traceback (most recent call last):
  File "main.py", line 2, in <module>
    1 / 0
ZeroDivisionError: division by zero
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "main.py", line 6, in <module>
    print("%v" % e)
ValueError: unsupported format character 'v' (0x76) at index 1
```

I can do something like in https://github.com/aqueducthq/aqueduct/pull/336 to add the exception class' name.

likawind: oh nvm I mixed go with python here

sdk/aqueduct/aqueduct_client.py
eunice-chan: Yeah that makes sense, thanks!

PR #346, opened 2022-08-15
This PR adds regression testing for version upgrades

HarisChoudhary:
We add a github actions that perform validation of workflows against old and new releases. The steps performed for validation are similar what laid out in this page:

https://www.notion.so/aqueducthq/Runbook-Cutting-an-OSS-Release-b72400b7584845f0b0b4ff84bab4aedc#186142d8613f4740b620c90b18e41fee

We take screenshots of the workflow via a browser to simulate the user clicks on the frontend and upload them as screenshots. We also perform a simple validation of the flow through the api client.

Sample actions run:
https://github.com/aqueducthq/aqueduct/actions/runs/2862378175


.github/workflows/regression-tests.yml
```diff
@@ -0,0 +1,150 @@
...
+
+      - name: Start the server
+        run: (aqueduct start > $SERVER_LOGS_FILE 2>&1 &)
+
+      - name: Install Packages needed by the notebooks
+        run: pip3 install sklearn transformers torch
+
+      - name: Wait for server
+        timeout-minutes: 1
+        run: while ! echo exit | nc localhost 8080; do sleep 1; done
```
kenxu95, line 50: Not necessary for this PR but there sure starting to be a lot of duplication for this server-setup process. Maybe we should file a task for refactoring that functionality out?

.github/workflows/regression-tests.yml
```diff
@@ -0,0 +1,150 @@
...
+      # Grabs the pid of the process bound to port 8080 and kills it
+      - name: Kill the server
+        run: kill -9 $(lsof -nP -iTCP -sTCP:LISTEN | grep 8080 | awk '{print $2}')
+      
+      # install_local.py requires ~/.aqueduct to exist.
+      - name: Update aqueduct with latest code
+        run: python3 scripts/install_local.py --gobinary --sdk --executor
+
+      - name: Start the server again
+        run: (aqueduct start > $SERVER_LOGS_FILE 2>&1 &)
```
kenxu95, line 92: we should probably dump the server logs of the second run into a different log file?

.github/workflows/regression-tests.yml
kenxu95: This flow seems a bit more unique though

regression_tests/tests/workflow_comparer.py
```diff
@@ -0,0 +1,82 @@
...
+
+flow = client.flow(flow_id)
+checkpoint_path = os.path.join(args.path, flow_id)
+if not os.path.isdir(args.path):
+    os.mkdir(args.path)
+
+if args.checkpoint == "create":
+    flow.describe()
+    with open(checkpoint_path, 'w+') as sys.stdout:
+        flow.describe()
```
kenxu95, line 65: why are there two describe()'s here?

regression_tests/tests/workflow_comparer.py
```diff
@@ -0,0 +1,82 @@
...
+
+if args.checkpoint == "diff":
+    with open(checkpoint_path, 'r') as file:
+        previous_flow_info = file.read()
+
+    str_io = StringIO()
+    sys.stdout = str_io
+    flow.describe()
+    flow_info = str_io.getvalue()
+    sys.stdout = sys.__stdout__
```
kenxu95, line 77: I think it would be more readable/cleaner if we did this the way we do in our function executors.  See ExecutionState:

```
                stdout_log = io.StringIO()
                try:
                    with redirect_stdout(stdout_log):
                        flow.describe()
```

kenxu95 (changes requested):
This is pretty cool! Can you record a demo video of what this would look like from an oncall's perspective?

regression_tests/tests/workflow_comparer.py
HarisChoudhary: I was originally thinking the script runner might want to view the contents of the describe. Might not be necessary becuase in the failure case we print out the info anyways and that the only point where id imagine a user would care to see the contents.

HarisChoudhary:
I was thinking of creating a demo video and linking it to the notion page where we have the process outline for releases. I think ill do it after checkin becuase that would allow me to show the start correctly where we can dispatch the workflow from GitHub actions page instead of trigger on branch push.

regression_tests/tests/workflow_comparer.py
```diff
@@ -0,0 +1,81 @@
...
+from aqueduct import Client, get_apikey
+
+"""
+This script is used in regression testing to compare workflows.
+
+The checkpoint flag specifies the action to take
+create - stores workflow info as a checkpoint
+diff - compares the current state of the workflow with the last stored checkpoint
+
+Current comparisons are comparing the description of the flow each time for equality
```
kenxu95, line 17: super nit: can we add periods and colons to the end of comment sentences? It's a tiny thing but I think it definitely helps with readability.

kenxu95 (approved):
Looks good! It would have been nice to see a video that I could give feedback on, even if it wasn't exactly the flow, but we can always improve it later :)

PR #348, opened 2022-08-15
Eng 1123 enable operator with no return

Boyuan-Deng:

Before this implementation, is the user declared a @op for a function with null return, our system will produce a error. I wrote the code to enable operator with no return type. 


sdk/aqueduct/artifacts/none_artifact.py
```diff
@@ -0,0 +1,50 @@
...
+from aqueduct.artifacts import utils as artifact_utils
+from aqueduct.artifacts.artifact import Artifact
+from aqueduct.dag import DAG
+from aqueduct.enums import ArtifactType
+from aqueduct.error import AqueductError
+
+
+class NoneArtifact(Artifact):
+    """This class represents a generic artifact within the flow's DAG.
+    Currently, a generic artifact can be any artifact other than table, numeric, bool, or parameter.
```
kenxu95, line 15: This entire comment is irrelevant to NoneArtifact right.

sdk/aqueduct/artifacts/none_artifact.py
```diff
@@ -0,0 +1,50 @@
...
+        self._content = content
+        self._type = type
+
+    def get(self, parameters: Optional[Dict[str, Any]] = None) -> Any:
+        """Materializes the artifact.
+
+        Returns:
+            The materialized value.
+
+        Raises:
```
kenxu95, line 39: this method doesn't raise anything...

kenxu95: Just say it returns None.

sdk/aqueduct/artifacts/none_artifact.py
kenxu95: I'm curious if we need a separate NoneArtifact, or can we do everything in GenericArtifact? @cw75 I know you use content == None, but I wonder if we could just allow for content to be Optional and set other internal fields. Conceptually, this artifact seems unnecessary to me.

sdk/aqueduct/deserialize.py
```diff
@@ -34,11 +34,16 @@ def _read_bytes_content(content: bytes) -> bytes:
     return content
 
 
+def _read_none_content(content: bytes) -> bytes:
+    return content
```
kenxu95, line 38: shouldn't this return None?

kenxu95: To check my understanding, `content` is always an empty bytestring here right?

src/python/aqueduct_executor/operators/utils/utils.py
```diff
@@ -164,13 +169,26 @@ def _write_json_output(
...
+def _write_none_output(
+    storage: Storage,
+    output_path: str,
+    output: Any,
+) -> None:
+    # default to write a empty bytes
+    if output is None:
+        storage.put(output_path, bytes())
+    else:
+        raise Exception("function output is not None, Wrong serialization method")
```
kenxu95, line 181: grammar/style nit: Capitalize "function", uncapitalize "Wrong".

src/python/aqueduct_executor/operators/utils/utils.py
```diff
@@ -282,6 +302,8 @@ def infer_artifact_type(value: Any) -> ArtifactType:
         return ArtifactType.DICT
     elif isinstance(value, tuple):
         return ArtifactType.TUPLE
+    elif isinstance(value, type(None)):
```
kenxu95, line 305: Better to just do `elif value is None`

src/ui/common/src/components/workflows/nodes/NoneArtifactNode.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { ReactFlowNodeData } from '../../../utils/reactflow';
+import Node from './Node';
+
+type Props = {
+  data: ReactFlowNodeData;
+  isConnectable: boolean;
+};
+
+const NoneArtifactNode: React.FC<Props> = ({ data, isConnectable }) => {
+  console.log("node data: ", data)
```
kenxu95, line 13: nit: remove this before merging. 

Do you have a screenshot of this?

kenxu95 (changes requested):
Nice! Some comments and one discussion topic.

sdk/aqueduct/deserialize.py
Boyuan-Deng: yes, _read_none_content should always be a empty bytes(). And you are right on that _read_none_artifact should always return None

src/ui/common/src/components/workflows/nodes/NoneArtifactNode.tsx
Boyuan-Deng: image

sdk/aqueduct/artifacts/none_artifact.py
Boyuan-Deng: I discussed this with Cg on last friday and we decided to proceed with creating a separate NoneArtifact mainly because GenericArtifact is a temporary solution for artifact types that we did not get to come up with meaning type-specific functions eg( head() for table artifact). I think in the future, Cg wants to slowly pull artifact type out of the GenericAritifact type and assign them type specific functions. From my personal standpoint, NoneArtifact and GenericArtifact shows the same code implementation so it was kinda redundant.

cw75: As long as we can distinguish cases where the content is expected to be None v.s. the content is None because we haven't evaluated the content yet (constructing the artifact class from a flow run), I'm ok with merging.

src/python/aqueduct_executor/operators/utils/utils.py
```diff
@@ -164,13 +169,26 @@ def _write_json_output(
     storage.put(output_path, json.dumps(output).encode(_DEFAULT_ENCODING))
 
 
+def _write_none_output(
+    storage: Storage,
+    output_path: str,
+    output: Any,
+) -> None:
+    # default to write a empty bytes
+    if output is None:
```
kenxu95, line 178: The assumption here is that output is None already for our inference, so we shouldn't even check it. Can you just remove `output` completely and just have this be a single `storage.put()` line like the others?

kenxu95 (dismissed):
Where are the integration tests for this?

sdk/aqueduct/artifacts/param_artifact.py
```diff
@@ -17,6 +18,7 @@ def __init__(self, dag: DAG, artifact_id: uuid.UUID, from_flow_run: bool = False
         self._artifact_id = artifact_id
         # This parameter indicates whether the artifact is fetched from flow-run or not.
         self._from_flow_run = from_flow_run
+        self._type = ArtifactType.PARAM
```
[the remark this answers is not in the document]
Boyuan-Deng: No, this does not relate to this PR but notice it for being inconsistent with other artifact. Will remove it from this PR and ping chenggang for this change

sdk/aqueduct/artifacts/none_artifact.py
kenxu95: I think we can, since we can use the `from_flow_run` flag to check the second case.

Boyuan-Deng:
> Where are the integration tests for this?

Done

src/ui/common/src/components/workflows/nodes/NoneArtifactNode.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+type Props = {
+  data: ReactFlowNodeData;
+  isConnectable: boolean;
+};
+
+const NoneArtifactNode: React.FC<Props> = ({ data, isConnectable }) => {
+  return (
+    <Node
+      icon={faChartLine}
```
agiron123, line 15: @vsreekanti is this the icon that we wish to use?

Perhaps we can use some sort of "function" or "computation" icon here?

agiron123 (approved):
FE parts here look good to me. Nice work!

cw75:
I'll merge this after making sure it works well with the lazy execution stuff.

PR #349, opened 2022-08-15
Introduce concurrency back into integration tests 

kenxu95:
This PR makes two tweaks that seem like they fix our concurrency issue. Both have to do with the sqlite configuration:
1) Uses write-ahead logging, so that writers and readers don't grab the same database locks.
2) Adds a busy-timeout of 3 seconds, meaning that we will wait up to 3 seconds for a lock before giving up. This should eliminate `database is locked` errors. If they come back, we should consider bumping this timeout.

This saves a couple minutes each run. Interestingly enough, it seems like concurrency at 5 with python 3.9 and 3.10 is significantly slower :/ Otherwise we would be halving our runtime.

I've run the tests 20 times with concurrency 5 on github actions, and all except 1 pass, which is much better than before. I will continue to see if I can repro that issue, but in the meantime this help us out.

ENG-1468

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -73,7 +73,7 @@ jobs:
         env:
           SERVER_ADDRESS: localhost:8080
           INTEGRATION: aqueduct_demo
-        run: pytest . -rP --publish -n 1
+        run: pytest . -rP --publish -n 5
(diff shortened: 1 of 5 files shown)


saurav-c (approved):
LGTM

PR #351, opened 2022-08-15
ENG-918 Artifact Details Page and ENG-921 Metric Details Page

agiron123:
This PR contains the new artifacts detail page and metrics detail page and their supporting components.

The UI hasn't yet been fully integrated with the backend, so there are some places on each page that are still using mock data. This should be resolved pretty quickly now that the necessary supporting API routes have been merged into the ui-redesign branch.

Loom Demo:
https://www.loom.com/share/a4f3b33450514019be287c75f42cc068

https://linear.app/aqueducthq/issue/ENG-918
https://linear.app/aqueducthq/issue/ENG-921
https://linear.app/aqueducthq/issue/ENG-1591


src/golang/cmd/server/handler/get_operator_result.go
```diff
@@ -48,6 +48,17 @@ type GetOperatorResultHandler struct {
 	WorkflowDagResultReader workflow_dag_result.Reader
 }
 
+type GetOperatorResultResponse struct {
+	Name        string                `json:"name"`
+	Description string                `json:"description"`
+	ExecState   shared.ExecutionState `json:"exec_state"`
+	// TODO: Roll these into just ExecutionState when done.
```
agiron123, line 55: Remove comments.

agiron123: work for putting everything into the ExecState should be done now ✅

src/ui/common/src/components/pages/artifact/id/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+type ArtifactDetailsHeaderProps = {
+  artifactName: string;
+  createdAt?: string;
+  sourceLocation?: string;
+};
+
+const ArtifactDetailsHeader: React.FC<ArtifactDetailsHeaderProps> = ({
+  artifactName,
+  // TODO: add these back once we have support for getting createdAt and sourceLocation.
```
agiron123, line 64: @likawind Are we able to get this information yet with the latest changes that you've merged recently? We'll also need similar info for operators too if possible.

src/ui/common/src/components/pages/artifact/id/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+        height="24px"
+        width="24px"
+        style={{ marginRight: '8px' }}
+        icon={faCircleCheck}
+        color={'green'}
+      />
+      <Typography variant="h4" component="div">
+        {artifactName}
+      </Typography>
+      {/* <Typography marginTop="4px" variant="caption" component="div">
```
agiron123, line 80: TODO: Uncomment when we can get live info.

src/ui/common/src/components/pages/artifact/id/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+  ],
+};
+
+type ArtifactDetailsHeaderProps = {
+  artifactName: string;
+  createdAt?: string;
+  sourceLocation?: string;
+};
+
+const ArtifactDetailsHeader: React.FC<ArtifactDetailsHeaderProps> = ({
```
agiron123, line 62: This details header should probably be renamed as this component is used in each of our respective pages.

Perhaps we should rename this component to DetailsPageHeader

src/ui/common/src/components/pages/artifact/id/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+};
+
+const ArtifactDetailsPage: React.FC<ArtifactDetailsPageProps> = ({
+  user,
+  Layout = DefaultLayout,
+}) => {
+  const workflow = useSelector((state: RootState) => state.workflowReducer);
+  const navigate = useNavigate();
+  const dispatch: AppDispatch = useDispatch();
+  const { workflowId, workflowDagResultId, artifactId } = useParams();
```
likawind, line 102: looks like we could dispatch with `handleGetWorkflowDagResult` here once we merged all changes

src/ui/common/src/components/pages/artifact/id/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+      if (Object.keys(artifactResults).length < 1) {
+        return null;
+      }
+
+      return artifactResults[artifactId];
+    }
+  );
+
+  useEffect(() => {
+    document.title = 'Artifact | Aqueduct';
```
agiron123, line 116: Add a useEffect here for artifactResult.result and set the name of the webpage to be the artifact's name.

```
  const artifactName: string = artifactResult.result.name;
```

src/ui/common/src/components/pages/metric/id/index.tsx
```diff
@@ -0,0 +1,462 @@
...
+  const workflowDag = workflow.dagResults.find((currentDag) => {
+    return currentDag.id === workflowDagResultId;
+  });
+
+  const workflowDagId = workflowDag?.workflow_dag_id;
+  const operator = (workflow.dags[workflowDagId]?.operators ?? {})[
+    metricOperatorId
+  ];
+
+  // // Get the operatorSpec so that we can show more metadata about the operator.
```
agiron123, line 99: Remove these comments.

likawind:
@agiron123 doesn't see much issue by staring at codes, do you have some screenshots / demos to put in description just for reference? Also, do you prefer to merge with mock data first or also add the dispatch to new backend route in this PR?

src/ui/common/src/components/pages/metric/id/index.tsx
```diff
@@ -0,0 +1,462 @@
...
+  Layout = DefaultLayout,
+}) => {
+  const dispatch: AppDispatch = useDispatch();
+  const navigate = useNavigate();
+  const { workflowId, workflowDagResultId, metricOperatorId } = useParams();
+
+  const [inputsExpanded, setInputsExpanded] = useState<boolean>(true);
+  const [outputsExpanded, setOutputsExpanded] = useState<boolean>(true);
+
+  const workflow = useSelector((state: RootState) => state.workflowReducer);
```
agiron123, line 89: @likawind after we get this PR merged in, we can start incorporating your new api route for getting results populated in the dag structure.

likawind (approved):
I think it's in good shape!

src/ui/common/src/components/pages/metric/id/index.tsx
```diff
@@ -0,0 +1,462 @@
...
+          workflowDagResultId,
+          operatorId: metricOperatorId,
+        })
+      );
+    }
+  }, [workflow.dagResults]);
+
+  // Set up different metric input types for rendering in the inputs list.
+  // TODO: transform/handle response from API and render these appropriately.
+  const mockMetricInputs = [
```
agiron123, line 139: Wanted to simulate each type of node and show their appropriate icons.

This kind of code is a good candidate for a storybook ui story. I have a small task to bring back storybook, so perhaps I can move it there in the not too distant future.

src/ui/common/src/components/pages/metric/id/index.tsx
```diff
@@ -0,0 +1,462 @@
...
+      nodeIcon: tableArtifactNodeIcon,
+    },
+  ];
+
+  const artifactTypeToIconMapping = {
+    [ArtifactType.String]: stringArtifactNodeIcon,
+    [ArtifactType.Bool]: boolArtifactNodeIcon,
+    [ArtifactType.Numeric]: numericArtifactNodeIcon,
+    [ArtifactType.Dict]: dictArtifactNodeIcon,
+    // TODO: figure out if we should use other icon for tuple
```
agiron123, line 202: @vsreekanti @cw75 Do you have any thoughts on what font awesome icons we should use for these artifact types?

src/ui/common/src/components/pages/artifact/id/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+        color={'green'}
+      />
+      <Typography variant="h4" component="div">
+        {artifactName}
+      </Typography>
+      {/* <Typography marginTop="4px" variant="caption" component="div">
+                Created: {createdAt}
+            </Typography>
+            <Typography variant="caption" component="div">
+                Source: <Link>{sourceLocation}</Link>
```
likawind, line 84: what is `sourceLocation` here? Is that the same as `content_path` in `ArtifactResultResponse`?

src/ui/common/src/components/pages/metric/id/index.tsx
```diff
@@ -0,0 +1,462 @@
...
+import { LoadingStatusEnum } from '../../../../utils/shared';
+import DefaultLayout from '../../../layouts/default';
+import { LayoutProps } from '../../types';
+
+type MetricDetailsHeaderProps = {
+  artifactName: string;
+  createdAt?: string;
+};
+
+const MetricDetailsHeader: React.FC<MetricDetailsHeaderProps> = ({
```
agiron123, line 48: Since this is shared with the Artifact details page, let's create a shared component "DetailsPageHeader" and use that.

src/ui/common/src/components/pages/metric/id/index.tsx
```diff
@@ -0,0 +1,462 @@
...
+      return <CircularProgress />;
+    }
+
+    return operator.outputs.map((artifactId) => {
+      const artifactResult = workflow.artifactResults[artifactId];
+      if (!artifactResult) {
+        return null;
+      }
+
+      // TODO: Check the serialization_type of the artifacts and show a link
```
agiron123, line 264: @likawind we can get this from the spec right?

src/ui/common/src/components/pages/metric/id/index.tsx
```diff
@@ -0,0 +1,462 @@
...
+      // to table vew artifacts when present.
+      // TODO: Do the same thing that we're doing over in inputs :)
+
+      if (artifactResult.result) {
+        if (artifactResult.result.artifact_type === 'table') {
```
agiron123, line 269: TODO: support more artifact types.

src/ui/common/src/components/tables/CheckTableItem.tsx
```diff
@@ -0,0 +1,57 @@
...
+  faMinus,
+  faTriangleExclamation,
+} from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import Box from '@mui/material/Box';
+import React from 'react';
+
+import { theme } from '../../styles/theme/theme';
+
+interface MetricTableItemProps {
```
agiron123, line 13: TODO: Rename to CheckTableItemProps

src/ui/common/src/components/tables/KeyValueTable.tsx
```diff
@@ -0,0 +1,101 @@
...
+  tableAlign?: string;
+  tableType: KeyValueTableType;
+}
+
+const kvSchema: DataSchema = {
+  fields: [
+    { name: 'Title', type: 'varchar' },
+    { name: 'Value', type: 'varchar' },
+  ],
+  pandas_version: '0.0.1', // TODO: Figure out what to set this value to.
```
agiron123, line 33: I'm not entirely sure what the pandas_version does with regards to how we render our table. If someone here can shed some light on this, that would be awesome. otherwise, we'll continue setting yet ignoring this value for now 🙂

src/ui/common/src/components/tables/KeyValueTable.tsx
```diff
@@ -0,0 +1,101 @@
...
+                  return (
+                    <TableCell
+                      key={`cell-${rowIndex}-${columnIndex}`}
+                      align={tableAlign as TableCellProps['align']}
+                    >
+                      {tableType === KeyValueTableType.Metric ||
+                      column.name === 'Title' ? (
+                        value
+                      ) : (
+                        <CheckTableItem checkValue={value as string} />
```
agiron123, line 87: Maybe we should rename this component from KeyValueTable to something less generic.

At the moment what this table does is show metric and check names next to their execution status. perhaps OperatorExecStateTable may be a better name?

src/ui/common/src/components/tables/StickyHeaderTable.tsx
```diff
@@ -0,0 +1,121 @@
...
+import Typography from '@mui/material/Typography';
+import * as React from 'react';
+
+import { Data } from '../../utils/data';
+
+interface StickyHeaderTableProps {
+  data: Data;
+}
+
+const StickyHeaderTable: React.FC<StickyHeaderTableProps> = ({ data }) => {
```
agiron123, line 19: Maybe rename this component to PaginatedTable?

While it has a sticky header, I'd argue that the pagination is more of a defining feature of this table.

src/ui/common/src/components/workflows/nodes/NodeStatus.tsx
```diff
@@ -9,29 +9,25 @@ type Props = {
...
 
 export const NodeStatus: React.FC<Props> = ({ execState }) => {
-  const statusIcons = [];
-  const runStatus = execState.status.toLowerCase();
+  let statusIcon = null;
+  const runStatus: string = execState.status.toLowerCase();
 
   if (runStatus === ExecutionStatus.Succeeded) {
-    statusIcons.push(<Chip label="Succeeded" color="success" size="small" />);
+    statusIcon = <Chip label="Succeeded" color="success" size="small" />;
```
agiron123, line 16: I have no idea why this was an array in the past, since we can only have one status, so i changed it to a singular value instead 🙂

src/ui/common/src/components/pages/metric/id/index.tsx
likawind: we don't have that for artifact yet but it's easy to add. I didn't include serialization_type in first cut just because I didn't think we are using it

likawind: and yes, assuming we have it we should get it from output artifact, which is easy to figure out using `.outputs` fields of this operator

src/ui/common/src/utils/data.ts
```diff
@@ -21,6 +21,9 @@ export type DataSchema = {
 
 export type Data = {
   schema?: DataSchema;
+  //data: (string | number | boolean)[][];
+  // data is an array of objects whose keys correspond to the schema above.
+  // each record of data is a row on the table.
   data: any[];
```
agiron123, line 27: Ready for some fun?
This is a weird typing issue that I encountered recently only when trying to make some mock data to be used in a table. I followed the type to a T (no pun intended), yet found that nothing was rendered on the table.

Next, I looked at the real payloads that we were rendering in preview, and they were  simply an array of objects, where each object in the array corresponds to a row, and each key in the object corresponds to a column in the table. This is different than the 2d array as described in the type above.

Now, as for why we didn't catch this issue earlier? Well, after getting preview data from the API, we would parse the response with JSON.parse(response), and then cast that as a Data object. 

i.e. ```const parsedData = JSON.parse(response) as Data;```

This doesn't actually get type checked and slipped through as a result 😄 . I need to do some more digging on the Typescript docs as to why, but it was a fun thing to learn at the time.

Last thing I promise :) 

The type for data can be made stronger (using ```any``` is frowned upon), so here's what we can do: use a key value pair where the values are of type string | number | boolean. 

Like so:

```data: {[key: string]: string | number | boolean}```

File: src/ui/common/src/utils/operators.ts [hunk 21]
```diff
@@ -140,7 +140,13 @@ export function normalizeOperator(op): Operator {
   return op;
 }
 
-export type GetOperatorResultResponse = ExecState;
+export type GetOperatorResultResponse = {
+  name: string;
+  description: string;
+  // TODO: Remove status and just have exec_state.
```
agiron123, line 146: I think we should be safe to do this now. Hopefully our typechecking has our back here :)

src/ui/common/src/components/pages/metric/id/index.tsx
vsreekanti: I think @cw75 already did this for the DAG visualization in the `Node` implementation. Maybe we could copy those over?

src/ui/common/src/components/tables/KeyValueTable.tsx
vsreekanti: @cw75 @kenxu95?

src/ui/common/src/components/workflows/nodes/NodeStatus.tsx
vsreekanti: This component was for some reason copied from the `WorkflowStatus` component. That one had multiple badges because we were showing a badge for success/failure but also for any warnings. I'm not sure why we have two separate components though -- I created a [task](https://linear.app/aqueducthq/issue/ENG-1630/refactor-workflowstatus-and-nodestatus-components-to-be-a-unified) a while back to merge them.

agiron123: @vsreekanti makes sense. I can tackle that after I get this PR in.

src/ui/common/src/components/pages/artifact/id/index.tsx
agiron123: @likawind good catch, sure can do that here. Will get that in after I verify that the other PR cleanup work is working properly.

src/ui/common/src/components/pages/artifact/id/index.tsx
agiron123: ✅

src/ui/common/src/components/workflows/nodes/NodeStatus.tsx
vsreekanti: I actually don't think we need to spend time on this one immediately. Let's get this PR in and keep pushing on the UI redesign, and we can come back to some of these smaller tasks later on if that makes sense to you.

src/ui/common/src/components/tables/CheckTableItem.tsx
agiron123: ✅

src/ui/common/src/components/tables/KeyValueTable.tsx
agiron123: ✅

PR #352, opened 2022-08-15
Eng 1557 Fix bug where the required Python packages on the client machine aren't automatically installed on the server

eunice-chan:
- The cause of the bug is that the server was unable to install the modules in requirement.txt because it is looking for local files that did not exist on that machine from in-place installs (`pip(3) install (-e) . `).
- Warn user if they have any local installs.
- Update blacklisted list to include aqueduct_sdk / aqueduct_ml (the format aqueduct_xyz would be in the case of installing with `pip3 install -e .`)

Checked with a requirements.txt that included modules in the following formats:
```
appnope==0.1.3
aqueduct-ml @ file:///Users/eunice/Desktop/aqueduct/src/python
-e git+ssh://git@github.com/aqueducthq/aqueduct.git@4990b42ed9fcc6758bc7b6fc2d7d475a5b1d7477#egg=aqueduct_sdk&subdirectory=sdk
```
Eng 1557

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -37,7 +37,7 @@ jobs:
           go-version: 1.16.3
       
       - name: Install aqueduct-ml
-        run: pip3 install aqueduct-ml
+        run: python3 -m pip install aqueduct-ml
 
       - name: Start the server
         run: (aqueduct start > $SERVER_LOGS_FILE 2>&1 &)
@@ -58,7 +58,7 @@ jobs:
         run: (aqueduct start > $SERVER_LOGS_FILE 2>&1 &)
 
       - name: Install packages needed for testing
-        run: pip3 install nltk matplotlib pytest-xdist
+        run: python3 -m pip install nltk matplotlib pytest-xdist
 
       - name: Wait for server again
         timeout-minutes: 5
--- a/src/python/aqueduct_executor/operators/function_executor/prune_requirements.py
+++ b/src/python/aqueduct_executor/operators/function_executor/prune_requirements.py
@@ -1,6 +1,5 @@
 import argparse
 
-
 def run(local_path: str, requirements_path: str, missing_path: str) -> None:
     with open(local_path, "r") as f:
         local_req = set(f.read().split("\n"))
@@ -10,9 +9,8 @@ def run(local_path: str, requirements_path: str, missing_path: str) -> None:
 
     missing = []
     for r in required:
-        if r not in local_req:
+        if r not in local_req and "@ file" not in r:
             missing.append(r)
-
     if len(missing) > 0:
         with open(missing_path, "w") as f:
             f.write("\n".join(missing))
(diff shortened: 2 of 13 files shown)


src/python/aqueduct_executor/operators/function_executor/prune_requirements.py
```diff
@@ -10,9 +10,8 @@ def run(local_path: str, requirements_path: str, missing_path: str) -> None:
 
     missing = []
     for r in required:
-        if r not in local_req:
+        if r not in local_req and "@ file" not in r:
             missing.append(r)
-
```
cw75, line 15: Could you add back this line break?

cw75 (approved):
One nit otherwise LGTM!

cw75:
@eunice-chan Also I think we also modified the `.sh` file to switch `pip freeze` to `pip3 freeze`?

eunice-chan:
@cw75 Good catch--added the `start-function-executor.sh`

.github/workflows/integration-tests.yml
```diff
@@ -58,7 +58,7 @@ jobs:
         run: (aqueduct start > $SERVER_LOGS_FILE 2>&1 &)
 
       - name: Install packages needed for testing
-        run: pip3 install nltk matplotlib pytest-xdist
+        run: python3 -m pip install nltk matplotlib pytest-xdist
```
kenxu95, line 61: was this an issue?

src/python/aqueduct_executor/operators/function_executor/prune_requirements.py
```diff
@@ -10,7 +10,7 @@ def run(local_path: str, requirements_path: str, missing_path: str) -> None:
 
     missing = []
     for r in required:
-        if r not in local_req:
+        if r not in local_req and "@ file" not in r:
```
kenxu95, line 13: Can you add a comment for why we're filtering out `@ file`?

.github/workflows/integration-tests.yml
kenxu95: I'm worried about changing everything to this because our user documentation doesn't usually say you have to prefix with `python -m`

eunice-chan: This is a general change to avoid the situation where user’s pip3 and python3 aren’t compatible. I encountered the issue because my pip3 was installing things to usr/local or something like that instead of the conda environment. 

I don't think it would be an issue on the Github Actions so I can change it back to what it was before. I made this change to the Github Actions commands to be consistent to my update to how we call pip in the rest of the codebase.

kenxu95: ok - I think its fine leaving it, since we do use `python -m` everywhere in our system. It might be helpful changing our gitbook documentation to help users that have similar incompatibilities? What do you think @cw75

cw75: Yeah I think in general Python3 -m is good in avoiding the potential inconsistencies. We can change the Gitbook documentation as well. I think we should be consistent to use it everywhere, including the Git actions @eunice-chan

eunice-chan: Added documentation https://github.com/aqueducthq/gitbook/pull/10

PR #353, opened 2022-08-15
Allows config to be modified for running server

saurav-c:
This PR allows the server configuration to be modified while the server is running. This is mostly important for the `StorageConfig` as we have features that require us to modify the storage config. In fact, right now the only field that can be modified is `StorageConfig`. All other fields are static and set once the server loads the config file when it starts up.

Specifically, the following changes have been made:
- The server reads the config from the config file at server start up and stores this config in a global variable accessible only by the `config` package. All access to config fields must be through this package instead of storing a server configuration field in the server struct or passing down some of these fields via the different handlers.
- Removes the `storageConfig` field from all handlers. Instead, we are making `StorageConfig` a field of the `AqContext` as the same StorageConfig should be used for the context of a request, since it is possible that the storageConfig may be modified while a request is in progress. This access of the `StorageConfig` only applies when we are performing an operation outside of the context of workflow execution. During workflow execution, we should not be using the current global `StorageConfig`, but instead use the `StorageConfig` that is associated with the workflow. This information is stored in the database.


src/golang/lib/job/spec.go
```diff
@@ -82,10 +82,11 @@ type WorkflowRetentionSpec struct {
...
 	BaseSpec
-	WorkflowId     string                `json:"workflow_id" yaml:"workflowId"`
-	GithubManager  github.ManagerConfig  `json:"github_manager" yaml:"github_manager"`
-	Parameters     map[string]string     `json:"parameters" yaml:"parameters"`
-	AqPath         string                `json:"aq_path" yaml:"aqPath"`
+	WorkflowId    string               `json:"workflow_id" yaml:"workflowId"`
+	GithubManager github.ManagerConfig `json:"github_manager" yaml:"github_manager"`
+	Parameters    map[string]string    `json:"parameters" yaml:"parameters"`
+	AqPath        string               `json:"aq_path" yaml:"aqPath"`
+	// TODO: Deprecate StorageConfig
```
saurav-c, line 89: We can't remove this just yet, since existing workflow specs will still have this field for deserialization might fail. (I need to confirm if that is the case or not.)

hsubbaraj-spiral (approved):
changes look good to me.

src/golang/config/config.go
```diff
@@ -6,30 +6,78 @@ import (
...
+func RetentionJobPeriod() string {
+	return globalConfig.RetentionJobPeriod
+}
+
+// APIKey returns the API key the user must use when issuing requests.
+func APIKey() string {
+	return globalConfig.ApiKey
+}
+
+// Storage returns the storage layer config.
```
kenxu95, line 48: This is a snapshot/copy of the storage config right? Can we make that explicit in the name (eg. CopyStorageConfig() or SnapshotStorageConfig()). I ask because everywhere else we seem to use pointers to the same config, and this place is the exception, and an important one.

src/golang/lib/context/context.go
```diff
@@ -21,7 +23,8 @@ const (
 
 type AqContext struct {
 	user.User
-	RequestId string
+	RequestId     string
+	StorageConfig *shared.StorageConfig
```
kenxu95, line 27: Can you add a comment here that the storage config is a snapshot that is only meant to live for the length of the request. I think that would be super helpful for readers

src/golang/lib/workflow/dag/workflow_dag.go
```diff
@@ -99,7 +98,7 @@ func NewWorkflowDag(
 			artifactIDToContentPath[artifactID],
 			artifactIDToMetadataPath[artifactID],
 			artifactResultWriter,
-			storageConfig,
+			&dbWorkflowDag.StorageConfig,
```
kenxu95, line 101: I got a little confused here, since `dbWorkflowDag` makes me think that the storage config is something set by the user. However, it looks like this is the same storage config that is in `AqContext` right? 

If so, then this seems like something that will become clearer once we stop serializing requests into `dbWorkflowDag`s.

kenxu95: And one more clarification: the storage config is set here at the beginning, and then written to the workflow spec. Subsequent scheduled runs will then use whatever is on the workflow spec, and that is immutable over the lifecycle  of the workflow?

kenxu95 (approved):
Looks good, I just have some comments around readability, as well as some general questions!

src/golang/config/config.go
saurav-c: I wanted to return a copy to make it explicit that the only way to modify the storage config is via the `UpdateStorage` function in this pkg. Do you think it's sufficient to just state this in the comments?

src/golang/lib/workflow/dag/workflow_dag.go
saurav-c: So the storage config is `dbWorkflowDag` is set based on the value of the global storage config at the time of the workflow registration. Once this value is set, we persist it to the database, and that storage config is used for the entire lifecycle of the workflow. Right now we don't support modifying the storage config for existing workflows.

PR #358, opened 2022-08-16
Update warning message when checking pip freeze requirements to be more informative

cw75:
And add the logic to only check when the aqueduct address is not localhost.

### DIFF
--- a/sdk/aqueduct/aqueduct_client.py
+++ b/sdk/aqueduct/aqueduct_client.py
@@ -124,12 +124,15 @@ class Client:
 
         # Check if "@ file" in pip freeze requirements and warn user.
         if not "localhost" in aqueduct_address:
+            skipped_packages = []
             for requirement in infer_requirements():
                 if "@ file" in requirement:
-                    warnings.warn(
-                        "Your local Python environment has at least one package installed from the local file system. These won't be installed into your server's Python environment."
-                    )
-                    break
+                    skipped_packages.append(requirement.split(" ")[0])
+            if len(skipped_packages) > 0:
+                warnings.warn(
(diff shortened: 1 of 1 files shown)


sdk/aqueduct/aqueduct_client.py
```diff
@@ -122,13 +122,14 @@ def __init__(
...
-                warnings.warn(
-                    "You have installed at least one module from your local file system. These won't be installed in the server environment."
-                )
-                break
+        # Check if "@ file" in pip freeze requirements and warn user.
+        if not "localhost" in aqueduct_address:
+            for requirement in infer_requirements():
+                if "@ file" in requirement:
+                    warnings.warn(
+                        "Your local Python environment has at least one package installed from the local file system. These won't be installed into your server's Python environment."
```
vsreekanti, line 130: I'd rather we just print out each requirement name.

vsreekanti: Server is also kinda vague here because we might be running on Airflow, k8s, etc. Can we say "These packages won't be installed when running your workflow."

vsreekanti (approved):

cw75:
@vsreekanti how about this:
Screen Shot 2022-08-15 at 7 25 52 PM

vsreekanti:
Yep, the only change I'd make is I'd put a colon after workflow and then print the list of packages right after. LGTM otherwise!

cw75:
@vsreekanti Yeah I just realized that as well. Also tweaked the wording a bit, see below:
Screen Shot 2022-08-15 at 7 28 16 PM

vsreekanti:
Perfect!

PR #362, opened 2022-08-16
Adds backend support for changing storage to S3 integration

saurav-c:
This PR adds the backend support for changing the metadata store to a newly connected S3 integration. There is a lot of complex logic because of differences in assumptions in what our S3 storage client expects for authentication vs. what the S3 integration connector client expects for authentication. We have to perform this translation.

ENG 1553

### DIFF
--- a/src/golang/cmd/server/handler/connect_integration.go
+++ b/src/golang/cmd/server/handler/connect_integration.go
@@ -80,12 +80,13 @@ func (h *ConnectIntegrationHandler) Prepare(r *http.Request) (interface{}, int, 
 		return nil, http.StatusBadRequest, errors.Newf("%s integration type is currently not supported", service)
 	}
 
-	setStorage, err := request.ParseSetIntegrationAsStorage(r, service)
+	config := auth.NewStaticConfig(configMap)
+
+	// Check if this integration should be used as the new storage layer
+	setStorage, err := checkIntegrationSetStorage(r.Context(), service, config)
 	if err != nil {
 		return nil, http.StatusBadRequest, errors.Wrap(err, "Unable to connect integration.")
 	}
-
-	config := auth.NewStaticConfig(configMap)
 
 	return &ConnectIntegrationArgs{
 		AqContext:  aqContext,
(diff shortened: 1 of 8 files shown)


src/golang/cmd/server/handler/connect_integration.go
```diff
@@ -47,10 +52,11 @@ func (*ConnectIntegrationHandler) Headers() []string {
...
 	*aq_context.AqContext
-	Name     string              // User specified name for the integration
-	Service  integration.Service // Name of the service to connect (e.g. Snowflake, Postgres)
-	Config   auth.Config         // Integration config
-	UserOnly bool                // Whether the integration is only accessible by the user or the entire org
+	Name       string              // User specified name for the integration
+	Service    integration.Service // Name of the service to connect (e.g. Snowflake, Postgres)
+	Config     auth.Config         // Integration config
+	UserOnly   bool                // Whether the integration is only accessible by the user or the entire org
+	SetStorage bool                // Whether the integration should be used as the storage layer
```
kenxu95, line 59: nit: maybe `SetAsStorage` is a bit more clear/

src/golang/cmd/server/handler/connect_integration.go
```diff
@@ -248,3 +277,117 @@ func validateAirflowConfig(
...
+		// Determine profile name by looking for [profile_name]
+		i := strings.Index(c.ConfigFileContent, "[")
+		if i < 0 {
+			return nil, errors.New("Unable to determine AWS credentials profile name.")
+		}
+
+		j := strings.Index(c.ConfigFileContent, "]")
+		if j < 0 {
+			return nil, errors.New("Unable to determine AWS credentials profile name.")
+		}
```
kenxu95, line 373: Might be better to use a regex library? https://gobyexample.com/regular-expressions

Just curious, do we control the format of `c.ConfigFileContent`? Eg. can we always expect a "[<profile_name>]" format or is it user-controlled

kenxu95 (approved):
Makes sense to me

src/golang/cmd/server/handler/connect_integration.go
saurav-c: I'm not 100% sure, I think in most cases we can. Wei would have more context, but I don't want to ping him on PTO. @cw75 Do you have any thoughts here?

I'm thinking that if the file does not have a profile name we can just go with `default`?

hsubbaraj-spiral (approved):

src/golang/cmd/server/handler/connect_integration.go
saurav-c: Actually after thinking about this more, since this is an AWS credentials file convention, I think we should throw an error if the file is not formatted correctly. Because, even if we assume `default`, the actual AWS API calls will fail when the AWS client tries to find the `default` profile in the config file.

kenxu95: Yeah I think its safest just to throw.

PR #363, opened 2022-08-16
Remove double workflow deletion from test_delete_workflow_* tests

eunice-chan:
Remove flow_id from flow_id_to_delete set once successfully deleted to avoid attempting to delete a workflow twice.
No more `The organization does not own this workflow` errors.

```
================================================================= PASSES ==================================================================
_______________________________________________ test_delete_workflow_invalid_saved_objects ________________________________________________
---------------------------------------------------------- Captured stdout call -----------------------------------------------------------
Url:  http://localhost:8080/workflow/ad273a81-82cc-4203-8aba-78e8ebdf4d39
Workflow registration succeeded. Workflow ID: ad273a81-82cc-4203-8aba-78e8ebdf4d39
Workflow ad273a81-82cc-4203-8aba-78e8ebdf4d39 was created and ran successfully at least 1 times!
Successfully deleted workflow ad273a81-82cc-4203-8aba-78e8ebdf4d39
___________________________________________________ test_delete_workflow_saved_objects ____________________________________________________
---------------------------------------------------------- Captured stdout call -----------------------------------------------------------
Url:  http://localhost:8080/workflow/39834122-8ee5-439c-8aa4-0bc671bf1dfb
Workflow registration succeeded. Workflow ID: 39834122-8ee5-439c-8aa4-0bc671bf1dfb
Workflow 39834122-8ee5-439c-8aa4-0bc671bf1dfb was created and ran successfully at least 1 times!
Url:  http://localhost:8080/workflow/39834122-8ee5-439c-8aa4-0bc671bf1dfb
Workflow registration succeeded. Workflow ID: 39834122-8ee5-439c-8aa4-0bc671bf1dfb
Workflow 39834122-8ee5-439c-8aa4-0bc671bf1dfb was created and ran successfully at least 2 times!
________________________________________________ test_delete_workflow_saved_objects_twice _________________________________________________
---------------------------------------------------------- Captured stdout call -----------------------------------------------------------
Url:  http://localhost:8080/workflow/e6546e3e-b8dc-49a9-9d11-2569920b92c0
Workflow registration succeeded. Workflow ID: e6546e3e-b8dc-49a9-9d11-2569920b92c0
Workflow e6546e3e-b8dc-49a9-9d11-2569920b92c0 was created and ran successfully at least 1 times!
Url:  http://localhost:8080/workflow/ac40ffef-a9de-4a93-af6b-3c0622f4255a
Workflow registration succeeded. Workflow ID: ac40ffef-a9de-4a93-af6b-3c0622f4255a
Workflow ac40ffef-a9de-4a93-af6b-3c0622f4255a was created and ran successfully at least 1 times!
=========================================================== 3 passed in 36.84s ============================================================
```


integration_tests/sdk/delete_workflow_test.py
```diff
@@ -136,17 +138,23 @@ def test_delete_workflow_saved_objects_twice(client):
...
+        assert "delete_table" in tables_1
 
         tables = client.flow(flow_2_id).list_saved_objects()
-        assert "delete_table" in [item.object_name for item in tables[get_integration_name()]]
+        tables_2 = set([item.object_name for item in tables[get_integration_name()]])
+        assert "delete_table" in tables_2
+
+        assert tables_1 == tables_2
 
         # No SDK function to do this so we query the endpoint directly to see delete_table is properly created at the integration.
```
kenxu95, line 150: Why can't we do `integration.sql("select * from delete_table").get()` to check that the table exists? I'd prefer to keep actual endpoints out of the SDK if possible.

integration_tests/sdk/delete_workflow_test.py
```diff
@@ -136,17 +138,23 @@ def test_delete_workflow_saved_objects_twice(client):
         flow_2_id = flow_list[1]
```
kenxu95, line 138: for line 130, aren't the two flow ids going to be exactly the same? It should be `num_runs=2` then, and we probably should just track a single flow ids to be less confusing. You can just check that the flow id returned by both `run_flow_test()` calls is the same, so you don't have to call `list_saved_objects()` twice with the exact same arguments.

integration_tests/sdk/delete_workflow_test.py
```diff
@@ -136,17 +138,23 @@ def test_delete_workflow_saved_objects_twice(client):
...
+        assert tables_1 == tables_2
 
         # No SDK function to do this so we query the endpoint directly to see delete_table is properly created at the integration.
         tables_response = get_response(client, endpoint).json()
         assert "delete_table" in set(tables_response["object_names"])
 
         client.delete_flow(flow_1_id, saved_objects_to_delete=tables, force=True)
 
+        flow_ids_to_delete.remove(flow_1_id)
+
```
kenxu95, line 157: Instead of sleeping 1 second, can we be a bit more robust and keep periodically checking for the table to no longer exist until a longer timeout is hit (eg. 1 min). When we run things concurrently this might slow down a bit.

integration_tests/sdk/delete_workflow_test.py
```diff
@@ -159,6 +167,12 @@ def test_delete_workflow_saved_objects_twice(client):
             client.delete_flow(flow_2_id, saved_objects_to_delete=tables, force=True)
         assert str(e_info.value).startswith("Failed to delete")
 
+        # Failed to delete tables, but flow should be removed.
+        with pytest.raises(Exception) as e_info:
+            client.delete_flow(flow_2_id)
+        assert str(e_info.value).startswith("The organization does not own")
+        flow_ids_to_delete.remove(flow_2_id)
```
kenxu95, line 174: maybe instead of this no-op, maybe can assert that `len(flow_ids_to_delete) == 0`? A `remove()` call that doesn't do anything is just another source of confusion.

kenxu95 (changes requested):
Thanks! I have some feedback for the readability of these tests. Remember tests are not only supposed to serve as checks on the system, but also as live documentation for us, so it would be great if they are as readable as possible.

integration_tests/sdk/delete_workflow_test.py
eunice-chan: For this specific test, I'm checking the successful deletion case and unsuccessful deletion case works as expected. To test this, I have two workflows that write to the same table. When I delete the table in the first workflow, it is successful but when I delete it in the second workflow, it is unsuccessful because the table has already been deleted.

integration_tests/sdk/delete_workflow_test.py
eunice-chan: That's a good idea! Thanks for the suggestion. I will just be checking the following:
```
        with pytest.raises(AqueductError) as e_info:
             integration.sql("SELECT * FROM delete_table").get()
```

Let me know if there is anything else you think would be good to check.

integration_tests/sdk/delete_workflow_test.py
eunice-chan: I think this comment is related to the previous one. I actually created two flows so this isn't a no-op.

integration_tests/sdk/delete_workflow_test.py
eunice-chan: If it's helpful, I can add the message above as a comment above the test function.

kenxu95: ah gotcha - yeah can we add a comment? It's not super clear that the name of the workflow is being randomized, which is different from the above test

integration_tests/sdk/delete_workflow_test.py
```diff
@@ -136,29 +139,48 @@ def test_delete_workflow_saved_objects_twice(client):
...
+        # Check delete_table is properly created at the integration.
+        integration.sql("SELECT * FROM delete_table").get()
 
         client.delete_flow(flow_1_id, saved_objects_to_delete=tables, force=True)
 
+        flow_ids_to_delete.remove(flow_1_id)
+
         # Wait for deletion to occur
-        sleep(1)
+        timeout = 500
```
kenxu95, line 159: nit: I think like 60 seconds is good enough. We also don't want to hang for too long and waste time :P

integration_tests/sdk/delete_workflow_test.py
kenxu95: Can you verify that the flow is indeed deleted? Maybe do a client.flow() and check that an InvalidUserArgumentException occurs.

integration_tests/sdk/delete_workflow_test.py
```diff
@@ -94,6 +95,8 @@ def test_delete_workflow_saved_objects(client):
 
         client.delete_flow(flow_id, saved_objects_to_delete=tables, force=True)
 
+        flow_ids_to_delete.remove(flow_id)
+
```
kenxu95, line 99: Can we also apply the same changes to `test_delete_workflow_saved_objects()`? The polling and use of .get(). The polling in particular seems useful as a helper in this file.

kenxu95: Also, if I'm understanding this test correctly, even if the `delete_flow()` partially succeeds by deleting all the metadata, you can still call it again to delete tables?

kenxu95 (changes requested):
Almost there! Just ping me when comments are addressed for final review :)

integration_tests/sdk/delete_workflow_test.py
eunice-chan: For this specific test I check the flow with object(s) saved with update_mode=APPEND can only be deleted if in force mode.

kenxu95 (approved):
There seems to be some leftover stuff here, maybe we should just delete the file and add it back later?

PR #365, opened 2022-08-16
Allows users to set S3 as storage layer when connecting integration

saurav-c:
This PR allows a user to set the S3 integration they are adding to be the new Aqueduct storage layer. This is only allowed for S3 at the moment. This is to make it easier for users to avoid having to provide the same AWS credentials in multiple places.

DEMO: https://www.loom.com/share/d3ec5d47a8994b71a68c04fed81fa6f5

ENG 1552

### DIFF
--- a/src/ui/common/src/components/integrations/dialogs/s3Dialog.tsx
+++ b/src/ui/common/src/components/integrations/dialogs/s3Dialog.tsx
@@ -22,7 +22,7 @@ const Placeholders: S3Config = {
   config_file_path: '',
   config_file_content: '',
   config_file_profile: '',
-  use_as_storage: false,
+  use_as_storage: '',
 };
 
 type Props = {
@@ -39,7 +39,7 @@ export const S3Dialog: React.FC<Props> = ({ setDialogConfig }) => {
   const [s3Type, setS3Type] = useState<S3CredentialType>(
     S3CredentialType.AccessKey
   );
-  const [useAsStorage, setUseAsStorage] = useState<boolean>(false);
+  const [useAsStorage, setUseAsStorage] = useState<string>('false');
 
   useEffect(() => {
     const config: S3Config = {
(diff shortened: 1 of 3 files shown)


src/ui/common/src/components/integrations/dialogs/s3Dialog.tsx
```diff
@@ -182,6 +201,11 @@ export const S3Dialog: React.FC<Props> = ({ setDialogConfig }) => {
       {s3Type === S3CredentialType.AccessKey && accessKeyTab}
       {s3Type === S3CredentialType.ConfigFilePath && configPathTab}
       {s3Type === S3CredentialType.ConfigFileContent && configUploadTab}
+
+      <FormControlLabel
+        label="Use this integration for Aqueduct storage."
+        control={<Checkbox checked={useAsStorage === 'true'} onChange={(event) => setUseAsStorage(event.target.checked ? 'true' : 'false')} />}
```
agiron123, line 207: is useAsStorage a string? Maybe we can do boolean instead?
Might also be able to do:
```checked={useAsStorage}```

Would need to update onChanged method as well to use a boolean.
```onChange={(event) => setUseAsStorage(event.target.checked ? true : false)}```

agiron123 (approved):
Looks good! Small change suggested, but you can go ahead and merge whenever you are ready.

cw75:
@saurav-c Can the user change back to local storage afterwards? Also, I assume since we're doing per-request context storage handle, there will never be cases where the storage config getting switched in the middle of a workflow execution right?

src/ui/common/src/components/integrations/dialogs/s3Dialog.tsx
saurav-c: Unfortunately we can't change this from a string to a boolean because of assumptions that our backend makes about what is stored in integration configs. The backend expects everything to be a string, and if this is a boolean it won't be JSON marshalled properly for the backend.

This is tracked in https://linear.app/aqueducthq/issue/ENG-1585/refactor-auth-pkg-to-allow-integration-credential-fields-to-be-non

saurav-c:
@cw75 Yes the user can switch back but not from the UI. They would have to use the CLI.

Yeah, that is correct, the storage layer will not change in the middle of workflow execution. This is also because once the storage config for a workflow has been written to the workflow dag table, that is the value that is used for all workflow execution.

vsreekanti:
This looks good @saurav-c! Two tiny nits:
* Let's add a placeholder for the region. 
* Let's change the checkbox to say "Aqueduct metadata storage" so it's clear to the user that this is a system internal thing.

saurav-c:
@vsreekanti Added

Screen Shot 2022-08-22 at 12 58 16 PM

PR #369, opened 2022-08-17
Fix sqlite transactional concurrency issue around SQLITE_BUSY and SQLITE_LOCKED

kenxu95:
It turns out the culprit to these "Unable to create workflow" transient errors boils down to the fact that we had concurrent write transactions that started with DEFERRED locking. With DEFERRED locking, the transaction only grabs locks when it has to, so when you have two concurrent transactions that both first grab read locks, and then upgrade to write locks, this upgrade fails with a SQLITE_BUSY error, and completely bypasses our `_busy_timeout` value.

The solution in the PR is to start everything as a write transaction, so any conflicting writes will wait for a `_busy_timeout` time before giving up. As long as `_busy_timeout` is set high enough, and our transactions are fast enough, this particular issue should no longer be a problem.

ENG-1468

### DIFF
--- a/src/golang/lib/database/sqlite.go
+++ b/src/golang/lib/database/sqlite.go
@@ -4,10 +4,10 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
-	stmt "github.com/aqueducthq/aqueduct/lib/database/stmt_preparers"
-	_ "github.com/mattn/go-sqlite3"
 	"os"
 	"path"
+
+	stmt "github.com/aqueducthq/aqueduct/lib/database/stmt_preparers"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


src/golang/lib/database/sqlite.go
```diff
@@ -21,10 +20,32 @@ const (
...
 
 var defaultSqliteOptions = map[string]string{
-	"mode":          "rwc",
-	"cache":         "shared",
-	"_journal_mode": "WAL",  // Enable Write-Ahead logging.
-	"_busy_timeout": "3000", // Wait for a bit on database locks before giving up.
+	"mode": "rwc",
+
+	// Write-Ahead Logging allows readers to continue to read even when a write transaction
+	// in progress. This does not resolve write-write conflicts.
```
HarisChoudhary, line 26: nit: "is" in progress

HarisChoudhary (approved):
lgtm! great explanation in code!

PR #371, opened 2022-08-17
Implements option to publish a k8s workflow in SDK

hsubbaraj-spiral:
Adds option to publish a k8s workflow in SDK 

```
from aqueduct import FlowConfig
k8s_integration = client.integration("k8s_integration")
no_input_flow_k8s = client.publish_flow(
    name = "No input k8s", 
    artifacts = [output],
    config = FlowConfig(engine=k8s_integration),
)
```
Eng-1555

### DIFF
--- a/sdk/aqueduct/__init__.py
+++ b/sdk/aqueduct/__init__.py
@@ -2,6 +2,7 @@ from typing import Any, List
 
 from aqueduct.aqueduct_client import Client, get_apikey, infer_requirements
 from aqueduct.constants import exports
+from aqueduct.dag import FlowConfig
 from aqueduct.decorator import check, metric, op, to_operator
 from aqueduct.enums import ArtifactType, CheckSeverity, LoadUpdateMode
 from aqueduct.flow import Flow
--- a/sdk/aqueduct/__init__.py
+++ b/sdk/aqueduct/__init__.py
@@ -1,8 +1,8 @@
 from typing import Any, List
 
 from aqueduct.aqueduct_client import Client, get_apikey, infer_requirements
+from aqueduct.config import FlowConfig
 from aqueduct.constants import exports
-from aqueduct.dag import FlowConfig
 from aqueduct.decorator import check, metric, op, to_operator
 from aqueduct.enums import ArtifactType, CheckSeverity, LoadUpdateMode
 from aqueduct.flow import Flow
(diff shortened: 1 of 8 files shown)


sdk/aqueduct/__init__.py
```diff
@@ -2,6 +2,7 @@
 
 from aqueduct.aqueduct_client import Client, get_apikey, infer_requirements
 from aqueduct.constants import exports
+from aqueduct.dag import FlowConfig
```
saurav-c, line 5: Why do we need the import here?

sdk/aqueduct/dag.py
```diff
@@ -50,10 +52,23 @@ class AirflowEngineConfig(BaseModel):
...
+
+
 class EngineConfig(BaseModel):
     type: RuntimeType = RuntimeType.AQUEDUCT
     aqueduct_config: Optional[AqueductEngineConfig]
     airflow_config: Optional[AirflowEngineConfig]
+    k8s_config: Optional[K8sEngineConfig]
+
+
+class FlowConfig(BaseModel):
```
saurav-c, line 66: I thought it made sense to leave this in `flow.py` since its technically the Flow config that a user specifies, whereas `Dag` and `EngineConfig` are more internal concepts that the user doesn't need to know about.

saurav-c (approved):
LGTM, just a couple minor nits

sdk/aqueduct/dag.py
```diff
@@ -50,10 +52,23 @@ class AirflowEngineConfig(BaseModel):
     integration_id: uuid.UUID
 
 
+class K8sEngineConfig(BaseModel):
+    integration_id: uuid.UUID
+
+
 class EngineConfig(BaseModel):
     type: RuntimeType = RuntimeType.AQUEDUCT
```
kenxu95, line 60: Could you add a quick note here that the runtime type dictates which engine config we expect to be set? The relationship between fields might be non-obvious to readers since we use different terms "runtime" and "engine" .

sdk/aqueduct/dag.py
kenxu95: I'm actually curious why these config classes are in the dag file to begin with? EngineConfigs seem like a separate concept from the dag, which is more structural info than runtime info.

kenxu95 (approved):
👍

vsreekanti:
API looks good to me!

PR #372, opened 2022-08-18
Double deletion 2

eunice-chan:
Add back in the two tests removed in https://github.com/aqueducthq/aqueduct/pull/363 since the SQLite concurrency issue is resolved.


integration_tests/sdk/delete_workflow_test.py
```diff
@@ -46,3 +46,135 @@ def test_delete_workflow_invalid_saved_objects(client):
...
+
+    finally:
+        for flow_id in flow_ids_to_delete:
+            delete_flow(client, flow_id)
+
+
+# Checking the successful deletion case and unsuccessful deletion case works as expected.
+# To test this, I have two workflows that write to the same table. When I delete the table in the first workflow,
+# it is successful but when I delete it in the second workflow, it is unsuccessful because the table has already
+# been deleted.
```
kenxu95, line 114: Thanks for the comments, super useful! I've been putting them as multi-line comments (""" """ format) inside the test functions, which I think is what the python style guide recommends.

integration_tests/sdk/delete_workflow_test.py
```diff
@@ -46,3 +46,135 @@ def test_delete_workflow_invalid_saved_objects(client):
...
+            client.delete_flow(flow_2_id, saved_objects_to_delete=tables, force=True)
+        assert str(e_info.value).startswith("Failed to delete")
+
+        # Failed to delete tables, but flow should be removed.
+        check_flow_doesnt_exist(client, flow_2_id)
+        flow_ids_to_delete.remove(flow_2_id)
+
+    finally:
+        for flow_id in flow_ids_to_delete:
+            delete_flow(client, flow_id)
```
kenxu95, line 180: Could you double check before merging that your flows are actually getting deleted? I remember seeing a lot of leftover flows when running this particular test file.

kenxu95 (approved):
Cool! Just a request to check that these flows are indeed getting deleted (also check stdout for any unexpected messages)

integration_tests/sdk/delete_workflow_test.py
eunice-chan: Ran the file in pytest and all the workflows got deleted afterward so I think we're good to go!

PR #376, opened 2022-08-19
Show other artifact types on the UI

cw75:
This PR enables showing other types of artifacts on the UI. We accomplish this by changing the `get_artifact_response` route to return a multipart form data that contains artifact result's metadata and data in bytes, and add logic on the client side (both sdk and UI) to correctly parse this response.

https://www.loom.com/share/cc1545f41eee45c7b67d8b71d318611b


src/golang/cmd/server/handler/get_artifact_result.go
```diff
@@ -40,16 +47,22 @@ type getArtifactResultArgs struct {
...
 	Name string `json:"name"`
 
 	// `Status` is redundant due to `ExecState`. Avoid consuming `Status` in new code.
 	// We are incurring this tech debt right now since there are quite a few usages of
 	// `status` in the UI.
-	Status    shared.ExecutionStatus `json:"status"`
-	ExecState shared.ExecutionState  `json:"exec_state"`
-	Schema    []map[string]string    `json:"schema"`
-	Data      string                 `json:"data"`
+	Status            shared.ExecutionStatus            `json:"status"`
```
agiron123, line 56: I'm currently doing some work on the UI for showing Artifact and Metric details in their own pages.

Perhaps we can take on this status refactor as part of my work as well?

agiron123: https://linear.app/aqueducthq/issue/ENG-1591/refactor-use-execstate-instead-of-status-on-ui

src/ui/common/src/components/layouts/data_previewer.tsx
```diff
@@ -74,9 +76,13 @@ const DataPreviewer: React.FC<Props> = ({ previewData, error }) => {
...
 
+  if (errorComponent) {
+    return <>{errorComponent}</>;
+  }
+
   let data: React.ReactElement;
-  if (previewData.result && previewData.result.data) {
-    if (previewData.result.schema.length > 0) {
+  if (previewData.result?.status === ExecutionStatus.Succeeded) {
+    if (previewData.result.serialization_type === SerializationType.Table) {
```
agiron123, line 85: This is probably a good use for a switch statement here.

src/ui/common/src/components/layouts/data_previewer.tsx
```diff
@@ -101,12 +107,45 @@ const DataPreviewer: React.FC<Props> = ({ previewData, error }) => {
...
+      previewData.result.serialization_type === SerializationType.Json
+    ) {
+      // Convert to pretty-printed version.
+      const prettyJson = JSON.stringify(
+        JSON.parse(previewData.result.data),
+        null,
+        2
+      );
+      data = (
+        <Typography sx={{ fontFamily: 'Monospace', whiteSpace: 'pre-wrap' }}>
```
agiron123, line 125: Might be a good idea to turn this Typography and the one down below on line 132 into their own components and pass in the data to be rendered there. From what I see, they pretty much are the same except for the content to be rendered.

agiron123 (approved):
Code looks good. Left a couple comments, but once they are resolved you can go ahead and merge.

cw75:
Addressed comments.

vsreekanti:
@cw75, this mostly looks great. A few tiny nits:
* Let's make the images left aligned instead of centered, since that's how we're showing other data as well. 
* For binary data: 
    * Let's not make the font teletype font; it's not code. 
    * Let's make the severity info instead of warning -- this isn't an issue we want them to be aware of, and nothing has gone wrong. 
* The icon for JSON/dictionary is a little confusing. It looks like a database or a server rack or something. Could we do a curly braces icon maybe? 
* The graph icon for numbers is also confusing now because a graph is an image type. Can we find a better icon here as well?

@agiron123, feel free to chime in with thoughts about icons.

cw75:
@vsreekanti curly braces icon is paid unfortunately. The current icon I have is a [list](https://fontawesome.com/icons/list?s=solid) icon. I'm not sure if there's a better alternative. Maybe [this](https://fontawesome.com/icons/list-ol?s=solid) one? Also, for numerics, should we use [this](https://fontawesome.com/icons/hashtag?s=solid)?

@agiron123 appreciate any tip for making the image left-aligned. Also what is "teletype font"? We've been using `Typography` from MUI but I don't see anywhere we specify the font type to be "teletype".

vsreekanti:
Sorry, by teletype font I meant monospace font. That font is typically associated with old-style telegraph-esque devices.

cw75:
@vsreekanti I see. If you look at the code, we are using the following components for all text block, but I don't think any of them involve code. So should we just remove `fontFamily: 'Monospace'`? And if we don't specify the font family, I assume it will use a default one which is fine? I can double check again.
```
const TextBlock: React.FC<Props> = ({ text }) => {
  return (
    <Typography sx={{ fontFamily: 'Monospace', whiteSpace: 'pre-wrap' }}>
      {text}
    </Typography>
  );
};
```

vsreekanti:
@cw75, I don't see why it's necessary to add that `TextBlock` component. We can just use the regular `Typography` component IMO and customize it as necessary.

cw75:
@vsreekanti fixed. Could you confirm the new icons are fine?
Screen Shot 2022-08-22 at 3 35 36 PM

vsreekanti:
The number icon is fine, but the list icon doesn't make sense for JSON -- JSON is definitionally unordered. We could either just do a Python object or try to think of something else.

PR #386, opened 2022-08-23
Clarifying comments in Python code

eunice-chan:
Add clarifying comments in Python code to changes done in part of the Integration Management project.

### DIFF
--- a/src/python/aqueduct_executor/operators/utils/utils.py
+++ b/src/python/aqueduct_executor/operators/utils/utils.py
@@ -261,6 +261,10 @@ def delete_object(name: str, delete_fn: Callable[[str], None]) -> SavedObjectDel
 def write_delete_saved_objects_results(
     storage: Storage, path: str, results: Dict[str, List[SavedObjectDelete]]
 ) -> None:
+    # `[object].json()`` gives me the json as a string which gets escaped in `json.dumps()``. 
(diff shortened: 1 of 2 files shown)


src/python/aqueduct_executor/operators/utils/utils.py
```diff
@@ -261,6 +261,10 @@ def delete_object(name: str, delete_fn: Callable[[str], None]) -> SavedObjectDel
 def write_delete_saved_objects_results(
     storage: Storage, path: str, results: Dict[str, List[SavedObjectDelete]]
 ) -> None:
+    # `[object].json()`` gives me the json as a string which gets escaped in `json.dumps()``.
```
likawind, line 264: should we use `"""` strings to be consistent with other changes?

likawind (approved):
LGTM!

src/python/aqueduct_executor/operators/utils/utils.py
eunice-chan: I used `"""` because I was explaining the whole function. Since this is explaining a line in the code, I just used `#` instead.
Looking at [this Python comment guide](https://www.askpython.com/python/python-comments), we can also comment with `'''` but we haven't done that in this codebase.

PR #392, opened 2022-08-24
Add support for lazy execution

cw75:
This PR adds support for lazy execution. We add `lazy` methods to all decorated functions, as well as an optional `lazy` argument to extract operators and artifact syntax sugar methods. For artifact generated from lazy execution, we set its type to `GenericArtifact`, with an empty `_content` field. Upon calling the `get()` method, its content field is then populated.

The PR includes more type-checking, but does not include the part where we check/backfill the type and content of  upstream lazy artifacts when previewing a downstream artifact. This will be in a follow-up PR.

The PR is ready for review, but needs relevant integration tests to be added before merging.


sdk/aqueduct/artifacts/generic_artifact.py
```diff
@@ -7,34 +7,36 @@
...
-        self._content = content
+        self.set_type(artifact_type)
+        self.set_content(content)
+
         if self._from_flow_run:
             # If the artifact is initialized from a flow run, then it should not contain any content.
-            assert self._content is None
-        else:
-            assert self._content is not None
+            assert self.content() is None
```
cw75, line 36: @HarisChoudhary I'm assuming you are fixing this as part of your parameter work, as we discussed that for artifact generated from flow run, its content should already be available via calling the backend's `get_artifact_result` endpoint. So the correct logic would be if `from_flow_run` is true, then content should never be `None`.

sdk/aqueduct/artifacts/generic_artifact.py
```diff
@@ -7,34 +7,36 @@
...
     """This class represents a generic artifact within the flow's DAG.
-    Currently, a generic artifact can be any artifact other than table, numeric, bool, or parameter.
+    Currently, a generic artifact can be any artifact other than table, numeric, bool, or parameter
+    generated from eager execution, or an artifact of unknown type generated from lazy execution.
     """
 
     def __init__(
         self,
         dag: DAG,
         artifact_id: uuid.UUID,
```
likawind, line 22: probably a separate thing - since it's already an `Artifact` object I'm curious if we should just use `id` and `type`

sdk/aqueduct/artifacts/numeric_artifact.py
```diff
@@ -126,6 +131,7 @@ def bound(
         equal: Optional[float] = None,
         notequal: Optional[float] = None,
         severity: CheckSeverity = CheckSeverity.WARNING,
+        lazy: bool = False,
```
likawind, line 134: should we use `execution_mode` here?

sdk/aqueduct/decorator.py
```diff
@@ -250,16 +281,27 @@ def wrapped(*input_artifacts: BaseArtifact) -> BaseArtifact:
...
             )
 
             return new_function_artifact
 
+        def wrapped(*input_artifacts: BaseArtifact) -> BaseArtifact:
+            return _wrapped_util(*input_artifacts, execution_mode=ExecutionMode.EAGER)
+
         # Enable the .local(*args) attribute, which calls the original function with the raw inputs.
-        def local_func(*inputs: Any) -> DataFrame:
+        def local_func(*inputs: Any) -> Any:
```
likawind, line 294: just for my own learning, what is `local_func` for? Also I slightly prefer `local` or `local_op` to make concepts consistent

sdk/aqueduct/decorator.py
```diff
@@ -473,24 +526,43 @@ def wrapped(
...
 
-            return new_check_artifact
+            return bool_artifact
+
+        """
+        @wraps(...): updates `wrapped()` to look like `func()` by copying attributes (eg. __name__, __doc__, etc.)
+        """
+
+        @wraps(func)
+        def wrapped(
```
likawind, line 549: we probably need to rethink the inheritance between these decorators given there are a number of similar patterns

sdk/aqueduct/integrations/s3_integration.py
```diff
@@ -118,18 +126,33 @@ def file(
...
+                raise InvalidArtifactTypeException(
+                    "The computed artifact is expected to be type %s, but has type %s"
+                    % (artifact_type, artifact.type())
+                )
 
-        return artifact
+            return artifact
+        else:
+            # We are in lazy mode.
+            if artifact_type == ArtifactType.TABLE:
```
likawind, line 148: Just for my own learning, why this checking is specific to s3? It appears to be something generalizable to all artifacts

sdk/aqueduct/decorator.py
likawind: Another question is - does these expose `.local` and `.lazy` for all wrapped artifacts or we only enable these method under specific modes? If it's the former case I feel it's hacky to use a `setattr` call. It seems that we could just expose `.local` and `.lazy` from `Artifact` objects?

likawind (commented):
Great! Mostly for my own learnings so asked some clarifying questions.

sdk/aqueduct/artifacts/generic_artifact.py
cw75: I used to have `id` and `type` and vscode warned me that these are Python keywords...

likawind: ahhh ouch, yeah this is quite brittle

sdk/aqueduct/artifacts/numeric_artifact.py
cw75: I also used to have `execution_mode ` here, but then I thought it may be easier for the users to type `lazy=True` instead of `execution_mode=ExecutionMode.LAZY`.

likawind: I see..does `ExecutionMode` get exposed to user in other places? If so I feel it might be fine to keep consistent and force user to always use `ExecutionMode` when needed, otherwise I agree the current way is better so that the user don't need to access `ExecutionMode` at all

likawind: Curious on how others think as well @kenxu95

sdk/aqueduct/decorator.py
cw75: `local_func` is used when the user just wants to use a decorated function as if it's a normal function. Maybe we can change the name to `local_mode`?

`.local` and `.lazy` aren't attached to an artifact; they are attached to a decorated function. So say we defined a function as:
```
@op 
def func(args):
  return "hello"
```

then `bar = func(foo)` executes in eager mode, `bar = func.lazy(foo)` executes in lazy mode.

likawind: Yeah both `.local` `.lazy` or `.local_mode` `.lazy_mode` makes sense, feels more consistent. Thanks for explaining how it's applied to operators, in this case I would try to make the `@op` wrapper returning an object with `__call__`, `.lazy()` and `.local` implemented so that it's more clear on what attributes are available. I don't think we need to do it in this PR but happy to provide more details as follow-up

sdk/aqueduct/integrations/s3_integration.py
cw75: For S3, we require the user to provide a type hint in the `file` API, so even in lazy mode we can try to use the type hint to create type-specific artifact classes. In SQL, since the type has to be Table, in lazy mode we can also "hardcode" the returned type to be Table, but we don't need to worry about cases like numeric, bool, etc. For generic operators (`@op`), since the output type can be anything and there's no hint from the user, we can only return a `GenericArtifact` with its type being `UNTYPED`.

sdk/aqueduct/artifacts/numeric_artifact.py
cw75: `ExecutionMode` isn't exposed to the user elsewhere, so I think `lazy=True` is probably easier.

likawind: yeah agreed, this way user don't  need another import and code looks cleaner

cw75:
@likawind could you take a quick look at the new tests added? If they make sense you can go ahead and stamp this.

likawind (approved):
Everything LGTM! Thanks for adding all tests

PR #402, opened 2022-08-26
Consolidates the notifications panes into a single box

saurav-c:
This PR consolidates the notification panes into a single box since we no longer have a distinction between org notifications and user notifications.

ENG 1610

Demo: https://www.loom.com/share/849d3456fce0434d813dcda9b716c757

### DIFF
--- a/src/ui/common/src/components/notifications/NotificationsPopover.tsx
+++ b/src/ui/common/src/components/notifications/NotificationsPopover.tsx
@@ -2,18 +2,14 @@ import { faBoxArchive, faInbox } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { List, Popover, Typography } from '@mui/material';
 import Box from '@mui/material/Box';
-import React, { useState } from 'react';
+import React from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 
 import { handleArchiveAllNotifications } from '../../reducers/notifications';
 import { AppDispatch, RootState } from '../../stores/store';
 import UserProfile from '../../utils/auth';
-import {
-  NotificationAssociation,
-  NotificationLogLevel,
-} from '../../utils/notifications';
+import { NotificationLogLevel } from '../../utils/notifications';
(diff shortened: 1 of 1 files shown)


src/ui/common/src/components/notifications/NotificationsPopover.tsx
```diff
@@ -40,35 +31,16 @@ export const NotificationsPopover: React.FC<NotificationsPopoverProps> = ({
...
-        return true;
-      } else if (activeTab === NotificationsTabs.Workflow) {
-        return (
-          association === NotificationAssociation.Workflow ||
-          association === NotificationAssociation.WorkflowDagResult
-        );
-      }
-
-      return true;
+      return notification.level !== NotificationLogLevel.Success;
```
saurav-c, line 43: @vsreekanti This logic was there from before where we only show non-success notifications. Wanted to confirm that this what we want?

agiron123:
So this works for the case of our OSS edition. How do we plan on having separate code for the Enterprise edition? Or is that not a priority for now? 
cc @vsreekanti

src/ui/common/src/components/notifications/NotificationsPopover.tsx
agiron123: Showing success notifications turned out to be a little too noisy, particularly when users had scheduled workflows, so I believe this is what we want.

agiron123 (approved):

src/ui/common/src/components/notifications/NotificationsPopover.tsx
vsreekanti: Yep, I think that's good for now. We can maybe make this an app setting in the future.

vsreekanti:
Good question @agiron123. I'm not super worried about it for the time being. I think in the future, we may want to separate out notifications and events in some form, as we've discussed in the past, so I think this is something that we'll revisit sooner rather than later. For the time being, I think simplifying this view is a win though.

agiron123:
@vsreekanti Sounds good, let's merge as is then. :shipit:

PR #403, opened 2022-08-26
Refactors execute workflow to improve persisting DAG status

saurav-c:
This PR refactors the `Engine.ExecuteWorkflow` implementation to capture as many errors as possible when persisting the DAG status to the WorkflowDagResult table.

Essentially, the more code we have BEFORE we initialize a new row in the WorkflowDagResult table, the more errors we will be unable to capture and update the WorkflowDagResult row accordingly. In order to do this, we want to minimize the code inside of `ExecuteWorkflow` BEFORE the WorkflowDagResult row is created, because as soon as this record is created, we can add a `deferred` function that captures the state of the DAG run and updates the WorkflowDagResult record accordingly.

This PR gets rid of the `PersistResults` method from the `WorkflowDag` interface, because this logic is now handled directly by the `ExecuteWorkflow` implementation. Similarly, `InitializeResults` has been refactored to only deal with initializing Artifact and Operator results.

ENG 1608

Test:
I modified the `Execute` function to always throw an error when the JobManager is created [here](https://github.com/aqueducthq/aqueduct/blob/main/src/golang/lib/engine/aq_engine.go#L225). Before this change the new workflow dag result would not be created and the user wouldn't see any new runs on the UI.

AFTER these changes, the user will now see the following on the UI. The run is marked as failed. 
Screen Shot 2022-08-26 at 1 49 47 PM

__EDIT: These issues are solved by #404__
~~There are still some things that should be improved from an error handling standpoint:~~
1. We have no way of letting the user know why their workflow failed, because errors are tied only to operator results. We don't have any fields that track an error that occurred before any operators even ran. An example of this is a system internal error, such as an error when setting up the job manager. Kind of solved by #404 (We show this as a system internal error)
2. The UI shows all of the operators as in progress even though the workflow run has failed. SOLVED by #404 
3. The UI crashes if a user clicks into an artifact or operator result that does not exist. (This is very high priority and the relevant task has been created. SOLVED by https://github.com/aqueducthq/aqueduct/pull/404

### DIFF
--- a/src/golang/lib/engine/aq_engine.go
+++ b/src/golang/lib/engine/aq_engine.go
@@ -200,6 +200,11 @@ func (eng *aqEngine) ExecuteWorkflow(
 
 	// Any errors after this point should be persisted to the WorkflowDagResult created above
 	defer func() {
+		if err != nil {
+			// Mark the workflow dag result as failed
+			status = shared.FailedExecutionStatus
+		}
+
 		workflow_utils.UpdateWorkflowDagResultMetadata(
 			ctx,
 			dbWorkflowDagResult.Id,
--- a/src/golang/lib/engine/aq_engine.go
+++ b/src/golang/lib/engine/aq_engine.go
@@ -186,6 +186,32 @@ func (eng *aqEngine) ExecuteWorkflow(
 		return shared.FailedExecutionStatus, errors.Wrap(err, "Error reading latest workflowDag.")
 	}
 
+	dbWorkflowDagResult, err := workflow_utils.CreateWorkflowDagResult(
+		ctx,
+		dbWorkflowDag.Id,
+		eng.WorkflowDagResultWriter,
+		eng.Database,
+	)
+	if err != nil {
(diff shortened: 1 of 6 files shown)


hsubbaraj-spiral (approved):
Yep, this is great!

src/golang/lib/engine/aq_engine.go
```diff
@@ -186,6 +186,37 @@ func (eng *aqEngine) ExecuteWorkflow(
 		return shared.FailedExecutionStatus, errors.Wrap(err, "Error reading latest workflowDag.")
```
likawind, line 186: just to confirm, this method will be called in `main.go` or `executor` binary, and if something's wrong when spinning up the executor, we will still get an untrackable failure? And I believe we would need go-routine based scheduling to solve this, which is def. out-of-scope for this PR.

saurav-c: Right, this decreases the amount of code that would result in an untrackable failure, but we could definitely do more as you said.

likawind (approved):

PR #406, opened 2022-08-29
Use the paths stored in the preview cache instead of copying the data over

kenxu95:
This should help the performance of the cache, especially when we start using slower storage like S3.

In order to accomplish this, I made the following two changes:
1) We check that the input paths exist before proceeding, throwing an appropriate error message. Ideally, use a different failure type like TRANSIENT which means the orchestrator can retry this issue. However, I don't expect this to be an issue since a cache hit should not be evicted that quickly.
2) We do not update the cache if the new entry is exactly the same as the existing one.

ENG-1582

### DIFF
--- a/src/golang/lib/engine/engine.go
+++ b/src/golang/lib/engine/engine.go
@@ -15,7 +15,7 @@ import (
 const (
 	DefaultExecutionTimeout     = 15 * time.Minute
 	DefaultCleanupTimeout       = 2 * time.Minute
-	DefaultPollIntervalMillisec = 100
+	DefaultPollIntervalMillisec = 300
 )
 
 var (
--- a/src/golang/lib/workflow/operator/base.go
+++ b/src/golang/lib/workflow/operator/base.go
@@ -75,30 +75,6 @@ func unknownSystemFailureExecState(err error, logMsg string) *shared.ExecutionSt
 	}
 }
 
-// For each output artifact, copy over the contents of the content and metadata paths.
-// This should only ever be used for preview routes. Returns an error if this operator did not succeed.
-// If it does not, the operator will fall back on traditional execution, overwriting anything we wrote
-// to the output paths here.
-func (bo *baseOperator) executeUsingCachedResult(ctx context.Context, allCachedOutputPaths []preview_cache.Entry) error {
-	for i, cachedOutputPaths := range allCachedOutputPaths {
-		err := utils.CopyPathContentsInStorage(ctx, bo.storageConfig, cachedOutputPaths.ArtifactContentPath, bo.outputExecPaths[i].ArtifactContentPath)
-		if err != nil {
-			return err
-		}
-
-		err = utils.CopyPathContentsInStorage(ctx, bo.storageConfig, cachedOutputPaths.ArtifactMetadataPath, bo.outputExecPaths[i].ArtifactMetadataPath)
-		if err != nil {
-			return err
-		}
-
-		err = utils.CopyPathContentsInStorage(ctx, bo.storageConfig, cachedOutputPaths.OpMetadataPath, bo.outputExecPaths[i].OpMetadataPath)
-		if err != nil {
-			return err
-		}
-	}
-	return nil
-}
-
 func (bo *baseOperator) launch(ctx context.Context, spec job.Spec) error {
 	// Check if this operator can use previously cached results instead of computing for scratch.
 	if bo.previewCacheManager != nil {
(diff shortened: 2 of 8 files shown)


src/golang/lib/engine/engine.go
```diff
@@ -15,7 +15,7 @@ import (
 const (
 	DefaultExecutionTimeout     = 15 * time.Minute
 	DefaultCleanupTimeout       = 2 * time.Minute
-	DefaultPollIntervalMillisec = 100
+	DefaultPollIntervalMillisec = 300
```
cw75, line 18: Is this change intentional?

src/golang/lib/workflow/operator/base.go
```diff
@@ -113,18 +89,18 @@ func (bo *baseOperator) launch(ctx context.Context, spec job.Spec) error {
...
-				// We've successfully used the cached result!
-				return nil
+			// Apply the cached results for each output artifact. This just means setting the output paths
+			// to be the same as the cached ones.
+			for i, outputArtifact := range bo.outputs {
+				cacheEntry := cacheEntryByKey[outputArtifact.Signature()]
+				bo.outputExecPaths[i].ArtifactMetadataPath = cacheEntry.ArtifactMetadataPath
+				bo.outputExecPaths[i].ArtifactContentPath = cacheEntry.ArtifactContentPath
+
+				// The op metadata path is updated for both for this operator and its output artifacts.
```
cw75, line 99: typo: "for both this operator and its output artifacts."

src/golang/lib/workflow/preview_cache/preview_cache.go
```diff
@@ -82,19 +77,53 @@ func (c *inMemoryPreviewCacheManagerImpl) Put(ctx context.Context, artifactSigna
...
 			ArtifactContentPath:  execPathsList[i].ArtifactContentPath,
 			ArtifactMetadataPath: execPathsList[i].ArtifactMetadataPath,
 			OpMetadataPath:       execPathsList[i].OpMetadataPath,
 		})
+
+		// After adding the new value to the cache, delete the orphaned data if this overwrote an existing
+		// entry. This is best-effort.
+		if existingEntry != nil {
+			deleteDataForEntry(ctx, c.storageConfig, *existingEntry)
+		}
```
cw75, line 126: Will we ever reach here?

src/golang/lib/workflow/preview_cache/preview_cache.go
```diff
@@ -82,19 +77,53 @@ func (c *inMemoryPreviewCacheManagerImpl) Put(ctx context.Context, artifactSigna
...
+		val, exists := c.cache.Peek(signatures)
+		if exists {
+			existingEntry = castCachedValueToEntry(val)
+			if existingEntry == nil {
+				return errors.New("Preview Artifact Cache is storing an unexpected data structure.")
+			}
+
+			// If the entry already exists and has the same data, short-circuit as there is no need
+			// to update anything.
+			if isEqual(execPathsList[i], existingEntry) {
```
cw75, line 111: If there is a cache hit, since we aren't copying these paths any more, isn't `execPathsList[i]` and `existingEntry` guaranteed to be equal? If so, I think we should do an assertion instead?

cw75 (commented):
Looks good overall! Left a few comments inline. I'm fine with shipping this PR, but let's create a task that handles the retry part if the data read failed due to an unlikely event of cache eviction.

src/golang/lib/engine/engine.go
kenxu95: yeah, I noticed during testing that we were polling super aggressively so I wanted to back that off a little bit.

src/golang/lib/workflow/preview_cache/preview_cache.go
kenxu95: You are right that if there is a cache hit, these are guaranteed to be equal. However, this is the in the Put() part of the cache, which is called by artifact.Finish(), which has no idea whether the artifact was generated from a cache hit or not. 

We could move this logic into artifact, and only Put() if the artifact was generated from a cache hit, but I feel it belongs better at the cache layer.

src/golang/lib/workflow/preview_cache/preview_cache.go
kenxu95: I don't think it will. Can drop a note here. But I do think it's totally reasonable to have this logic, since it means the cache doesn't rely on any assumptions about the cache user. Eg. if we change the cache entry to be mutable in the future, we won't have to remember to include this logic, since it's already here.

src/golang/lib/workflow/preview_cache/preview_cache.go
cw75: I agree it should belong in the caching layer, but since this is already within the `if exists` block, which means we are in the cache hit case, I think we should change this to an assertion?

src/golang/lib/workflow/preview_cache/preview_cache.go
cw75: Yeah let's drop a note here.

src/golang/lib/workflow/preview_cache/preview_cache.go
kenxu95: gotcha, that makes sense. I'll throw an error since Go doesn't have assertions.

src/golang/lib/workflow/preview_cache/preview_cache.go
kenxu95: actually on second thought, it looks better without this logic now that we error out if the entries are not equal. So its just less confusing if we remove it.

cw75 (approved):

PR #407, opened 2022-08-29
Makes exported function code more user-friendly to read

saurav-c:
This PR refactors how we capture and return exported functions to the user.
The code is captured when a workflow is created by adding a special read-only file called `source.py`. This file isn't used as part of any operator execution, but it is just there as a user-friendly copy of the code that has been serialized into `model.pkl`.

ENG 1190

UPDATE:
The files now look like
Screen Shot 2022-08-29 at 4 48 40 PM

Demo: https://www.loom.com/share/a553c71f4d9b4216a7c2b22db30073ad

### DIFF
--- a/sdk/aqueduct/utils.py
+++ b/sdk/aqueduct/utils.py
@@ -13,7 +13,6 @@ from typing import Any, Callable, Dict, List, Mapping, Optional, Union
 import cloudpickle as pickle
 import multipart
 import numpy as np
-import pandas as pd
 import requests
 from aqueduct.config import AirflowEngineConfig, EngineConfig, FlowConfig, K8sEngineConfig
 from aqueduct.dag import DAG, RetentionPolicy, Schedule
@@ -110,11 +109,9 @@ def retention_policy_from_latest_runs(k_latest_runs: int) -> RetentionPolicy:
 
 MODEL_FILE_NAME = "model.py"
 MODEL_PICKLE_FILE_NAME = "model.pkl"
-FUNCTION_SOURCE_FILE_NAME = "source.py"
 PYTHON_VERSION_FILE_NAME = "python_version.txt"
 CONDA_VERSION_FILE_NAME = "conda_version.txt"
 RESERVED_FILE_NAMES = [
-    FUNCTION_SOURCE_FILE_NAME,
     MODEL_FILE_NAME,
     MODEL_PICKLE_FILE_NAME,
     PYTHON_VERSION_FILE_NAME,
@@ -160,6 +157,7 @@ def make_zip_dir() -> str:
 
 def serialize_function(
     func: Union[UserFunction, MetricFunction, CheckFunction],
+    op_name: str,
     file_dependencies: Optional[List[str]] = None,
     requirements: Optional[Union[str, List[str]]] = None,
 ) -> bytes:
@@ -169,6 +167,8 @@ def serialize_function(
     Arguments:
         func:
             The function to package
+        op_name:
+            The name of the function operator to package.    
         file_dependencies:
             A list of relative paths to files that the function needs to access.
         requirements:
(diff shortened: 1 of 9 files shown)


sdk/aqueduct/utils.py
```diff
@@ -110,9 +110,11 @@ def retention_policy_from_latest_runs(k_latest_runs: int) -> RetentionPolicy:
 
 MODEL_FILE_NAME = "model.py"
 MODEL_PICKLE_FILE_NAME = "model.pkl"
+FUNCTION_SOURCE_FILE_NAME = "source.py"
```
saurav-c, line 113: @vsreekanti Is it too restrictive to make `source.py` a reserved file?

vsreekanti: I would prefer not to. Would it be possible to make the filename in the zipfile the name of the function?

saurav-c: Sure, we'll just have to prevent users from uploading file dependencies with the same name as the function.

vsreekanti: Yep, I think that's fine.

saurav-c: Fixed, the files now look like:
Screen Shot 2022-08-29 at 4 48 40 PM

vsreekanti: Yep, that's perfect.

sdk/aqueduct/utils.py
```diff
@@ -197,6 +199,13 @@ def serialize_function(
         with open(os.path.join(dir_path, MODEL_PICKLE_FILE_NAME), "wb") as f:
             pickle.dump(func, f)
 
+        # Write function source code to file
+        source_file = "{}.py".format(op_name)
+        with open(os.path.join(dir_path, source_file), "w") as f:
+            source = inspect.getsource(func)
+            print(source)
```
kenxu95, line 206: is this print supposed to be here? If it is, was should use the logger.

src/golang/cmd/server/handler/export_function.go
```diff
@@ -125,13 +135,92 @@ func (h *ExportFunctionHandler) Perform(ctx context.Context, interfaceArgs inter
...
+		// than the current directory name that is a unique UUID
+		parts[0] = operatorName
+		zipFileName := strings.Join(parts, "/")
+
+		if err := writeZipFile(zipWriter, zipFile, zipFileName); err != nil {
+			return nil, err
+		}
+	}
+
+	if err := zipWriter.Close(); err != nil {
```
kenxu95, line 197: I'm not familiar with the zipWriter interface, is there a reason we don't `defer` this after the zipWriter creation?

sdk/aqueduct/utils.py
```diff
@@ -197,6 +199,13 @@ def serialize_function(
         with open(os.path.join(dir_path, MODEL_PICKLE_FILE_NAME), "wb") as f:
             pickle.dump(func, f)
 
+        # Write function source code to file
+        source_file = "{}.py".format(op_name)
+        with open(os.path.join(dir_path, source_file), "w") as f:
+            source = inspect.getsource(func)
```
kenxu95, line 205: Based on the loom vid, this won't capture the import statements right? Should that be a nice-to-have follow up (is it even possible)

kenxu95 (approved):
nice!

src/golang/cmd/server/handler/export_function.go
saurav-c: We have to explicitly close the `zipWriter` before we can read its content: https://stackoverflow.com/a/55623815

sdk/aqueduct/utils.py
saurav-c: Removed

sdk/aqueduct/utils.py
saurav-c: Yup that's right. I don't think it's easily possible and not sure if its worth the engineering effort cc @vsreekanti

PR #410, opened 2022-08-30
[BE/3] ENG-1509 Add flag to Update Mode indicating which is currently used

eunice-chan:
## Describe your changes and why you are making these changes
Add `created_at` as a field to the workflow objects list so we can later signify to the user which update mode is the active one (since the active workflow is currently the latest workflow).

## Related issue number (if any)
ENG-1509
BE: https://github.com/aqueducthq/aqueduct/pull/410
SDK: https://github.com/aqueducthq/aqueduct/pull/411
UI: https://github.com/aqueducthq/aqueduct/pull/412

## Checklist before requesting a review
- [x] I have created a descriptive PR title. The PR title should complete the sentence "This PR...".
- [x] I have performed a self-review of my code.
- [N/A] I have included a small demo of the changes. For the UI, this would be a screenshot or a Loom video.
- [N/A] If this is a new feature, I have added unit tests and integration tests.
- [N/A] I have run the integration tests locally and they are passing.
- [x] I have run the linter script locally (See `python3 scripts/run_linters.py -h` for usage).
- [x] All features on the UI continue to work correctly.
- [x] Added one of the following CI labels:
    - `run_integration_test`: Runs integration tests
    - `skip_integration_test`: Skips integration tests (Should be used when changes are ONLY documentation/UI)

PR #412, opened 2022-08-30
[UI/3] ENG-1509 Add flag to Update Mode indicating which is currently used

eunice-chan:
Signify to the user which update mode is the active one (since the active workflow is currently the latest workflow).

update_mode_active

ENG-1509
BE: https://github.com/aqueducthq/aqueduct/pull/410
SDK: https://github.com/aqueducthq/aqueduct/pull/411
UI: https://github.com/aqueducthq/aqueduct/pull/412

### DIFF
--- a/src/ui/common/src/components/workflows/WorkflowSettings.tsx
+++ b/src/ui/common/src/components/workflows/WorkflowSettings.tsx
@@ -432,48 +432,54 @@ const WorkflowSettings: React.FC<WorkflowSettingsProps> = ({
   const listSavedObjects = (
     <FormGroup>
       {Object.entries(savedObjects).map(
-        ([integrationTableKey, savedObjectsList]) => (
-          <FormControlLabel
-            key={integrationTableKey}
-            control={
-              <Checkbox
-                id={integrationTableKey}
-                onChange={updateSelectedObjects}
-              />
-            }
-            label={
-              <Box>
-                {displayObject(
-                  savedObjectsList[0].integration_name,
-                  savedObjectsList[0].object_name
-                )}
+        ([integrationTableKey, savedObjectsList]) => {
+          const sortedObjects = [...savedObjectsList].sort(
+            (object) => object.created_at
+          );
+          return (
+            <FormControlLabel
+              key={integrationTableKey}
+              control={
+                <Checkbox
+                  id={integrationTableKey}
+                  onChange={updateSelectedObjects}
+                />
+              }
+              label={
+                <Box>
+                  {displayObject(
+                    savedObjectsList[0].integration_name,
+                    savedObjectsList[0].object_name
+                  )}
 
-                <Typography
-                  style={{
-                    color: theme.palette.gray[600],
-                    paddingRight: '8px',
-                  }}
-                  variant="body2"
-                  display="inline"
-                >
-                  Update Mode:{' '}
-                  {savedObjectsList
-                    .map((object) => object.update_mode)
-                    .join(', ')}
-                </Typography>
-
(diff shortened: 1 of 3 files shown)


vsreekanti:
@eunice-chan, this looks great! Two tiny things:
1. Showing the tooltip next to every table is a little confusing; it seems to imply that you're going to say something table-specific, but looking at the code, it looks like every tooltip will just say "Multiple updates modes have been associated with this table." That doesn't actually seem true since some of the tables only have one update mode. Maybe we could include this information somewhere else? 
2. Can we align the checkbox with the table name rather than the middle of the `Box` with the text?

eunice-chan:
@vsreekanti Thanks for your attention to detail!
> Showing the tooltip next to every table is a little confusing; it seems to imply that you're going to say something table-specific, but looking at the code, it looks like every tooltip will just say "Multiple updates modes have been associated with this table." That doesn't actually seem true since some of the tables only have one update mode. Maybe we could include this information somewhere else?

Do you think it'll be better to just have the tooltip next to the tables with multiple update modes? Or should we remove it completely? Since now we mark the active update mode, it might not really be necessary anymore.

> Can we align the checkbox with the table name rather than the middle of the Box with the text?

Yes, I can do that!

vsreekanti:
I think you're right that the tooltip is probably not necessary now. Let's remove it for the time being, and we can always add it back if it is something users are confused by in the future.

eunice-chan:
Remove tooltip & align checkbox with the start of the line (@agiron123 I think the way I aligned it is kind of bad but I can't find a cleaner way that works, so would appreciate any ideas you have)
image

src/ui/common/src/components/workflows/WorkflowSettings.tsx
```diff
@@ -446,36 +463,13 @@ const WorkflowSettings: React.FC<WorkflowSettingsProps> = ({
                 />
               }
               label={
-                <Box>
+                <div style={{ marginTop: '1.2em' }}>
```
agiron123, line 466: Can you use the box component here please?

We're also not really using units like em at the moment. Our design system is based on units of 8px. 

See if you can line things up with a box and the sx prop and a marginTop of x * 8px, where x is whatever looks aligned..

src/ui/common/src/components/workflows/WorkflowSettings.tsx
```diff
@@ -436,8 +450,11 @@ const WorkflowSettings: React.FC<WorkflowSettingsProps> = ({
           const sortedObjects = [...savedObjectsList].sort(
             (object) => object.created_at
           );
+          // Cannot align the checkbox to the top of a multi-line label.
+          // Using a weird marginTop workaround.
           return (
             <FormControlLabel
+              style={{ marginTop: '-1.2em' }}
```
agiron123, line 457: Same as below, let's try using units of 8px here.

agiron123:
For alignment here, i'd try wrapping each of the lines in an MUI box component and set the paddingTop for each label.

You can see an example of this here:
```
      <FormControlLabel
          value="start"
          control={<Checkbox />}
          label={<Box sx={{ paddingTop: '24px' }}><Typography variant="body1">Line1 Text</Typography><Typography variant="body1">Line2Text</Typography></Box>}
          labelPlacement="start"
        />
```

Code Sandbox to play around with:
https://codesandbox.io/s/multi-line-labels-fm4sd0?file=/index.tsx

agiron123 (changes requested):
Couple small comments, but should be easy to resolve.

eunice-chan:
Updated `div` -> `box`; `1.2em` -> `24px`. Visually looks about the same.

image

agiron123 (approved):
Nice work!

PR #420, opened 2022-09-01
Fix integration test created_at issue

eunice-chan:
- Update `created_at` --> `modified_at`; string of datetime.
- Fix SQL query.
Passed integration tests: https://github.com/aqueducthq/aqueduct/actions/runs/2968194809


likawind (changes requested):
Thanks for fixing! Could you check gh action failures in this PR?

likawind (approved):

PR #421, opened 2022-09-01
Implements backfilling of lazily realized artifacts

kenxu95:
In order to implement this, I pushed all the type-checking and type-annotating logic into `preview_artifact()`. The reason this works is because type information is always stored in the underlying shared dag, and is never annotated on the artifact itself.

Content is still annotated on the artifact itself, so you can't backfill contents. Moreover, you cannot ever have content != None and type == UNTYPED. This ensures that we can backfill the type freely, without having to make sure the content is compatible (because when you backfill, content is expected to be None).

ENG-1338


sdk/aqueduct/artifacts/bool_artifact.py
```diff
@@ -70,23 +68,14 @@ def get(self, parameters: Optional[Dict[str, Any]] = None) -> Union[bool, np.boo
...
-        if parameters:
-            return previewed_artifact._get_content()
-        else:
-            # We are materializing an artifact generated from lazy execution.
-            assert self._get_content() is None
-            self._set_content(previewed_artifact._get_content())
-            return self._get_content()
+        # If the artifact was previously generated lazily, materialize the contents.
+        if self._get_content() is None:
+            self._set_content(content)
```
cw75, line 76: I think we should set the content only when `parameter == None`. Otherwise when the user previews the artifact again without the parameter, we will just return the version that's evaluated with the custom parameter, which is incorrect.

sdk/aqueduct/artifacts/generic_artifact.py
```diff
@@ -24,20 +23,19 @@ def __init__(
         content: Optional[Any] = None,
         from_flow_run: bool = False,
     ):
+        if content is not None:
+            assert artifact_type != ArtifactType.UNTYPED
+
```
cw75, line 28: Is `artifact_type` only needed for this assertion purpose?

sdk/aqueduct/artifacts/generic_artifact.py
```diff
@@ -58,23 +56,13 @@ def get(self, parameters: Optional[Dict[str, Any]] = None) -> Any:
...
 
-        if parameters:
-            return previewed_artifact._get_content()
-        else:
-            # We are materializing an artifact generated from lazy execution.
-            assert self._get_content() is None
-            self._set_content(previewed_artifact._get_content())
-            self._set_type(previewed_artifact._get_type())
-            return self._get_content()
+        # If the artifact was previously generated lazily, materialize the contents.
```
cw75, line 61: same comment about not setting the content when parameter is not None.

sdk/aqueduct/artifacts/numeric_artifact.py
```diff
@@ -93,25 +91,16 @@ def get(self, parameters: Optional[Dict[str, Any]] = None) -> Union[int, float,
...
         )
 
-        if parameters:
-            return previewed_artifact._get_content()
-        else:
-            # We are materializing an artifact generated from lazy execution.
-            assert self._get_content() is None
-            self._set_content(previewed_artifact._get_content())
-            return self._get_content()
+        if self._get_content() is None:
```
cw75, line 100: same comment about not setting the content when parameter is not None.

sdk/aqueduct/artifacts/table_artifact.py
```diff
@@ -123,21 +121,14 @@ def get(self, parameters: Optional[Dict[str, Any]] = None) -> pd.DataFrame:
...
-            )
+        content = previewed_artifact._get_content()
+        assert isinstance(content, pd.DataFrame)
 
-        assert isinstance(previewed_artifact._get_content(), pd.DataFrame)
+        # If the artifact was previously generated lazily, materialize the contents.
+        if self._get_content() is None:
+            self._set_content(content)
 
-        if parameters:
```
cw75, line 134: same comment about not setting the content when parameter is not None.

sdk/aqueduct/artifacts/utils.py
```diff
@@ -39,24 +44,53 @@ def preview_artifact(
...
+def _get_content_from_artifact_result_resp(artifact_result: ArtifactResult) -> Any:
+    """Deserialize and validate the type of the content for a given artifact result."""
+    serialization_type = artifact_result.serialization_type
     if serialization_type not in deserialization_function_mapping:
         raise Exception("Unsupported serialization type %s." % serialization_type)
 
-    artifact_type = artifact_response.artifact_type
     content = deserialization_function_mapping[serialization_type](
-        base64.b64decode(artifact_response.content)
+        base64.b64decode(artifact_result.content)
```
cw75, line 93: reminder to resolve conflict here since we merged in the base64 removal PR.

sdk/aqueduct/artifacts/utils.py
```diff
@@ -39,24 +44,53 @@ def preview_artifact(
...
+
+    # Any non-target artifacts are guaranteed to be upstream of the target artifact (due to the SubgraphDAGDelta),
+    # so if any of them are the result of a lazy operation, we'll want to backfill their types. *We do NOT backfill
+    # their contents*, as those are stored on the Artifact class itself and not the underlying shared dag.
+    for artifact_id, artifact_result in preview_resp.artifact_results.items():
+        # We've already processed the target artifact.
+        if artifact_id == target_artifact_id:
+            continue
+
+        dag.update_artifact_type(artifact_id, artifact_result.artifact_type)
```
cw75, line 72: If an upstream artifact already has type set and there is a mismatch, I think we should also raise exception as in line 58?

cw75 (changes requested):
@kenxu95 LGTM overall. Left a few comments inline to address. Maybe also add a test that catches type mismatch error during the backfilling process?

sdk/aqueduct/artifacts/generic_artifact.py
kenxu95: yep I think so. Added a comment to make it more clear.

sdk/aqueduct/artifacts/utils.py
kenxu95: The only way this is possible if there is a preview cache miss on the upstream right? It's a little challenging to test this without an `ignore_cache` flag.

cw75: @kenxu95 I think it might be possible to test if an artifact upstream involves a custom parameter?

cw75 (approved):
LGTM! I responded to the upstream backfill type error testing; I think it might be possible to test with an upstream artifact that involves custom parameter values? Then it won't be treated as a cache hit.

sdk/aqueduct/artifacts/utils.py
kenxu95: oh right I forgot parameter values are computed in the signature. Nice.

kenxu95:
The issue is that I can't ever get it to trigger the error on the sdk side, because the backend also does the type enforcement. So if an upstream operator returns an unexpected type, the orchestrator will throw a type output mismatch error and will not proceed further, so the SDK error is more of a defensive check. I've added a test for this.

cw75:
@kenxu95 But the backend's type enforcement only happens after workflow publish right? So I think the SDK side's error checking can still trigger during the preview phase?

kenxu95:
That is not how it's currently written. Do you want to change only type-enforce only in publish mode? I'm not sure I get how enforcing it on the backend is different from enforcing it on the SDK. The types are encoded in the SDK dag anyways.

cw75:
@kenxu95 wait, since we only persist data after publishing a workflow, how can backend enforce type consistency pre-publish?

kenxu95:
In previews, the type of the artifact on the uploaded DAG (which is maintained on the SDK side) is used as the expected artifact type.

cw75:
@kenxu95 Oh I see. Didn't realize that. :-) I think that's fine then. We should be good to go.

PR #423, opened 2022-09-01
Enforce typing constraints for save()

kenxu95:
Provides the following guiderails:
- Tabular data can be saved in S3, but S3FileFormat needs to be provided (this was enforced by the backend but I don't see why we can't do this earlier in the SDK)
- Non-Tabular data cannot be saved into a relational data store.

Our testing around S3 is non-existent, I found at least one bug that was fixed in this PR. I really think we should invest in automating this in the near future.

```
from aqueduct import op, Client

client = Client(api_key="AZ2U7TQ4Y8X6LOPWK0CSNB39M5GDIJHV")
s3 = client.integration("test_integration")
db = client.integration("aqueduct_demo")

####### This works and writes correctly to the S3 integration ######
@op
def generate_str():
    return "hello man"

output = generate_str()
output.save(s3.config("hello_path"))

client.publish_flow("S3 Test", artifacts=[output])

###### This fails if no S3FileFormat is supplied (see first image)
sql_output = db.sql("Select * from hotel_reviews")
sql_output.save(s3.config("tabular"))
client.publish_flow("Tabular into S3", artifacts=[sql_output])

##### This fails (see second image)
@op
def generate_str():
    return "hello man"

output = generate_str()
output.save(db.config("table_name", update_mode="append"))

client.publish_flow("S3 Failed", artifacts=[output])
```
Screen Shot 2022-09-01 at 4 04 17 PM
Screen Shot 2022-09-01 at 4 06 06 PM

ENG-1338


vsreekanti:
From a UX perspective, this looks great @kenxu95!

src/golang/lib/collections/operator/connector/load_params.go
```diff
@@ -37,8 +37,8 @@ type SalesforceLoadParams struct {
 }
 
 type S3LoadParams struct {
-	Filepath string `json:"filepath"`
-	Format   string `json:"format"`
+	Filepath string  `json:"filepath"`
+	Format   *string `json:"format"`
```
cw75, line 41: Is the reason we need to change this to a pointer because we allow it to be None on the Python side? If so let's add a comment.

sdk/aqueduct/integrations/integration.py
```diff
@@ -25,6 +25,20 @@ def describe(self) -> None:
...
 
+    def is_relational(self) -> bool:
+        """Returns whether the integration connects to a relational data store."""
+        return self.service in [
+            ServiceType.POSTGRES,
+            ServiceType.SNOWFLAKE,
+            ServiceType.MYSQL,
+            ServiceType.REDSHIFT,
+            ServiceType.MARIADB,
+            ServiceType.SQLSERVER,
```
cw75, line 36: Reminder to add service type ATHENA. Although since this helper is currently only relevant to load operators, Athena shouldn't be relevant.

cw75 (commented):
LGTM. A couple comments inline. Also I think we should make `save` a parent class function in `base_artifact.py`? This way it also applies to numeric and bool artifact classes.

src/golang/lib/collections/operator/connector/load_params.go
kenxu95: yeah its because the pydantic model we expect to deserialize this into is Optional[enum], which was failing when getting the empty string.

cw75:
@kenxu95 did you see the comment of making `save` available to other artifact classes as well? Maybe we can put it under `base_artifact.py` since the implementation seems to be the same. Otherwise LGTM.

kenxu95:
@cw75 I'm trying to figure one leftover failing test, but I've done the move of save() to base_artifact

cw75 (approved):

PR #429, opened 2022-09-06
Shows users workflows without any runs

saurav-c:
This PR adds frontend and backend changes to show users workflows without any runs. This is important as it at least gives the user some confidence that Aqueduct registered their workflow.

ENG 1485

DEMO: https://www.loom.com/share/bc3d00dfda7e43fe9609c07eaba4f632

### DIFF
--- a/src/golang/lib/collections/workflow/postgres.go
+++ b/src/golang/lib/collections/workflow/postgres.go
@@ -26,7 +26,7 @@ func (r *postgresReaderImpl) GetWorkflowsWithLatestRunResult(
 	ctx context.Context,
 	organizationId string,
 	db database.Database,
-) ([]latestWorkflowResponse, error) {
+) ([]LatestWorkflowResponse, error) {
 	// Get workflow metadata (id, name, description, creation time, last run time, and last run status)
 	// for all workflows whose `organization_id` is `organizationId` ordered by when the workflow was created.
 	// Get the last run DAG by getting the max created_at timestamp for all workflow DAGs associated with each
@@ -47,7 +47,7 @@ func (r *postgresReaderImpl) GetWorkflowsWithLatestRunResult(
 	) AS temp
 	ORDER BY created_at DESC;`
 
-	var latestWorkflowResponse []latestWorkflowResponse
+	var latestWorkflowResponse []LatestWorkflowResponse
 	err := db.Query(ctx, &latestWorkflowResponse, query, organizationId)
 	return latestWorkflowResponse, err
 }
--- a/src/golang/lib/collections/workflow/postgres.go
+++ b/src/golang/lib/collections/workflow/postgres.go
@@ -31,14 +31,21 @@ func (r *postgresReaderImpl) GetWorkflowsWithLatestRunResult(
 	// for all workflows whose `organization_id` is `organizationId` ordered by when the workflow was created.
 	// Get the last run DAG by getting the max created_at timestamp for all workflow DAGs associated with each
 	// workflow in the organization.
-	query := `SELECT * FROM (SELECT DISTINCT ON (wf.id) wf.id AS id, wf.name AS name, 
-		 wf.description AS description, wf.created_at AS created_at, 
-		 wfdr.created_at AS last_run_at, wfdr.status as status 
-		 FROM workflow AS wf, app_user, workflow_dag AS wfd, workflow_dag_result AS wfdr 
-		 WHERE app_user.organization_id = $1 
-		 AND wf.user_id = app_user.id AND wfd.workflow_id = wf.id AND wfdr.workflow_dag_id = wfd.id 
-		 ORDER BY wf.id, wfdr.created_at DESC) AS temp
-		 ORDER BY created_at DESC;`
+	query := `
+	SELECT * FROM (
+		SELECT DISTINCT ON 
+			(wf.id) wf.id AS id, wf.name AS name, 
+		 	wf.description AS description, wf.created_at AS created_at, 
+		 	wfdr.created_at AS last_run_at, wfdr.status as status 
+		FROM 
+			workflow AS wf 
+			INNER JOIN app_user ON wf.user_id = app_user.id
+			LEFT JOIN workflow_dag AS wfd ON wf.id = wfd.workflow_id
+			LEFT JOIN workflow_dag_result AS wfdr ON wfd.id = wfdr.dag_id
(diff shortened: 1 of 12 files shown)


agiron123:
What do we call the status of a workflow with no runs? Is it 'Registered'?

src/golang/lib/collections/workflow/workflow.go
```diff
@@ -19,13 +20,13 @@ type Workflow struct {
...
 }
 
-type latestWorkflowResponse struct {
-	Id          uuid.UUID              `db:"id" json:"id"`
-	Name        string                 `db:"name" json:"name"`
-	Description string                 `db:"description" json:"description"`
-	CreatedAt   time.Time              `db:"created_at" json:"created_at"`
-	LastRunAt   time.Time              `db:"last_run_at" json:"last_run_at"`
-	Status      shared.ExecutionStatus `db:"status" json:"status"`
+type LatestWorkflowResponse struct {
```
saurav-c, line 23: This needed to be exported to access in the test

src/golang/lib/collections/workflow/postgres.go
```diff
@@ -26,21 +26,28 @@ func (r *postgresReaderImpl) GetWorkflowsWithLatestRunResult(
...
+	SELECT * FROM (
+		SELECT DISTINCT ON 
+			(wf.id) wf.id AS id, wf.name AS name, 
+		 	wf.description AS description, wf.created_at AS created_at, 
+		 	wfdr.created_at AS last_run_at, wfdr.status as status 
+		FROM 
+			workflow AS wf 
+			INNER JOIN app_user ON wf.user_id = app_user.id
+			INNER JOIN workflow_dag AS wfd ON wf.id = wfd.workflow_id
+			LEFT JOIN workflow_dag_result AS wfdr ON wfd.id = wfdr.workflow_dag_id
```
saurav-c, line 44: The changes here are making all of the `INNER JOIN`s explicit and replacing the final `INNER JOIN` with a `LEFT JOIN`.

src/golang/lib/collections/workflow/sqlite.go
```diff
@@ -53,26 +53,39 @@ func (r *sqliteReaderImpl) GetWorkflowsWithLatestRunResult(
...
-		  FROM workflow AS wf, app_user, workflow_dag AS wfd, workflow_dag_result AS wfdr 
-		  WHERE app_user.organization_id = $1 
-		  AND wf.user_id = app_user.id AND wfd.workflow_id = wf.id AND wfdr.workflow_dag_id = wfd.id 
-		  GROUP BY wf.id) AS wflr 
-		 WHERE app_user.organization_id = $1 
-		 AND wf.user_id = app_user.id AND wfd.workflow_id = wf.id AND wfdr.workflow_dag_id = wfd.id AND 
-		 wf.id = wflr.id AND wfdr.created_at = wflr.last_run_at 
-		 ORDER BY created_at DESC;`
+	query := `
+		WITH workflow_results AS
```
saurav-c, line 62: There's a lot of rewriting her to make things simpler imo. It's easier to understand views than subqueries, when the subquery isn't necessary.

saurav-c:
@agiron123 It should be `Pending`. That's what the backend returns.

saurav-c:
cc @vsreekanti To take a look at the demo

cw75:
@saurav-c I'd replace the green check mark with a yellow warning mark when you click a workflow with no runs.

vsreekanti:
Let's actually do an info alert rather than success or warning. It's neither good nor bad, it's just a fact. 

Other than that, this looks good to me. I agree with @agiron123's point that we might want to have a different status. In Progress is a little misleading because for Airflow, the new workflow might not even be registered yet. How hard would it be to add that @saurav-c?

src/golang/lib/collections/workflow/postgres.go
```diff
@@ -26,21 +26,28 @@ func (r *postgresReaderImpl) GetWorkflowsWithLatestRunResult(
...
+		WHERE app_user.organization_id = $1 
+		ORDER BY wf.id, wfdr.created_at DESC
+	) AS temp
```
kenxu95, line 47: Is `AS temp` always necessary in `SELECT ... ORDER BY` cases??

src/golang/lib/collections/workflow/postgres.go
kenxu95: Can you explain the left join here a little more? I've been staring at it for a while but want to make sure :P 

My understanding is that a left join takes all the records of the left table (which are `workflow` records), and any matching records on the right table. Does this mean that at least one record is returned for each workflow, but there can be workflows with no workflow dag results, meaning the `last_run_at` and `status` fields of such workflows are empty?

If thats the case, we can add a comment noting that potential nil-ness? I'd rather our SQL queries be really well documented than not since these take a lot of brainpower to read.

src/golang/lib/collections/workflow/sqlite.go
```diff
@@ -53,26 +53,39 @@ func (r *sqliteReaderImpl) GetWorkflowsWithLatestRunResult(
...
+			SELECT id, MAX(run_at) AS last_run_at
+	  		FROM workflow_results
+	  		GROUP BY id
+		)
+		SELECT wfr.id, wfr.name, wfr.description, wfr.created_at, wfr.run_at AS last_run_at, wfr.status
+		FROM workflow_results AS wfr, latest_result AS lr
+		WHERE wfr.id = lr.id
+		AND 
+		(	wfr.run_at = lr.last_run_at
+			OR (wfr.run_at IS NULL AND lr.last_run_at IS NULL)
```
kenxu95, line 84: Similar comment as above - maybe document what returned fields are potentially null?

kenxu95: I guess one could infer it from the field types in `LatestWorkflowResponse`, but I still think its useful context for the reader.

kenxu95 (approved):
Nice! Left some readability comments. I had fun learning/reading sql again

saurav-c:
@vsreekanti it depends, are we just adding a new status called `Pending` to show to the users on the UI when a workflow has no runs?

vsreekanti:
I think so, yeah. Were you imagining something different?

saurav-c:
Screen Shot 2022-09-07 at 5 50 07 PM

@vsreekanti I changed it to the following

vsreekanti:
LGTM!

PR #425, opened 2022-09-02
[1/2] ENG-1625 Adds route implementation to fetch a single workflow dag result

likawind:
## Describe your changes and why you are making these changes
This PR adds the route to retrieve full results for a given dag. It's intended to unblock immediate UI redesign work and should not affect any existing APIs. On the other hand we hope the same response structure could be reused for future work to formalize our REST API organization. I'm currently drafting some motivations here: https://www.notion.so/aqueducthq/Formalize-Application-Client-Interfaces-389fe79284424c8190d4d9823dc45989 .

This PR doesn't include integration tests which will be introduced in an immediate follow-up PR.

## Related issue number (if any)
ENG-1625

## Checklist before requesting a review
Tested response in postman for a workflow run:
```
API: http://3.14.83.194:8080/api/workflow/449744a6-edd3-48cc-b8e7-5723b98f8987/result/82c43c2b-4423-4b72-b3d4-83efdf1bd265
{
    "dag_id": "32efecb1-425c-47f4-80aa-c462bc02aa6c",
    "dag_created_at": "2022-09-02T00:28:01.359416223Z",
    "storage_config": {
        "type": "file",
        "file_config": {
            "directory": "/home/ubuntu/.aqueduct/server/storage"
        }
    },
    "engine_config": {
        "type": "aqueduct"
    },
    "workflow_id": "449744a6-edd3-48cc-b8e7-5723b98f8987",
    "workflow_created_at": "2022-09-02T00:28:01.359121767Z",
    "user_id": "0678fdc2-1641-4d42-98e4-8e4566825ba1",
    "name": "test",
    "description": "test",
    "schedule": {
        "trigger": "manual",
        "cron_schedule": "",
        "disable_manual_trigger": false,
        "paused": false
    },
    "retention_policy": {
        "k_latest_runs": -1
    },
    "result": {
        "id": "82c43c2b-4423-4b72-b3d4-83efdf1bd265",
        "status": "succeeded",
        "created_at": "2022-09-02T00:28:01.38329606Z"
    },
    "operators": {
        "0dc0b479-ba84-4b05-941b-f98acc676021": {
            "id": "0dc0b479-ba84-4b05-941b-f98acc676021",
            "name": "aqueduct_demo query 1",
            "description": "",
            "spec": {
                "type": "extract",
                "extract": {
                    "service": "SQLite",
                    "integration_id": "1aee20a6-a1de-4458-9992-0a8854b3f4f1",
                    "parameters": {
                        "github_metadata": null,
[...]

PR #435, opened 2022-09-07
[2/2] Eng 1625 Adds frontend structs and handler for getWorkflowDagResult API

likawind:
This PR adds frontend response structs and async thunk code to handle `getWorkflowDagResult` API introduced in #425 . It mainly mirrors response structs from golang with a simple async thunk to process the request. We need to test this together with the new UI code

ENG-1625

This PR only adds stuff that's not called, so no test is necessary as long as linter passes.


src/ui/common/src/handlers/responses/dag.ts
```diff
@@ -0,0 +1,42 @@
...
+  schedule?: WorkflowSchedule;
+  retention_policy?: WorkflowRetentionPolicy;
+};
+
+export type DagResponse = DagMetadataResponse & {
+  operators: { [id: string]: OperatorResponse };
+  artifacts: { [id: string]: ArtifactResponse };
+};
+
+export type DagRawResultResponse = {
```
eunice-chan, line 32: I wonder if there is a more descriptive name for this result. DagRawResultResponse doesn't let me know what it contains. I think DagMetadatResponse would be more descriptive but that name is already in use.

eunice-chan (approved):
I like the comments about the associated `.go` file for the `.ts` files you wrote ("this file should mirror...")!

src/ui/common/src/handlers/responses/dag.ts
likawind: yeah I had a hard time figuring this out as well. This is technically a `ResultResponse` but that's also used by the main `ResultResponse`

agiron123: Maybe use something like DagResultStatus here @likawind ?

agiron123 (approved):
Nice work! Small comment about naming below, but after resolved you can go ahead and merge.

PR #448, opened 2022-09-08
Fix bug where artifact with None value doesn't work with flow_run.artifact() api

cw75:
Previously, if an artifact's content is `None`, when we retrieve it from `flow_run.artifact()` and call `get()`, it'll mistakenly complain the artifact wasn't computed successfully. This PR fixes this by distinguishing the two cases via an extra execution status argument passed into `GenericArtifact`.

### DIFF
--- a/integration_tests/sdk/no_return_test.py
+++ b/integration_tests/sdk/no_return_test.py
@@ -1,6 +1,7 @@
 import pytest
-from aqueduct import op
 from utils import run_flow_test
+
+from aqueduct import op
 
 
 @op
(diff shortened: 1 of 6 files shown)


integration_tests/sdk/no_return_test.py
```diff
@@ -0,0 +1,22 @@
...
+@op
+def no_return() -> None:
+    return None
+
+
+@pytest.mark.publish
+def test_operator_with_no_return(client):
+    result = no_return()
+    assert result.get() is None
+    flow = run_flow_test(client, artifacts=[result], delete_flow_after=False)
```
kenxu95, line 16: nit: can we put this in the try-statement? That way if it publishes but fails, we still delete it. We have a lot of tests like this actually, but I'm planning a refactor of the test utils soon.

kenxu95 (approved):
Makes sense to me!

PR #449, opened 2022-09-09
Deprecate ParamArtifact class

cw75:
When calling `create_param()`, we now return a regular artifact class instead; there is no need for a separate `ParamArtifact` class.


kenxu95 (approved):
This has made me realize, we should add the `validate_overwriting_parameters()` check to the `.get()` method of every artifact. Can you also include that in this PR?

But other than that thought, looks good to me!

cw75:
@kenxu95 inside `get()`, when we call `preview_artifract()`, it uses `UpdateParametersDelta` which already makes a call to `validate_overwriting_parameters()`. So it seems we don't have to add it in `get()`?

PR #453, opened 2022-09-09
Expand parameters support to non-json types

kenxu95:
NOTE: 
- This change breaks custom parameter triggering from the UI. I will send out a follow-up to fix up that caller.
- This change is not backwards compatible with previous versions because it changes the parameter schema in the backend. We will need to run a data migration (also will be a follow-up)

This change also enforces consistent types on parameter on the backend.

Tagging @vsreekanti to look at what this unlocks in `test_non_jsonable_param_types`.

ENG-1565


sdk/aqueduct/dag_deltas.py
```diff
@@ -0,0 +1,329 @@
+import copy
```
kenxu95, line 1: I had to move this into its own file so that I could import from utils. Conceptually, dag deltas should be the layer above dag.py.

sdk/aqueduct/utils.py
```diff
@@ -516,6 +521,23 @@ def infer_artifact_type(value: Any) -> ArtifactType:
...
+    # what encoding it was written in (eg. Image types are not encoded as "utf8").
+    return ParamSpec(
+        val=serialize_param_val(val, serialization_type),
+        serialization_type=serialization_type,
+    )
+
+
+def serialize_param_val(val: Any, serialization_type: SerializationType) -> str:
+    val_bytes = serialization_function_mapping[serialization_type](val)
+    return bytes_to_base64_string(val_bytes)
```
cw75, line 538: I think we can rename this to `serialize_val` and put it in `serialization.py`.

cw75 (commented):
@kenxu95 Looks good! Just to confirm this PR doesn't contain the part where the Go side writes the base64 encoded data to a file, which is read by the Python side right? Seems that we are still directly passing the data as part of the job spec. Are we making this change in a follow-up?

Since we are duplicating a bunch of the serialization logic in the SDK and in the executor, maybe we can let the executor import the sdk and call these helper functions, but this can be in a follow-up PR.

cw75 (approved):

kenxu95:
> @kenxu95 Looks good! Just to confirm this PR doesn't contain the part where the Go side writes the base64 encoded data to a file, which is read by the Python side right? Seems that we are still directly passing the data as part of the job spec. Are we making this change in a follow-up?

Yep that's a follow up https://github.com/aqueducthq/aqueduct/pull/479

> Since we are duplicating a bunch of the serialization logic in the SDK and in the executor, maybe we can let the executor import the sdk and call these helper functions, but this can be in a follow-up PR.

Sounds like a tech debt task, dunno if we should prioritize it now: https://linear.app/aqueducthq/issue/ENG-1706/decide-on-and-implement-a-way-to-share-python-code-between-the-sdk-and

PR #454, opened 2022-09-09
Implicitly create parameter when passing non artifact to decorated functions

cw75:
When non-artifact types are passed to decorated functions, we attempt to implicitly create a parameter with name being the function argument's name and default value being the user-specified value. We print out a warning when this happens. If there is already an operator with the same name in the dag, we throw an error.

### DIFF
--- a/integration_tests/sdk/param_test.py
+++ b/integration_tests/sdk/param_test.py
@@ -95,17 +95,20 @@ def test_implicitly_created_parameter(client):
 
     result = func(2)
     assert result.get() == 2
-    client._dag.must_get_operator(with_name="foo")
+    assert result.get(parameters={"foo": 10}) == 10
 
-    client.create_param("bar", default="hello")
+    bar_param = client.create_param("bar", default="hello")
 
     @op
     def another_func(bar):
-        return foo
+        return bar
 
     with pytest.raises(InvalidUserArgumentException):
         # This should error because we try to implicitly create a parameter "bar" but it already exists.
         result = another_func(2)
+
+    result = another_func(bar_param)
+    assert result.get() == "hello"
 
 
 @op
--- a/integration_tests/sdk/param_test.py
+++ b/integration_tests/sdk/param_test.py
@@ -88,6 +88,26 @@ def test_get_with_custom_parameter(client):
         param_doubled.get(parameters={"number": "NOT A NUMBER"})
 
 
+def test_implicitly_created_parameter(client):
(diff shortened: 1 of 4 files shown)


integration_tests/sdk/param_test.py
```diff
@@ -88,6 +88,26 @@ def test_get_with_custom_parameter(client):
...
+
+    client.create_param("bar", default="hello")
+
+    @op
+    def another_func(bar):
+        return foo
+
+    with pytest.raises(InvalidUserArgumentException):
+        # This should error because we try to implicitly create a parameter "bar" but it already exists.
+        result = another_func(2)
```
kenxu95, line 108: Can you also check that `another_func(<bar_param>)` still works here? Names can coexist as long as you don't use them implicitly right.

integration_tests/sdk/param_test.py
```diff
@@ -88,6 +88,26 @@ def test_get_with_custom_parameter(client):
...
 
 
+def test_implicitly_created_parameter(client):
+    @op
+    def func(foo):
+        return foo
+
+    result = func(2)
+    assert result.get() == 2
+    client._dag.must_get_operator(with_name="foo")
```
kenxu95, line 98: In the interest of keeping our SDK tests at the user level, can't we simply verify this with `result.get(parameters={"foo": 10} == 10`, instead of inspecting the internal dag.

sdk/aqueduct/decorator.py
```diff
@@ -181,6 +184,36 @@ def _type_check_decorated_function_arguments(
...
+    dag = dag_module.__GLOBAL_DAG__
+
+    artifacts = list(input_artifacts)
+    for idx, artifact in enumerate(artifacts):
+        if not isinstance(artifact, BaseArtifact):
+            arg_name = function_argument_names[idx]
+            if dag.get_operator(with_name=arg_name) is not None:
+                raise InvalidUserArgumentException(
+                    """Input to function argument "%s" is not an artifact type. We tried implicitly \
+creating a parameter named "%s", but an existing operator or parameter with the same name already exist."""
```
kenxu95, line 202: grammar: "already exists"

kenxu95 (changes requested):
Looks good, just a few testing requests. Can stamp when those are implemented

cw75:
@kenxu95 addressed comments.

kenxu95 (approved):

PR #456, opened 2022-09-11
Implements Lambda Workflows [SDK] [1/5]

hsubbaraj-spiral:
Implements Lambda Workflows.

### DIFF
--- a/sdk/aqueduct/integrations/lambda_integration.py
+++ b/sdk/aqueduct/integrations/lambda_integration.py
@@ -1,0 +1,15 @@
+from aqueduct.integrations.integration import Integration, IntegrationInfo
+
+
+class LambdaIntegration(Integration):
+    """
+    Class for K8s integration.
+    """
+
+    def __init__(self, metadata: IntegrationInfo):
+        self._metadata = metadata
+
+    def describe(self) -> None:
+        """Prints out a human-readable description of the K8s integration."""
+        print("==================== K8s Integration  =============================")
+        self._metadata.describe()
--- a/sdk/aqueduct/utils.py
+++ b/sdk/aqueduct/utils.py
@@ -14,7 +14,13 @@ import cloudpickle as pickle
 import multipart
 import numpy as np
 import requests
-from aqueduct.config import AirflowEngineConfig, EngineConfig, FlowConfig, K8sEngineConfig, LambdaEngineConfig
+from aqueduct.config import (
+    AirflowEngineConfig,
+    EngineConfig,
+    FlowConfig,
+    K8sEngineConfig,
+    LambdaEngineConfig,
+)
 from aqueduct.dag import DAG, RetentionPolicy, Schedule
 from aqueduct.enums import ArtifactType, OperatorType, RuntimeType, TriggerType
 from aqueduct.error import *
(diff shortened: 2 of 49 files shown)


sdk/aqueduct/integrations/lambda_integration.py
```diff
@@ -0,0 +1,15 @@
+from aqueduct.integrations.integration import Integration, IntegrationInfo
+
+
+class LambdaIntegration(Integration):
```
eunice-chan, line 4: May want to update the doc strings

sdk/aqueduct/utils.py
```diff
@@ -556,5 +557,12 @@ def generate_engine_config(config: Optional[FlowConfig]) -> EngineConfig:
                 integration_id=config.engine._metadata.id,
             ),
         )
+    elif isinstance(config.engine, LambdaIntegration):
+        return EngineConfig(
+            type=RuntimeType.LAMBDA,
+            lambda_config=LambdaEngineConfig(
```
eunice-chan, line 563: I wonder if we can have a EngineConfig(type=..., config=...) instead of a lambda_config variable, and infer what the config is for by looking at the engine type. Is that something we can do? Why do we choose to have a variable specifically for lambda_config instead?

eunice-chan: Also I don't see the EngineConfig file in this PR. Is this a change already in main?

hsubbaraj-spiral: We need a way for the Golang DBWorkflowDag to identify both what type of engine it is using and the config for that engine integration, which is why we include the EngineConfig as part of the DBWorkflowDag.
```
type DBWorkflowDag struct {
	Id            uuid.UUID            `db:"id" json:"id"`
	WorkflowId    uuid.UUID            `db:"workflow_id" json:"workflow_id"`
	CreatedAt     time.Time            `db:"created_at" json:"created_at"`
	StorageConfig shared.StorageConfig `db:"storage_config" json:"storage_config"`
	EngineConfig  shared.EngineConfig  `db:"engine_config" json:"engine_config"`

	/* Field not stored in DB */
	Metadata  *workflow.Workflow                `json:"metadata"`
	Operators map[uuid.UUID]operator.DBOperator `json:"operators,omitempty"`
	Artifacts map[uuid.UUID]artifact.DBArtifact `json:"artifacts,omitempty"`
}
```

```
type EngineConfig struct {
	Type           EngineType      `yaml:"type" json:"type"`
	AqueductConfig *AqueductConfig `yaml:"aqueductConfig" json:"aqueduct_config,omitempty"`
	AirflowConfig  *AirflowConfig  `yaml:"airflowConfig" json:"airflow_config,omitempty"`
	K8sConfig      *K8sConfig      `yaml:"k8sConfig" json:"k8s_config,omitempty"`
	LambdaConfig   *LambdaConfig   `yaml:"lambdaConfig" json:"lambda_config,omitempty"`
}
```
The `EngineConfig` struct follows a similar pattern to how we create a field with different values. The Pydantic struct has to mirror this Golang struct.

The EngineConfig is declared in `sdk/aqueduct/config.py`. Lmk if this clears things up!

kenxu95 (approved):
Seems harmless enough - agree that we should update the docstrings

PR #461, opened 2022-09-12
Eng 1598 allow details pages to be displayed as MUI Drawer

agiron123:

This PR switches out the Aqueduct sidesheet component for an MUI Drawer component.
Future work here will let us show the contents of our newly redesigned UI right in this sidesheet.

ENG-1598 (https://linear.app/aqueducthq/issue/ENG-1598/allow-details-pages-to-be-displayed-as-mui-drawers)

Loom Demo: https://www.loom.com/share/d28c0445ab1d49d5926d29f409dd45f6


src/ui/common/src/components/pages/workflow/id/index.tsx
```diff
@@ -236,13 +235,13 @@ const WorkflowPage: React.FC<WorkflowPageProps> = ({
     openSideSheetState.workflowStatusBarOpen,
     fullWindowWidth
   );
-  let contentBottomOffsetInPx;
+  const contentBottomOffsetInPx = 0;
```
agiron123, line 238: Let's remove this contentBottomOffsetInPx variable entirely since it's not being used anymore.

src/ui/common/src/components/pages/workflow/id/index.tsx
```diff
@@ -324,14 +323,31 @@ const WorkflowPage: React.FC<WorkflowPageProps> = ({
       </Box>
 
       {currentNode.type !== NodeType.None && (
```
agiron123, line 325: AqueductSidebar component can now also safely be deleted.

src/ui/common/src/components/pages/workflow/id/index.tsx
```diff
@@ -232,17 +230,18 @@ const WorkflowPage: React.FC<WorkflowPageProps> = ({
...
-    fullWindowWidth
-  );
-  let contentBottomOffsetInPx;
-
-  if (openSideSheetState.bottomSideSheetOpen) {
-    contentBottomOffsetInPx = `${BottomSidebarHeightInPx + 20}px`;
-  } else {
-    contentBottomOffsetInPx = `${BottomSidebarHeaderHeightInPx + 20}px`;
-  }
+  // const contentWidth = getBottomSideSheetWidth(
```
agiron123, line 233: remove comments.

src/ui/common/src/components/DataPreviewer/index.tsx
```diff
@@ -99,6 +99,8 @@ const DataPreviewer: React.FC<Props> = ({ previewData, error }) => {
               width: '100%',
               overflow: 'auto',
               overflowY: 'hidden',
+              minWidth: '400px',
+              position: 'absolute',
```
vsreekanti, line 103: Just out of curiosity, why do we need the absolute positioning here?

src/ui/common/src/components/pages/workflow/id/index.tsx
```diff
@@ -222,28 +220,6 @@ const WorkflowPage: React.FC<WorkflowPageProps> = ({
     return null;
   }
 
-  // NOTE(vikram): This is a compliated bit of nonsense code. Because the
```
vsreekanti, line 225: @agiron123, do you have thoughts on the tradeoffs of shrinking the viewport when we open the drawer vs. not? I'm happy to be rid of this code, but I did think that was a slick feature.

That said, I didn't really notice it in the demo, so I don't have a super strong opinion.

src/ui/common/src/components/pages/workflow/id/index.tsx
```diff
@@ -324,14 +300,31 @@ const WorkflowPage: React.FC<WorkflowPageProps> = ({
...
       {currentNode.type !== NodeType.None && (
-        <AqueductSidebar
-          zIndex={10}
-          position={SidebarPosition.bottom}
-          getSideSheetTitle={getNodeLabel}
-          getSideSheetHeadingContent={getNodeActionButton}
-        >
-          {getDataSideSheetContent(user, currentNode)}
-        </AqueductSidebar>
+        <Drawer anchor="right" variant="persistent" open={true}>
```
vsreekanti, line 303: How come open is hardcoded to `true` here? Is it because we rely on the content of the sidesheet to be `null`/`undefined` when nothing is selected and therefore nothing gets rendered?

I'm curious just because it might be cleaner in the long run to have an explicit control rather than rely on state emptiness.

vsreekanti (approved):
LGTM, @agiron123! I had a couple small questions inline, mostly out of curiosity. Great work!

src/ui/common/src/components/DataPreviewer/index.tsx
agiron123: Good question, without having position:absolute, our artifact previews were not being shown in the drawer.

I need to dig a little deeper into the stylesheet of the Drawer component, but this is what I had to do to get things to work :)

src/ui/common/src/components/pages/workflow/id/index.tsx
agiron123: The code here was adding a bottom offset to the page content, so every time the drawer was opened from the side, content was shifted upwards, which didn't make much sense anymore.

What we can do instead is change this to shift page content to the left every time that the drawer opens from the right side.

src/ui/common/src/components/pages/workflow/id/index.tsx
agiron123: The drawer's open/close state is actually controlled by whether or not a node is selected. to close, we simply reset the selectedNode.

So in our case, we can leave the node open since this code is wrapped in a conditional check to show the entire drawer.

We could also move this conditional check to be inside the open prop as well.

PR #452, opened 2022-09-09
Eng 1614 add update timestamps for operators and artifacts

likawind:
## Describe your changes and why you are making these changes
This PR adds relevant timestamps for operator and artifacts. Concretely we added a `ExecutionTimestamps` struct to `ExecutionState` to reflect `PendingAt`, `RunningAt`, `SucceededAt` and `FailedAt` time for the execution.

In order to update application logic that updates those timestamps, we made the following changes to operator execution:
* make `Operator` object stateful to record its current `execState`. I think this change is valid since the object already track runtime information like metadata path.
* Split current `GetExecState` to `Poll` (which access job manager and make changes) and `ExecState()` (which just fetch the state). The split provides better decoupling between job manager access and state retrieval, and provide better clarity on caller level. A consequence being `PersistResult` now becomes a pure DB API as we replaced `GetExecState()` call which access underlying job poll, with `ExecState()` which just look at the current `execState` based on any previous `Poll()`.
* To cleanly deal with operator result from caching, we choose to mare succeed / fail timestamp based on the timestamp of calling `Poll()` rather than the timestamp of execution.

## Related issue number (if any)
ENG-1614

## Checklist before requesting a review
We added integration tests separately in #463 , tracked in https://linear.app/aqueducthq/issue/ENG-1640/add-integration-tests-for-fetching-workflow-and-results
Tested using `api/workflow/{wfId}/result/{resultId}` and confirmed all op and artf are generated with expected timestamps:
```
{
    "dag_id": "27633b8f-5b4e-43ee-8106-52af844e81bb",
    "dag_created_at": "2022-09-09T22:24:04.854525698Z",
    "storage_config": {
        "type": "file",
        "file_config": {
            "directory": "/home/ubuntu/.aqueduct/server/storage"
        }
    },
    "engine_config": {
        "type": "aqueduct"
    },
    "workflow_id": "68b00347-6527-4951-8dcb-b0b2fbda181c",
    "workflow_created_at": "2022-09-09T22:24:04.854267779Z",
    "user_id": "0678fdc2-1641-4d42-98e4-8e4566825ba1",
    "name": "test_timestamp_failed_more",
    "description": "test",
    "schedule": {
        "trigger": "manual",
        "cron_schedule": "",
        "disable_manual_trigger": false,
        "paused": false
    },
    "retention_policy": {
        "k_latest_runs": -1
    },
    "result": {
        "id": "4834b0e0-c70a-4230-8892-1a1b9597212b",
        "status": "failed",
[...]

PR #463, opened 2022-09-13
Eng 1640 Add integration test for `get_workflow_dag_result` route

likawind:
This PR adds backend tests for `get_workflow_dag_result` route based on changes introduced in #425 and #452 . Specifically we check a failed workflow and try to fetch its dag result. We verify that workflow result, each operator's `exec_state`, and each artifact's `exec_state` are expected. We also verified all timestamps are consistent.

Most responses are handled as a dict, which we could improve to use pydantic. This is tracked separately in ENG-1685.

ENG-1640


integration_tests/backend/setup/flow_with_failure.py
```diff
@@ -0,0 +1,32 @@
+import aqueduct
+
+###
+# Workflow that loads a table from the `aqueduct_demo` then saves it to `table_1` in append mode.
+# This save operator is then replaced by one that saves to `table_1` in replace mode.
+# In the next deployment of this run, it saves to `table_1` in append mode.
+# In the last deployment, it saves to `table_2` in append mode.
```
kenxu95, line 7: This comment doesn't seem relevant at all?

kenxu95 (approved):
Thanks! I think we're missing the RUNNING status coverage - I wonder if we can just a separate flow that has just one operator which sleeps for a while until we delete it? Nothing too fancy.

likawind:
@kenxu95 I believe we are not persisting the state when the workflow or operator is being launched, which means the state will be `pending`. It does make sense to add the test though

PR #467, opened 2022-09-13
ENG-1639 Switch to Newest workflow on Run

agiron123:
This PR loads the most recent workflow in the workflow details page after the user triggers a workflow from the UI.

This PR contains work that I did while trying to resolve ENG-1639.
It's adjacent, yet still doesn't solve the bug in 1639.

## Loom Demo:
https://www.loom.com/share/92c68f4c437d48b7bed39d13d037d8e5

ENG-1639

### DIFF
--- a/examples/churn_prediction/Build and Deploy Churn Ensemble.ipynb
+++ b/examples/churn_prediction/Build and Deploy Churn Ensemble.ipynb
@@ -36,7 +36,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 45,
+   "execution_count": 1,
    "id": "4dff0960-95f2-4e43-9c01-7409fbb12c3f",
    "metadata": {},
    "outputs": [
@@ -178,7 +178,7 @@
        "4       True  False  "
       ]
      },
-     "execution_count": 45,
+     "execution_count": 1,
      "metadata": {},
      "output_type": "execute_result"
     }
@@ -204,7 +204,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 46,
+   "execution_count": 2,
    "id": "e8960448-0658-4bfc-aac6-f87fa9b415af",
    "metadata": {},
    "outputs": [],
@@ -232,7 +232,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 47,
+   "execution_count": 3,
    "id": "3e6f4d6b-7795-4043-b724-d2f9fe5483b3",
    "metadata": {},
    "outputs": [
@@ -424,7 +424,7 @@
        "4         1.098612  "
       ]
      },
-     "execution_count": 47,
+     "execution_count": 3,
      "metadata": {},
      "output_type": "execute_result"
     }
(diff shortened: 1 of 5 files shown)


examples/churn_prediction/Build and Deploy Churn Ensemble.ipynb
```diff
@@ -36,7 +36,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 45,
```
agiron123, line 39: Wish there was a better way to ignore notebooks so that when I run them they don't appear in git diffs. i'm not sure if this exists :)

src/ui/common/src/components/workflows/StatusBar.tsx
```diff
@@ -124,9 +124,10 @@ const ActiveWorkflowStatusTab: React.FC<ActiveWorkflowStatusTabProps> = ({
       }}
     >
       {listItems.map((listItem, index) => {
+        const key = listItem.nodeId.length > 0 ? listItem.nodeId : index;
```
agiron123, line 127: Goal of the change here was to use the ids of each node as the key for the list.
React wants unique keys to be used to help it know when to re-render things.

Since the list items here change often, using the index of the listItems isn't always ideal. In our case, we have placeholders which have no id set, so we just use the number (when user isn't in a selected tab).

src/ui/common/src/components/workflows/workflowHeader.tsx
```diff
@@ -43,8 +51,12 @@ const WorkflowHeader: React.FC<Props> = ({ user, workflowDag }) => {
   const [showSuccessToast, setShowSuccessToast] = useState(false);
   const [showSettings, setShowSettings] = useState(false);
 
-  const handleSuccessToastClose = () => {
+  const handleSuccessToastClose = async () => {
```
agiron123, line 54: Here's the main code that goes to the newly created workflow run after the user selects run workflow on the UI.

src/ui/common/src/components/workflows/workflowHeader.tsx
```diff
@@ -43,8 +51,12 @@ const WorkflowHeader: React.FC<Props> = ({ user, workflowDag }) => {
...
     setShowSuccessToast(false);
+    await dispatch(handleGetWorkflow({ apiKey: user.apiKey, workflowId }));
```
likawind, line 56: Will this block execution until the action get to `fulfilled` or `rejected` ?

likawind (approved):
Looks good beside some clarifications!

I'm curious why some artifacts looks succeeded while upstream operators are still in-progress? I guess there's some race condition on how we fetch artifact and operator states

agiron123:
> Looks good beside some clarifications!
> 
> I'm curious why some artifacts looks succeeded while upstream operators are still in-progress? I guess there's some race condition on how we fetch artifact and operator states

@likawind yes, that was part of the bug i originally set out to figure out. The issue here is most likely because the artifact and operator results that we are caching locally are keyed by their artifact and operator id, not the artifact result and operator result id.

src/ui/common/src/components/workflows/workflowHeader.tsx
agiron123: yes, it will wait till the handleGetWorkflow call completes before calling the next one, which is what we want to ensure we have our data ready.

One thing I could do is add a try catch here and handle errors a little better.

agiron123:
@likawind updated to add try/catch and an error message if fails:
```typescript
  const handleSuccessToastClose = async () => {
    setShowSuccessToast(false);

    try {
      await dispatch(handleGetWorkflow({ apiKey: user.apiKey, workflowId }));
      await dispatch(handleLoadIntegrations({ apiKey: user.apiKey }));
      dispatch(selectResultIdx(0));
      navigate(`/workflow/${workflowId}`, { replace: true });
    } catch (error) {
      setErrorMessage(`We're having trouble getting the latest workflow. Please try refreshing the page.`);
      setShowErrorToast(true);
    }
  };
```

PR #470, opened 2022-09-14
Eng 1641 include values when fetching small artifacts

likawind:
This PR adds an additional field in artifact response to provide the value of that artifact. This field should only work on small artifacts so that we are not returning a gigantic response object for getting the wf dag result. For larger artifacts, this field should be `null` and the caller need to call another route to fetch the contents.

We support includes values directly for `boolean` and `numeric` and `string` values. 

ENG-1641
Further improvements tracked in ENG-1687

Added integration tests.
Verified in postman for a workflow with metrics / checks:
```
"c04d6ba6-29f9-4379-ae5f-a1e987036cd6": {
            "id": "c04d6ba6-29f9-4379-ae5f-a1e987036cd6",
            "name": "less than 0.4 artifact",
            "description": "",
            "type": "boolean",
            "from": "5e10b1a4-78dc-49f7-8e02-54bc7c5372bb",
            "to": null,
            "result": {
                "id": "33733449-6f5f-48c9-99ac-8d898b1eaea2",
                "content_path": "39570465-882b-49f9-8e7c-1d1906bfdabc",
                "content_serialized": "true",
                "exec_state": {
                    "user_logs": {
                        "stdout": "",
                        "stderr": ""
                    },
                    "status": "succeeded",
                    "failure_type": null,
                    "error": null,
                    "timestamps": {
                        "registered_at": null,
                        "pending_at": "2022-09-13T19:00:00.023618648Z",
                        "running_at": "2022-09-13T19:00:16.0584769Z",
                        "finished_at": "2022-09-13T19:00:18.570771363Z"
                    }
                }
            }
        },
```


src/golang/lib/workflow/artifact/response.go
```diff
@@ -23,9 +23,17 @@ type Response struct {
...
-	ExecState   *shared.ExecutionState `json:"exec_state"`
+	Id uuid.UUID `json:"id"`
+
+	// If `ContentSerialized` is set, the content is small and we directly send
+	// it as a part of response. It's consistent with the object stored in `ContentPath`.
+	//
+	// Otherwise, the content is large and
+	// one should send an additional request to fetch the content.
+	ContentPath       string  `json:"content_path"`
+	ContentSerialized *string `json:"content_serialized"`
```
likawind, line 34: this is intentionally put as as pointer to distinguish 'value not available' vs 'value being an empty string'

integration_tests/backend/setup/flow_with_metrics_and_checks.py
```diff
@@ -0,0 +1,29 @@
+import aqueduct
+
+###
```
eunice-chan, line 3: May want to update the description

src/golang/cmd/server/handler/get_workflow_dag_result.go
```diff
@@ -153,10 +154,43 @@ func (h *GetWorkflowDagResultHandler) Perform(ctx context.Context, interfaceArgs
...
+func getArtifactContents(
+	ctx context.Context,
+	dbWorkflowDag *workflow_dag.DBWorkflowDag,
+	dbArtifactResults []artifact_result.ArtifactResult,
+) (map[string]string, error) {
+	contents := make(map[string]string, len(dbArtifactResults))
+	storageObj := storage.NewStorage(&dbWorkflowDag.StorageConfig)
+	for _, artfResult := range dbArtifactResults {
+		if artf, ok := dbWorkflowDag.Artifacts[artfResult.ArtifactId]; ok {
+			// These artifacts has small content size and we can safely include them all in response.
```
eunice-chan, line 181: Should we directly check if the content size is below some threshold rather than the artifact type?

eunice-chan (approved):

src/golang/cmd/server/handler/get_workflow_dag_result.go
```diff
@@ -153,10 +154,43 @@ func (h *GetWorkflowDagResultHandler) Perform(ctx context.Context, interfaceArgs
...
 	return dag.NewResultResponseFromDbObjects(
 		constructedDag,
 		dbWorkflowDagResult,
 		operatorResults,
 		artifactResults,
+		contents,
 	), http.StatusOK, nil
 }
+
+// TODO: We should replace this once we migrate to services based on `artifact` objects
```
kenxu95, line 171: Can we link a task here? Want to avoid hanging TODO's in our codebase.

src/golang/cmd/server/handler/get_workflow_dag_result.go
kenxu95: Might be useful for string types since those are somewhat unbounded. I don't know if the additional complexity is worth it though. I'm sort of on the side of keeping the semantics simple for now, to avoid more confusion in the future.

src/golang/lib/workflow/artifact/response.go
kenxu95: Can you note what it means for the content to be "serialized"? It's the raw bytes data at the storage path, converted to a string right?

src/golang/cmd/server/handler/get_workflow_dag_result.go
kenxu95: Can you note in the function comment that we only grab artifact contents for certain small artifact types? Essentially pull up the comment from line 181

kenxu95 (approved):
Fine with me! Just some documentation suggestions to make readability a bit easier

src/golang/cmd/server/handler/get_workflow_dag_result.go
likawind: Yeah this is a good idea!

PR #473, opened 2022-09-15
Add multi-type parameter support to the UI

kenxu95:
In order to get this working on the UI, I had to send up the parameter spec to /refresh_workflow. This resulted in some changes across both the UI and SDK.

Enforcement of parameter types happens at runtime. We discussed doing it in the clients, but for consistency chose to do it all in the orchestration for now.

Demo: https://www.loom.com/share/3362f7585953404995bd82c99aab5043

@vsreekanti to view the demo.

ENG-1565


sdk/aqueduct/api_client.py
```diff
@@ -371,16 +371,19 @@ def _construct_register_workflow_request(
...
+            parameter specs, which include the base64-encoded parameter value and the
+            parameter serialization type.
+        """
         headers = self._generate_auth_headers()
         url = self.construct_full_url(self.REFRESH_WORKFLOW_ROUTE_TEMPLATE % flow_id)
 
-        body = {}
-        if serialized_params is not None:
-            body["parameters"] = serialized_params
+        body = {"parameters": serialized_param_specs}
```
kenxu95, line 384: The format of the body is - a dictionary from parameter name to json-serialized parameter spec. This parameter spec includes a base64-encoded & serialized parameter value, along with the serialization type.

Because of this, the backend needs to unwrap twice to get all the information - once to get the parameter spec, and once to get the parameter value in the spec.

src/ui/common/src/components/workflows/workflowHeader.tsx
```diff
@@ -89,17 +91,45 @@ const WorkflowHeader: React.FC<Props> = ({ user, workflowDag }) => {
...
+        })
+      }
+    })
+    return serializedParams
+  }
+
   const triggerWorkflowRun = () => {
-    const parameters = new FormData();
-    parameters.append('parameters', JSON.stringify(paramNameToValMap));
+    // const parameters = new FormData();
```
agiron123, line 121: remove these commented out lines if you don't need them anymore 🙂

src/ui/common/src/components/workflows/workflowHeader.tsx
```diff
@@ -89,17 +91,45 @@ const WorkflowHeader: React.FC<Props> = ({ user, workflowDag }) => {
...
+    let serializedParams = {}
+    Object.entries(paramNameToValMap).forEach(([key, strVal]) => {
+      // Serialize the user's input string appropriately into base64. The input can either be a
+      // 1) number 2) string 3) json.
+      try {
+        let val = JSON.parse(strVal)
+
+        // All jsonable values are serialized as json.
+        serializedParams[key] = JSON.stringify({
+          val: btoa(JSON.stringify(val)),
```
agiron123, line 106: could you just use strval here instead of having to stringify val back again?

```typescript
val: btoa(strVal)
```

agiron123 (approved):
FE code here looks good on my end!

Nice demo!

src/ui/common/src/components/workflows/workflowHeader.tsx
```diff
@@ -89,17 +91,45 @@ const WorkflowHeader: React.FC<Props> = ({ user, workflowDag }) => {
...
+          serialization_type: SerializationType.Json,
+        })
```
kenxu95, line 108: What matters is just that the serialization type matches the serialization of the value. In the SDK, we serialize JSON strings with SERIALIZATION_TYPE.STRING, but here we can json serialize JSON strings! As long as the serialization type matches, the backend can always infer the correct type.

The reason is because the SDK and executor use `infer_artifact_type()`, and then use this inferred type to index into a map. However, the frontend is different because the input is always a string, so really the value is either JSONable or not. The user can still enter a raw json string eg `"{\"hello\": \"world\"}"` and things will work just fine.

cw75:
@kenxu95 Demo looks good! What if the param type when registering the workflow is non-json serializable types like bytes, images or picklable? I think then we should disable parameter customization from the UI for now right?

kenxu95:
Oh I don't think thats possible right now from the UI. If its not JSONable, then its just a string. I guess that means you can't customize those types of parameters, yes. Do you want me to grey those out (or remove them) in the dialog for now?

cw75:
@kenxu95 I think we should show those parameters but grey them out with a message saying we don't support customizing these yet. Also if the parameter type is say image, what default value are we showing on the UI's input field?

vsreekanti:
@kenxu95, this looks great. A few things:
* I agree with CGWu. Let's grey out the field and add a message underneath saying that we don't support uploading those types from the UI. 
* Why does the error message say that it's unexpected and ask the user to create an issue? It seems quite expected. 
* What happens if I give a bad type altogether -- e.g, I enter `{"another key"}`?

kenxu95:
@vsreekanti yeah I can easily get rid of the UNKNOWN error cruft. It looks like this now:
image

In the case of {"another key"}, it'll parse it as a string, since it's not jsonable. Is that ok? I feel like its a bit hard for us to figure out a user's intention of writing a json string, but just doing it wrong. Type enforcement should catch it if the parameter is supposed to be a json type (dict, num, bool), but you just made a typo. It will not catch it if the parameter was initially defined as a string, but that seems fine.

vsreekanti:
* New error message looks great. We have a task [for this](https://linear.app/aqueducthq/issue/ENG-1480/fix-bug-where-we-spuriously-call-an-error-unexpected) on the backlog. Does your fix generalize to that? 
* Parsing as a string is fine. I agree that it's hard for us to tell. The only thing I was wondering is if we could pass some metadata (i.e., we know the backend expects a type `dict`, so we enforce that? It's not a huge deal, but we might want to throw it on the backlog for the future.

kenxu95:
This is what the greyed-out field will look like. Is that alright? I added some helper text at the bottom.

image

vsreekanti:
Looks good. Tiny nit: Let's just drop the `n/a` and leave it blank.

kenxu95:
> New error message looks great. We have a task [for this](https://linear.app/aqueducthq/issue/ENG-1480/fix-bug-where-we-spuriously-call-an-error-unexpected) on the backlog. Does your fix generalize to that?

I don't think this PR or that tasks makes any general changes. We have a general way of surfacing errors messages from python operators and marking them as user or system already, and I think both these tasks just utilize that.

> Parsing as a string is fine. I agree that it's hard for us to tell. The only thing I was wondering is if we could pass some metadata (i.e., we know the backend expects a type dict, so we enforce that? It's not a huge deal, but we might want to throw it on the backlog for the future.

Yeah, Chengang and I were talking about whether to enforce this earlier on the clients, or do to it on the backend. For the sake of consistency across clients, we figured the best thing to do now is to have dumb clients and enforce all the parameter typing stuff during orchestration. I created this [task](https://linear.app/aqueducthq/issue/ENG-1684/catch-parameter-type-changes-before-orchestration) a few days ago to track this.

kenxu95:
@cw75 

> Also if the parameter type is say image, what default value are we showing on the UI's input field?

If I include the default value as a placeholder, we have this non-sensical output.
image

However, we're getting rid of the placeholder field for non-customizable parameters so this is no longer an issue.

src/ui/common/src/components/workflows/workflowHeader.tsx
kenxu95: yep, good catch. Fixed.

PR #475, opened 2022-09-15
Add migrator for serialization type in param operators

HarisChoudhary:
We introduce a db migrator to make sure that persisted param operator information correctly has the serialization_type field.


src/golang/cmd/migrator/versions/000018_add_serialization_value_to_param_op/types.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"database/sql/driver"
+
+	"github.com/aqueducthq/aqueduct/lib/collections/utils"
+)
+
+type Spec struct {
+	Type  string
+	Param map[string]string
```
kenxu95, line 11: so I understand correctly, even though Spec.Param is defined as a struct with two string fields, we can declare it to be read like a map?

kenxu95 (approved):
Can you test after rebasing on https://github.com/aqueducthq/aqueduct/pull/473? I imagine you'd create a parameter and publish it with the old version, then upgrade and trigger the flow again with both default and custom parameters?

src/golang/cmd/migrator/versions/000018_add_serialization_value_to_param_op/types.go
HarisChoudhary: Yeah, the database we are working on will hold the spec as a struct like the one below
`{"Type":"param","Param":{"serialization_type":"json","val":"\"Aqueduct\""}}`
where there will be a key for "type" and a key for spec based on the type. that will hold the spec and is another map corresponding to the number of variables in the Param Spec on the service side. so in the migrator we just ensure the spec struct has the correct variable names. In this case we only care about the type key (string) and if the key is param (map[string]string).

src/golang/cmd/migrator/versions/000018_add_serialization_value_to_param_op/queries.go
```diff
@@ -0,0 +1,69 @@
...
+	err := db.Query(ctx, &response, query)
+	return response, err
+}
+
+func updateParamOperatorWithNewSpec(
+	ctx context.Context,
+	operator Operator,
+	db database.Database,
+) error {
+	// Before the migration all param were json serializable only so we can assign the type as json to all
```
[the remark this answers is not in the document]
cw75: @HarisChoudhary not really. Just because we required the value to be json serializable doesn't mean the underlying value should now correspond to artifact type or serialization type `JSON`. I implemented logic to infer the types in my migration step 16 for artifact results, so maybe you can borrow the logic there? Worst case we might need to trigger some Python jobs to run the `infer_artifact_type` function to figure out the exact type.

cw75 (commented):
@HarisChoudhary In addition to my comment inline, I think when @kenxu95 finishes the "passing param value via file" part, we also need to migrate existing param spec's value to a path.

src/golang/cmd/migrator/versions/000018_add_serialization_value_to_param_op/queries.go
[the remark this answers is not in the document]
HarisChoudhary: Got it, thanks for the clarification. I'll adjust this code accordingly

HarisChoudhary:
> @HarisChoudhary In addition to my comment inline, I think when @kenxu95 finishes the "passing param value via file" part, we also need to migrate existing param spec's value to a path.

Got it thanks for the heads up, I'll work on that as part of this check-in as well.

kenxu95:
@HarisChoudhary I think the scope of this PR should be kept the same. Once I finish the other one, we can do a separate migration then - I don't think we need to shove it all in one migration.

kenxu95:
Actually, we decided we're not even going to change the param spec for now, and will wait for the Aqueduct 2.0 refactor where we separate our in-memory representations from the db completely.

kenxu95 (commented):
I think we need to change the way the parameter value is serialized in the database too right? The current value is the result of `json.dumps()`, but the new one is the base64 encoding of the serialized value. I don't see a migration of the parameter value field here.

kenxu95 (changes requested):
See above comment.

HarisChoudhary:
After the migrator is run, the "spec" column for a param operator changes from:
`{"type":"param","param":{"val":"\"kiropo\""}}`

to

`{"type":"param","param":{"serialization_type":"string","val":"Imtpcm9wbyI="}}`

src/golang/cmd/migrator/versions/000019_add_serialization_value_to_param_op/utils.go
```diff
@@ -0,0 +1,110 @@
...
+	db database.Database,
+) ([]Operator, error) {
+	query := "SELECT id, spec FROM operator;"
+
+	var response []Operator
+	err := db.Query(ctx, &response, query)
+	return response, err
+}
+
+func updateParamOperatorWithNewSpec(
```
cw75, line 42: Is this function idempotent? If some steps fail and we re-execute the migration script, will the already-migrated rows be affected? I think probably this is not idempotent now because we also tweaked the spec val?

src/golang/cmd/migrator/versions/000019_add_serialization_value_to_param_op/utils.go
```diff
@@ -0,0 +1,110 @@
...
+	db database.Database,
+) error {
+	param_val := operator.OpSpec.Param[value_key]
+	// Launch the Python job to infer the type of the parameter value
+	cmd := exec.Command(
+		"python3",
+		"-m",
+		fmt.Sprintf("%s.%s", pythonExecutorPackage, typeInferencePythonPath),
+		"--val",
+		param_val,
```
cw75, line 54: I assume this is fine because we are assuming previous parameter values won't be very big, so passing it directly as cmd will work?

src/golang/cmd/migrator/versions/000019_add_serialization_value_to_param_op/utils.go
```diff
@@ -0,0 +1,110 @@
...
+		return err
+	}
+
+	param_type := strings.TrimSuffix(outb.String(), "\n")
+	operator.OpSpec.Param[serilization_type_key] = param_type
+
+	// We also change the param value to be a base64 encoding
+	operator.OpSpec.Param[serilization_type_key] = param_type
+
+	base64val := base64.StdEncoding.EncodeToString([]byte(param_val))
```
cw75, line 74: I think `param_val` is a json-serialized string, so I'm not sure whether directly converting it to byte and then base64 encoding to a string works. Maybe we should let the conversion happen entirely in the Python script: the script first do `json.loads(param_val)`, and then do a base64 encode. And then the Golang side picks up this value and store back into the db.

src/golang/cmd/migrator/versions/000019_add_serialization_value_to_param_op/utils.go
```diff
@@ -0,0 +1,110 @@
...
+func updateParamOperatorWithOldSpec(
+	ctx context.Context,
+	operator Operator,
+	db database.Database,
+) error {
+	// If we want to migrate back to old version we delete the serialization type in the spec
+	delete(operator.OpSpec.Param, serilization_type_key)
+	// Decode base64 value back to string
+	decodedVal, _ := base64.StdEncoding.DecodeString(operator.OpSpec.Param[value_key])
+	operator.OpSpec.Param[value_key] = string(decodedVal)
```
cw75, line 98: I'm not sure if `string(decodedVal)` restores the value back to its original json-serialized form. If so this will imply utf-8 encoding and json serialization are the same, which I don't think it's true? Worst case we may need to do this reverse conversion on the Python side as well.

src/python/bin/aqueduct
```diff
@@ -20,7 +20,7 @@ import yaml
 from packaging.version import parse as parse_version
 from tqdm import tqdm
 
-SCHEMA_VERSION = "18"
+SCHEMA_VERSION = "19"
```
cw75, line 23: We also need to change this on the install local script.

cw75 (commented):
@HarisChoudhary @kenxu95 left some comments inline. Thx!

src/golang/cmd/migrator/versions/000019_add_serialization_value_to_param_op/utils.go
HarisChoudhary: yes, the val was passed as part of the spec and had to be accepted via cmd so it should preserve the property

kenxu95: I think that's a reasonable assumption if we don't expect to ever downgrade and then re-upgrade this version right?
Otherwise, I wonder if it'll hang on the re-upgrade part, if they had large parameters in the newer version.

cw75: Nah let's not worry about the re-upgrade part. :)

src/golang/cmd/migrator/versions/000019_add_serialization_value_to_param_op/utils.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	fmt.Println(fmt.Sprintf("WE GOT %s", outb.String()))
+	outputs := strings.Split(outb.String(), "\n")
+	param_type := outputs[0]
+	param_val = outputs[1]
+	operator.OpSpec.Param[serilization_type_key] = param_type
+
+	// We also change the param value to be a base64 encoding
+	operator.OpSpec.Param[value_key] = param_val[2:len(param_val)-1]
```
cw75, line 80: @HarisChoudhary could you make a comment why we are doing `[2:len(param_val)-1]`? In general I think it's cleaner to have the Python side write the output to a file, which is then read by the Golang side (like in migration 16) instead of trying to parse the output to stdout. But if this is too much work and the current way works, I'm fine with keeping it as is for this time.

HarisChoudhary: @cw75 I didnt mean for this to get reviews yet despite the push, my other implementation doesnt have this part

src/golang/cmd/migrator/migrator/register.go
```diff
@@ -136,4 +137,10 @@ func init() {
 		downPostgres: _000018.DownPostgres,
 		name:         "add exec state to workflow_dag_result",
 	}
+
+	registeredMigrations[19] = &migration{
+		upPostgres: _000019.Up, upSqlite: _000019.Up,
+		downPostgres: _000019.Down,
+		name:         "add serialization value to param op",
```
cw75, line 144: should this be serialization type?

src/python/aqueduct_executor/migrators/parameter_val_type_inference_000019/execute.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def run_type_check_and_encode(json_val: str) -> None:
+    """
+    Infers the type for a given param value and print to std out
+    Encode to base 64 and print out as well
+    """
+    val = json.loads(json_val)
+    artifact_type = infer_artifact_type(val)
+    encoded_val = base64.b64encode(json.dumps(val).encode("utf8"))
```
cw75, line 17: This is not used right?

src/python/aqueduct_executor/migrators/parameter_val_type_inference_000019/execute.py
```diff
@@ -0,0 +1,30 @@
...
+    serialization_type = serialize.artifact_type_to_serialization_type(artifact_type, val)
+    print(artifact_type.value)
```
cw75, line 19: I thought we are capturing the serialization type?

src/python/aqueduct_executor/migrators/parameter_val_type_inference_000019/execute.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def run_decode(val: str, val_type: str) -> None:
+    """
+    Decodes from base64 and prints out the value
+    """
+    decoded_val = base64.b64decode(val)
+    serialization_type = SerializationType(val_type)
+    deserialized_val = serialize.deserialization_function_mapping[serialization_type](decoded_val)
+    print(deserialized_val)
```
cw75, line 30: Shouldn't we output a json-serialized version of this to restore its original form?

cw75 (changes requested):

src/golang/cmd/migrator/migrator/register.go
```diff
@@ -136,4 +137,10 @@ func init() {
 		downPostgres: _000018.DownPostgres,
 		name:         "add exec state to workflow_dag_result",
 	}
+
+	registeredMigrations[19] = &migration{
+		upPostgres: _000019.Up, upSqlite: _000019.Up,
+		downPostgres: _000019.Down,
+		name:         "add serialization type value to param op",
```
cw75, line 144: do you mean "type and value"?

src/python/aqueduct_executor/migrators/parameter_val_type_inference_000019/execute.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def run_decode(val: str, val_type: str) -> None:
+    """
+    Decodes from base64 and prints out the value
+    """
+    decoded_val = base64.b64decode(val)
+    serialization_type = SerializationType(val_type)
+    deserialized_val = serialize.deserialization_function_mapping[serialization_type](decoded_val)
+    loaded_val = json.loads(deserialized_val)
```
cw75, line 29: why do we need this step?

cw75 (commented):
Left a couple more questions. @HarisChoudhary

cw75 (approved):
LGTM!

PR #478, opened 2022-09-15
Adds support for updating Airflow workflows

saurav-c:
This PR allows users to update the metadata and DAG structure of a workflow running on Airflow.
This requires the user to copy over the updated Airflow DAG file to their server.

ENG 1423

TEST: https://www.loom.com/share/dacedfcd9a864003a158a45096f1d3a2

### DIFF
--- a/src/golang/cmd/server/handler/register_airflow_workflow.go
+++ b/src/golang/cmd/server/handler/register_airflow_workflow.go
@@ -6,8 +6,14 @@ import (
 
 	"github.com/aqueducthq/aqueduct/cmd/server/request"
 	"github.com/aqueducthq/aqueduct/lib/airflow"
+	"github.com/aqueducthq/aqueduct/lib/collections/artifact_result"
+	"github.com/aqueducthq/aqueduct/lib/collections/notification"
+	"github.com/aqueducthq/aqueduct/lib/collections/operator_result"
+	"github.com/aqueducthq/aqueduct/lib/collections/user"
+	"github.com/aqueducthq/aqueduct/lib/collections/workflow"
 	"github.com/aqueducthq/aqueduct/lib/collections/workflow_dag"
 	"github.com/aqueducthq/aqueduct/lib/collections/workflow_dag_edge"
+	"github.com/aqueducthq/aqueduct/lib/collections/workflow_dag_result"
 	aq_context "github.com/aqueducthq/aqueduct/lib/context"
 	"github.com/aqueducthq/aqueduct/lib/database"
 	dag_utils "github.com/aqueducthq/aqueduct/lib/workflow/dag"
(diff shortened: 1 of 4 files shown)


src/golang/cmd/server/handler/register_airflow_workflow.go
```diff
@@ -120,6 +137,45 @@ func (h *RegisterAirflowWorkflowHandler) Perform(ctx context.Context, interfaceA
...
+			h.OperatorReader,
+			h.ArtifactReader,
+			h.WorkflowDagEdgeReader,
+			h.Database,
+		)
+		if err != nil {
+			return emptyResp, http.StatusInternalServerError, errors.Wrap(err, "Unable to update workflow.")
+		}
+
+		// NOTE (saurav): This is not perfect because if there are any in progress Airflow DAGRuns, those will
```
saurav-c, line 156: TODO: Create linear task for this

saurav-c:
cc @vsreekanti for demo

src/golang/cmd/server/handler/register_airflow_workflow.go
```diff
@@ -179,22 +256,26 @@ func (h *RegisterAirflowWorkflowHandler) Perform(ctx context.Context, interfaceA
...
+			return emptyResp, http.StatusInternalServerError, errors.Wrap(err, "Unable to add user who created the workflow to watch.")
+		}
 	}
 
 	return &registerAirflowWorkflowResponse{
-		Id:   workflowId,
-		File: string(airflowFile),
+		Id:       workflowId,
+		File:     string(airflowFile),
+		IsUpdate: args.isUpdate,
```
likawind, line 279: why the new field is required in response?

src/golang/cmd/server/handler/register_airflow_workflow.go
likawind: I'm curious if this is caused by us restricting sync to finished runs here: https://github.com/aqueducthq/aqueduct/blob/main/src/golang/lib/airflow/writer.go#L28 , or something else?

Does removing that restriction solve the problem without any additional concerns?

likawind (approved):
A few clarifying questions. Otherwise I think the PR looks good!

likawind:
Friendly note about test failure. It looks like some race condition on workflow deletion but not 100% sure (@kenxu95 )

vsreekanti:
@saurav-c, this looks great. One question that you didn't cover in the video: What do we show on the UI (if anything) when Aqueduct has a new DAG structure but Airflow has done no runs of that DAG structure? It would be nice if we could show a warning similar to what you printed out on the SDK creation step, letting the user know that they need to take action if they haven't already.

src/golang/cmd/server/handler/register_airflow_workflow.go
saurav-c: For the SDK, if its an update then we can tell the user that if they don't update the Airflow DAG file, new runs won't sync over properly.

saurav-c:
@vsreekanti It's hard for us to detect whether the current Airflow DAG file matches the Aqueduct DAG structure, so I'm not sure when we would remove the warning.

src/golang/cmd/server/handler/register_airflow_workflow.go
saurav-c: It will actually add more complexity, because we then need to check all workflow dag results to see if there are any with a status of in progress and then try to update their status on each subsequent sync call.

likawind: Ahh understood. Then agreed what we have now is good enough

vsreekanti:
I think for now we could remove the warning once the DAG execution has finished. If it's finished and we can match the Airflow DAG to ours, great. Until then, show the warning. What do you think?

saurav-c:
I'm working on the changes discussed with @vsreekanti above as part of https://linear.app/aqueducthq/issue/ENG-1705/add-ui-warning-if-user-has-not-yet-updated-airflow-dag-file, so I'm merging this PR for now.

PR #479, opened 2022-09-15
Fix bug where the parameter value is too large to be passed to the python operators

kenxu95:

During dag construction, we write the parameter value to a path and pass that path in through the spec. The parameter operator now expects that the output path has already been populated, and relegates itself to simply writing metadata and enforcing types.

This solution does not rely on any DB changes, but it does mean that for every run of a flow, we will write the parameter values to storage before calling orchestration.

@saurav-c to validate that this will work with airflow.

ENG-1565

### DIFF
--- a/src/golang/lib/engine/aq_engine.go
+++ b/src/golang/lib/engine/aq_engine.go
@@ -12,6 +12,7 @@ import (
 	"github.com/aqueducthq/aqueduct/lib/collections/integration"
 	"github.com/aqueducthq/aqueduct/lib/collections/notification"
 	operator_db "github.com/aqueducthq/aqueduct/lib/collections/operator"
+	"github.com/aqueducthq/aqueduct/lib/collections/operator/param"
 	"github.com/aqueducthq/aqueduct/lib/collections/operator_result"
 	"github.com/aqueducthq/aqueduct/lib/collections/shared"
 	"github.com/aqueducthq/aqueduct/lib/collections/user"
@@ -169,7 +170,7 @@ func (eng *aqEngine) ExecuteWorkflow(
 	ctx context.Context,
 	workflowId uuid.UUID,
 	timeConfig *AqueductTimeConfig,
-	parameters map[string]string,
+	parameters map[string]param.Param,
 ) (shared.ExecutionStatus, error) {
 	dbWorkflowDag, err := workflow_utils.ReadLatestWorkflowDagFromDatabase(
 		ctx,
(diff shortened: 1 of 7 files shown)


kenxu95:
@cw75 @HarisChoudhary I'd like to write an integration test to validate that this fixes the "parameter-value-too-large" problem. Do either of you have the image data that was causing this issue? Any idea how big is too big?

HarisChoudhary:
> @cw75 @HarisChoudhary I'd like to write an integration test to validate that this fixes the "parameter-value-too-large" problem. Do either of you have the image data that was causing this issue? Any idea how big is too big?

The two image i used to discover the issue. One is a smaller resized version that should work, the larger one will error out.

wss-aqueduc
![aqimg
t-roman-france_0](https://user-images.githubusercontent.com/6863633/190524341-4e91daf5-4fe2-45c1-8a28-323b64ac59fb.jpeg)

kenxu95:
cool - I kinda think that instead of adding this to the repo, we can have this in the S3 bucket that we can pull from in our SDK integration tests. https://linear.app/aqueducthq/issue/ENG-1635/build-s3-support-into-our-sdk-integration-tests

Can add it as a note to that task. I'd love to get to this soon actually - there's a bunch of improvements we can make to our suite.

HarisChoudhary:
> cool - I kinda think that instead of adding this to the repo, we can have this in the S3 bucket that we can pull from in our SDK integration tests. https://linear.app/aqueducthq/issue/ENG-1635/build-s3-support-into-our-sdk-integration-tests
> 
> Can add it as a note to that task. I'd love to get to this soon actually - there's a bunch of improvements we can make to our suite.

I dont think we need to test with the image specifically, when passed into the executor its a base64 encoded byte buffer of the image which comes out to a fairly large string. If we can just add a string of that fairly large size then we can replicated this in test without the image needed.

src/golang/lib/engine/aq_engine.go
```diff
@@ -659,7 +656,7 @@ func (eng *aqEngine) TriggerWorkflow(
 	workflowId uuid.UUID,
 	name string,
 	timeConfig *AqueductTimeConfig,
-	parameters map[string]param.Param,
+	parameters map[string]string,
```
HarisChoudhary, line 659: I don't think this will compile

HarisChoudhary (changes requested):
Seems like maybe some changes were on local that didnt get reflected on the pr but the current state wont work as is i think

src/golang/lib/engine/aq_engine.go
kenxu95: oh yes I think I checked out the wrong branch's file here

saurav-c (commented):
This looks compatible with Airflow

HarisChoudhary (approved):

PR #480, opened 2022-09-15
Fixes Airflow syncing bug with API response limit

saurav-c:
This PR fixes a bug where if there were more runs than the default Airflow API limit, they would never be synced with Aqueduct. This PR ensures that all DAG runs get synced from Airflow into Aqueduct.

It doesn't do anything to improve the performance of syncing since that would be more involved and can be implemented for a later release.

ENG 1531

### DIFF
--- a/src/golang/lib/airflow/client.go
+++ b/src/golang/lib/airflow/client.go
@@ -39,15 +39,34 @@ func newClient(ctx context.Context, authConf auth.Config) (*client, error) {
 
 // getDagRuns returns all of the Airflow DAGRuns for the Airflow DAG specified.
 func (c *client) getDagRuns(dagId string) ([]airflow.DAGRun, error) {
-	dagRunsResp, resp, err := c.apiClient.DAGRunApi.GetDagRuns(
-		c.ctx,
-		dagId,
-	).OrderBy("start_date").Execute()
-	if err != nil {
-		return nil, wrapApiError(err, "GetDagRuns", resp)
+	limit := 100 // This is the max number of DAG runs that can be returned in each response.
+	offset := 0
+	var dagRuns []airflow.DAGRun
+
+	// Keep paginating through DAG runs until there are none in response
+	for {
+		dagRunsResp, resp, err := c.apiClient.DAGRunApi.GetDagRuns(
+			c.ctx,
+			dagId,
+		).
+			OrderBy("start_date").
+			Limit(int32(limit)).
+			Offset(int32(offset)).
+			Execute()
+		if err != nil {
+			return nil, wrapApiError(err, "GetDagRuns", resp)
+		}
+
+		if len(*dagRunsResp.DagRuns) == 0 {
+			// There are no more DAG Runs
+			break
+		}
+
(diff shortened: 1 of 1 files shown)


src/golang/lib/airflow/client.go
```diff
@@ -39,15 +39,34 @@ func newClient(ctx context.Context, authConf auth.Config) (*client, error) {
 
 // getDagRuns returns all of the Airflow DAGRuns for the Airflow DAG specified.
 func (c *client) getDagRuns(dagId string) ([]airflow.DAGRun, error) {
-	dagRunsResp, resp, err := c.apiClient.DAGRunApi.GetDagRuns(
-		c.ctx,
-		dagId,
-	).OrderBy("start_date").Execute()
-	if err != nil {
-		return nil, wrapApiError(err, "GetDagRuns", resp)
+	limit := 100 // This is the max number of DAG runs that can be returned in each response.
```
likawind, line 42: nit: consider `limitPerPage`. I thought this is the total limit until I read the rest of the code

src/golang/lib/airflow/client.go
```diff
@@ -39,15 +39,34 @@ func newClient(ctx context.Context, authConf auth.Config) (*client, error) {
...
+			return nil, wrapApiError(err, "GetDagRuns", resp)
+		}
+
+		if len(*dagRunsResp.DagRuns) == 0 {
+			// There are no more DAG Runs
+			break
+		}
+
+		dagRuns = append(dagRuns, *dagRunsResp.DagRuns...)
+		offset += limit
```
likawind, line 66: should this be `len(*dagRunsResp.DagRuns)`? I'm curious if it only has say 57 runs and the next API call will try to fetch with offset 100, will this cause any issue?

src/golang/lib/airflow/client.go
```diff
@@ -39,15 +39,34 @@ func newClient(ctx context.Context, authConf auth.Config) (*client, error) {
...
+	offset := 0
+	var dagRuns []airflow.DAGRun
+
+	// Keep paginating through DAG runs until there are none in response
+	for {
+		dagRunsResp, resp, err := c.apiClient.DAGRunApi.GetDagRuns(
+			c.ctx,
+			dagId,
+		).
+			OrderBy("start_date").
```
likawind, line 52: off topic - this looks really cool, idk how hard it is for us to implement something similar. Though this might be backed by graphQL or ORM when it gets to query management, and we don't want to do either for now

likawind (approved):
Have we tested it? I assume the bottleneck being the size limit for a single response, rather than, for example, the size of data sent to a single client per sec?

src/golang/lib/airflow/client.go
saurav-c: Since we are doing this in order, if there are less than 100 runs returned, this means there are only less than 100 runs returned. The next fetch will return 0 runs.

I tested this with a DAG that had 9 runs. The first fetch returned 9 and the next one returned 0.

saurav-c:
@likawind Yup, I tested it against some actual Airflow DAGs.

PR #481, opened 2022-09-16
Deletes extracted function file for all engines and execution status

saurav-c:
This PR ensures that the extracted function file is deleted for all engine types regardless of the execution status.
Previously, the file would not be deleted for Airflow execution if there was some failure.

ENG 1496

## Test
- Tested manually that the fn file was deleted for an Airflow workflow and an Aqueduct workflow


likawind (approved):
This is because airflow only run the `execute` python function?

likawind (changes requested):
Actually I don't get why the `_cleanup(spec)` is only called during exception handling. Should it be called in a `Finally` block?

likawind (approved):
LGTM!

PR #483, opened 2022-09-17
Eng 1570 add endpoint to fetch historical data given an artf ID

likawind:
This PR adds a route `api/artifact/{artfId}/results` to list all historical results given an artifact ID. It reuses `RawResultResponse` object introduced in #425 . For small artifacts like `bool`, `string` and `numeric`, we also include the artifacts' value as response.

This route will be immediately used to fetch metrics / check history for UI redesign.

ENG-1570

Verified in postman, together with a new integration test.
```
{
    "results": [
        {
            "id": "c1e5514d-ab0c-47c2-897f-1886084d553b",
            "content_path": "5a71b1e8-0a8f-4085-b070-e2eb7fee07be",
            "content_serialized": "true",
            "exec_state": {
                "user_logs": {
                    "stdout": "",
                    "stderr": ""
                },
                "status": "succeeded",
                "failure_type": null,
                "error": null,
                "timestamps": null
            }
        },
        {
            "id": "5109ce64-a44a-41c5-8f24-a481c6cca800",
            "content_path": "bde0a196-f454-45f2-8f73-4065bd632ab2",
            "content_serialized": "true",
            "exec_state": {
                "user_logs": {
                    "stdout": "",
                    "stderr": ""
                },
                "status": "succeeded",
                "failure_type": null,
                "error": null,
                "timestamps": null
            }
        },
        {
            "id": "6618d3d2-c5c3-4940-a7cd-fc609cea0059",
            "content_path": "64f36922-4434-49b6-b1f8-1f62b04b35c3",
            "content_serialized": "true",
            "exec_state": {
                "user_logs": {
                    "stdout": "",
                    "stderr": ""
                },
                "status": "succeeded",
                "failure_type": null,
                "error": null,
                "timestamps": null
            }
        },
        {
            "id": "7a44cd6f-0632-4b98-806c-8bac5455203d",
            "content_path": "a5cc4391-dc05-44e1-a908-ad931268ca4e",
            "content_serialized": "true",
            "exec_state": {
                "user_logs": {
                    "stdout": "",
                    "stderr": ""
                },
                "status": "succeeded",
                "failure_type": null,
                "error": null,
                "timestamps": null
            }
        },
        ...
   ]
}
```


src/golang/lib/collections/artifact/types.go
```diff
@@ -15,3 +15,11 @@ const (
 	Image     Type = "image"
 	Picklable Type = "picklable"
 )
+
+// `IsCompact` indicates if the value is 'small' enough to pass around in-memory.
```
likawind, line 19: Continuing the discussion in #470 , for now I decided to determine if value is 'small' only on type. The motivation is that if we need to assume the content with certain type could be large, we shouldn't even assume the content would fit in memory and thus there's no point to read it in memory and determine based on size of value.

In future, we could expose more `storage` API to return the content size, or persist size in DB, to optimize the behavior here.

agiron123 (approved):
Nice work!

src/golang/cmd/server/handler/list_artifact_results.go
```diff
@@ -0,0 +1,146 @@
...
+		resultIds = append(resultIds, result.Id)
+	}
+
+	dbDagsByResultIds, err := h.WorkflowDagReader.GetWorkflowDagsMapByArtifactResultIds(ctx, resultIds, h.Database)
+	if err != nil {
+		return emptyResponse, http.StatusInternalServerError, errors.Wrap(err, "Unable to retrieve workflow dags.")
+	}
+
+	// maps from db dag Ids
+	dbDagsMapByDagIds := make(map[uuid.UUID]workflow_dag.DBWorkflowDag, len(dbDagsByResultIds))
```
kenxu95, line 102: nit on naming: I don't think you need the `Map` phrase since it's implicit in the "<something>By<something>" pattern, and I think it should use singular terminology since its one dag id per one dag, so this would be `dbDagByDagID`. I've been using that convention throughout the codebase so it might be good to settle on a consistent pattern?

Same for all the other maps here.

src/golang/cmd/server/handler/list_artifact_results.go
```diff
@@ -0,0 +1,146 @@
...
+// Params: None
+// Request:
+//	Headers:
+//		`api-key`: user's API Key
+// Response:
+//	Body:
+//		serialized `listArtifactResultsResponse`
+
+type listArtifactResultsResponse struct {
+	Results []artifact.RawResultResponse `json:"results"`
```
kenxu95, line 32: Can you note what order these results are returned in? Is it last executed first?

kenxu95 (commented):
Makes sense to me, a couple clarifying questions:
1) I see the timestamps are null in the example, is this currently expected, and we'll fix them up in the future?
2) This only tracks history for the same artifact on the same DAG right? When do we update an artifact's id, is it when some operator in the DAG changes, or when this specific artifact changes? I guess what I'm asking is will an unrelated update somewhere else in the DAG break the history for all artifacts.
3) Just curious, how hard would it be to change it to track history through the artifact name - does that just involve crafting a different SQL query?

src/golang/cmd/server/handler/list_artifact_results.go
likawind: there's no order assumption here, we should handle that on client side

kenxu95: how does the client order it? Using the timestamps field in ExecState?

likawind:
@kenxu95 thanks for ideas! I think it's a good question on how to track the version update today, let's discuss in eng sync offline.

Ideally I think we need some version-free ID to track changes across versions. For now if name is good enough I think we should just change to name, which is not very hard (we just need to change artf results from fetch by ID to fetch by name)

likawind:
For null timestamp, these are from artifacts created in older version of workflows. We could migrate these data if necessary but I don't think it's high-pri for now. Newly created wf after #452 will have these timestamps, as reflected in integration tests

likawind:
Oh and also for #1, this version tracks the same artifact from potentially different dag as we get all results by artf ID.

src/golang/cmd/server/handler/list_artifact_results.go
likawind: yes

likawind:
@kenxu95 updated the query to bound both on artifact name and workflowId

src/golang/cmd/server/handler/list_artifact_results.go
kenxu95: Can we just add a comment here to that effect?

src/golang/cmd/server/handler/list_artifact_results.go
```diff
@@ -0,0 +1,155 @@
...
+		resultIds = append(resultIds, result.Id)
+	}
+
+	dbDagsByResultIds, err := h.WorkflowDagReader.GetWorkflowDagsMapByArtifactResultIds(ctx, resultIds, h.Database)
+	if err != nil {
+		return emptyResponse, http.StatusInternalServerError, errors.Wrap(err, "Unable to retrieve workflow dags.")
+	}
+
+	// maps from db dag Ids
+	dbDagsByDagIds := make(map[uuid.UUID]workflow_dag.DBWorkflowDag, len(dbDagsByResultIds))
```
kenxu95, line 111: oh I meant like I usually use singular terminology for one-to-one mappings, so `dbDagByDagId`. But up to you, not a huge deal

kenxu95 (approved):

PR #487, opened 2022-09-20
Adds UI warning for out of sync Airflow workflow

saurav-c:
This PR adds a warning on the workflow viewing page if there is a potential out of sync conflict between the DAG registered with Airflow vs the latest Aqueduct dag. For now we just tell the user that they must update the Airflow DAG file.

We can add support later for a user to download the most recent Airflow DAG file from the UI in case they misplaced the one returned by the SDK.

This PR also invokes workflow syncing when the `GET` `/workflow` route is invoked. Previously syncing only occurred during the LIST workflows route.

ENG 1705

This is what the UI shows when DAGs are out of sync.
Screen Shot 2022-09-20 at 11 01 00 AM


saurav-c:
cc @vsreekanti for UI changes

src/ui/common/src/components/workflows/workflowHeader.tsx
```diff
@@ -93,6 +94,20 @@ const WorkflowHeader: React.FC<Props> = ({ user, workflowDag, workflowId }) => {
     );
   }
 
+  let airflowUpdateWarning = null;
+  if (workflowDag.engine_config.type == EngineType.Airflow 
+    && !workflowDag.engine_config.airflow_config?.matches_airflow) {
+      airflowUpdateWarning = (
+        <Box>
```
saurav-c, line 101: @agiron123 @vsreekanti What's the best way to make this box's width smaller. It takes up almost the entire page right now.

vsreekanti: I would just set a max width in pixels on the box.

vsreekanti:
Is it just a coincidence that the whole workflow is marked as failed in your screenshot, @saurav-c? The warning looks good to me.

saurav-c:
@vsreekanti Yup, the failure is a coincident

src/ui/common/src/components/workflows/workflowHeader.tsx
agiron123: Yeah, something like maxWidth="800px" would be good to have here. @saurav-c

src/ui/common/src/components/workflows/workflowHeader.tsx
```diff
@@ -254,6 +271,8 @@ const WorkflowHeader: React.FC<Props> = ({ user, workflowDag, workflowId }) => {
         {runWorkflowDialog}
       </Box>
 
+      {airflowUpdateWarning}
```
agiron123, line 274: small nit here, I prefer to conditionally render things in a case like this, so we know when the warning will show from reading the render function.

You can do something like:
```
const showAirflowUpdateWarning = workflowDag.engine_config.type == EngineType.Airflow &&
    !workflowDag.engine_config.airflow_config?.matches_airflow;
```

Then, inside render:
```
{showAirflowUpdateWarning && airflowUpdateWarning}
```

agiron123 (approved):
Frontend stuff looks good. Left a small comment, but that can be resolved before you merge in.

cw75 (approved):

PR #506, opened 2022-09-22
Eng 1359 fix a bug where tables with integer

kenxu95:
We cannot support non-string column names for dataframes due to json-serialization (and also parquet if we chose to migrate), so if we detect such cases during serialization, we throw an error.

When you try to do something like this, you'll get:
image

In order to implement this nicely, we introduce the concept of an `ExecFailureException`, which can be translated to a failed `ExecutionState` at the top level. This allows us to not have to pass an exec_state around all the time to catch user errors. In this case, we had to throw an exception within the serialization helpers. It would have been somewhat to have to pass an exec_state all the way down.

In order to keep the two types of ExecutionState paths separate, we support the following pattern now, in addition to our `@user_fn_redirected` pattern:
```
raise ExecFailureException(...)
...
except ExecFailureException as e:
     write(ExecutionState.from_exception(e, user_logs=...))
```

@vsreekanti to inspect the error message.

ENG-1359

### DIFF
--- a/src/python/aqueduct_executor/operators/function_executor/execute.py
+++ b/src/python/aqueduct_executor/operators/function_executor/execute.py
@@ -186,9 +186,7 @@ def run(spec: FunctionSpec) -> None:
     """
     print("Started %s job: %s" % (spec.type, spec.name))
 
-    # TODO:
     exec_state = ExecutionState(user_logs=Logs())
-
     storage = parse_storage(spec.storage_config)
     try:
         validate_spec(spec)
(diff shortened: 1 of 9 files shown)


vsreekanti:
This looks good to me. My only very slight concern is if the wording might make people think that all column types have to be strings? But I can't come up with anything clearer off the top of my head, so I'm happy with it as is.

kenxu95:
I exactly copied the error that we get when trying to serialize a dataframe `to_parquet()`. I know we use json-serialization now, but there was some discussion about moving to a better serialization format like parquet in the future. So I figured if it's a constraint there, we might as well adopt it now and see if anyone complains.

src/python/aqueduct_executor/operators/function_executor/execute.py
```diff
@@ -237,18 +228,19 @@ def run(spec: FunctionSpec) -> None:
                 # Cast np.bool_ to a bool.
                 result = bool(result)
             else:
-                handle_type_error_and_exit(spec, storage, exec_state, TIP_NOT_BOOL)
+                raise ExecFailureException(
```
likawind, line 231: I don't know if this is crucial, but it seems that we will lose all user printed logs for such failures?

likawind (approved):
Thanks for changing the pattern. My only question is to confirm with @vsreekanti that losing user logs under failures not caused by user function is acceptable, which I think so.

vsreekanti:
It's not ideal, but it's fine for now.

src/python/aqueduct_executor/operators/function_executor/execute.py
kenxu95: We won't lose any user logs. The except clause will add the user logs to the exec_state before writing. `exec_state.from_exception()` forces you to add user logs.

PR #507, opened 2022-09-22
Eng 1710 use new workflow result route for metrics and artifact page.

likawind:
This PR routes metrics detail page to use new API routes

ENG-1710


src/ui/common/src/components/pages/workflow/id/index.tsx
```diff
@@ -419,7 +419,7 @@ const WorkflowPage: React.FC<WorkflowPageProps> = ({
                 <Typography variant="h5" padding="16px">
                   {getNodeLabel()}
                 </Typography>
-                <Box sx={{ mx: 2, alignSelf: 'center', marginLeft: 'auto' }}>
```
eunice-chan, line 422: How does `marginLeft: auto` affect the look of the page?

src/ui/common/src/reducers/workflowDagResults.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+const initialState: WorkflowDagResultsState = { results: {} };
+
+export const workflowDagResultsSlice = createSlice({
+  name: 'workflowDagResultsReducer',
+  initialState: initialState,
+  reducers: {},
+  extraReducers: (builder) => {
+    builder.addCase(handleGetWorkflowDagResult.pending, (state, { meta }) => {
```
eunice-chan, line 25: We always add a pending, fulfilled, and rejected case for each handler. I wonder if it is worth it to create a helper function that creates these cases for us & puts the loading status to loading/succeeded/failed for us so all we need to do is provide function inputs of what else we would want to do in those cases.

src/ui/common/src/components/pages/metric/id/index.tsx
```diff
@@ -50,109 +42,68 @@ const MetricDetailsPage: React.FC<MetricDetailsPageProps> = ({
...
-      nodeIcon: tableArtifactNodeIcon,
-    },
-  ];
+  useEffect(() => {
+    if (!!operator) {
+      document.title = `${operator.name} | Aqueduct`;
+    }
+  }, [operator]);
 
   const artifactTypeToIconMapping = {
```
eunice-chan, line 108: I think it would be helpful if you could move this out to a common file so other details page can use this mapping.

src/ui/common/src/components/pages/metric/id/index.tsx
```diff
@@ -176,164 +127,171 @@ const MetricDetailsPage: React.FC<MetricDetailsPageProps> = ({
...
+              color: 'rgba(0,0,0,0.54)',
+            }}
+          >
+            <FontAwesomeIcon
+              icon={artifactTypeToIconMapping[artifactResult.type]}
+            />
+          </Box>
+          <Link
+            to={`${getPathPrefix()}/workflow/${workflowId}/result/${workflowDagResultId}/artifact/${artifactId}`}
+            component={RouterLink as any}
```
eunice-chan, line 218: Why do we need "as any"?

src/ui/common/src/components/pages/metric/id/index.tsx
```diff
@@ -50,109 +42,68 @@ const MetricDetailsPage: React.FC<MetricDetailsPageProps> = ({
...
       );
     }
-  }, [workflow.dagResults]);
+  }, []);
 
   useEffect(() => {
-    if (!metricResult || !metricResult.result) {
-      return;
+    // Load artifact history once workflow dag results finished loading
+    // and the result is not cached
```
eunice-chan, line 84: What does "and the result is not cached" mean? Is it referring to the fact we query the api each time workflowDagResultWithLoadingStatus or artifactId is changed?

src/ui/common/src/components/pages/metric/id/index.tsx
```diff
@@ -176,164 +127,171 @@ const MetricDetailsPage: React.FC<MetricDetailsPageProps> = ({
...
-          return (
-            <Box>
-              <Typography variant="body1">
-                {artifactResult.result.data}
-              </Typography>
-            </Box>
-          );
-        }
-      }
+    // TODO: Check the serialization_type of the artifacts and show a link
```
eunice-chan, line 160: I'm assuming we plan on dealing with these TODOs in a later PR? Maybe we should create tasks for these in Linear

eunice-chan (approved):
I don't think there are many changes other than incorporating the API & I think adding some alerts, but would appreciate small demos for the changes. I don't think there's anything glaringly bad and since we want to get the first draft all merged to ui-redesign and iteratively improve, I'm approving it for now.

src/ui/common/src/reducers/workflowDagResults.ts
likawind: That's a good idea. The design wouldn't be trivial though since these pattern applies to different state object, and we require some level of customization to control these behaviors. It may require some work to get the typing right. Feel free to take a look at the current cases and add your suggestions!

src/ui/common/src/components/pages/metric/id/index.tsx
likawind: Defer this to @agiron123 who added the original version. I assume it's to bypass some type checking

src/ui/common/src/components/pages/metric/id/index.tsx
likawind: redux store can also serve as a cache. If `artifactHistoryWithLoadingStatus` is not `undefined`, that means there's an entry in entry in `artifactResultsReducer` for the given `artifactId`. In this case we don't have to fetch the result again. Otherwise the result is missing and we dispatch the handler.

And yes, this caching help us to not call this API every time the wfDagResult is changed

PR #512, opened 2022-09-27
Clean up metrics artf detail pages

likawind:
This PR cleans up artifact and metrics page. By managing large components individually, we reduced `index` file of each page from ~300+ lines to ~150 lines. Some components are also reusable. For example, `artifact/summaryList` can be reused for inputs / outputs for operator details page as well.

ENG-1747

## Demo (if any)
Verified these refactored pages works properly.
Screen Shot 2022-09-26 at 11 51 49 PM
Screen Shot 2022-09-26 at 11 52 37 PM

### DIFF
--- a/src/ui/common/src/components/pages/metric/id/index.tsx
+++ b/src/ui/common/src/components/pages/metric/id/index.tsx
@@ -2,7 +2,7 @@ import { CircularProgress } from '@mui/material';
 import Alert from '@mui/material/Alert';
 import Box from '@mui/material/Box';
 import Typography from '@mui/material/Typography';
-import React, { useEffect, useState } from 'react';
+import React, { useEffect } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { useParams } from 'react-router-dom';
 
@@ -28,9 +28,6 @@ const MetricDetailsPage: React.FC<MetricDetailsPageProps> = ({
 }) => {
   const dispatch: AppDispatch = useDispatch();
   const { workflowId, workflowDagResultId, metricOperatorId } = useParams();
-
-  const [inputsExpanded, setInputsExpanded] = useState<boolean>(true);
-  const [outputsExpanded, setOutputsExpanded] = useState<boolean>(true);
 
   const workflowDagResultWithLoadingStatus = useSelector(
     (state: RootState) =>
@@ -89,12 +86,6 @@ const MetricDetailsPage: React.FC<MetricDetailsPageProps> = ({
     }
   }, [operator]);
 
-  const listStyle = {
-    width: '100%',
-    maxWidth: 360,
-    bgcolor: 'background.paper',
-  };
-
   if (
     !workflowDagResultWithLoadingStatus ||
     isInitial(workflowDagResultWithLoadingStatus.status) ||
@@ -117,18 +108,17 @@ const MetricDetailsPage: React.FC<MetricDetailsPageProps> = ({
     );
   }
 
-  const inputs = operator.inputs
-    .map(
-      (artifactId) =>
-        (workflowDagResultWithLoadingStatus.result?.artifacts ?? {})[artifactId]
-    )
-    .filter((artf) => !!artf);
-  const outputs = operator.outputs
-    .map(
-      (artifactId) =>
-        (workflowDagResultWithLoadingStatus.result?.artifacts ?? {})[artifactId]
-    )
-    .filter((artf) => !!artf);
+  const mapArtifacts = (artfIds: string[]) =>
+    artfIds
+      .map(
+        (artifactId) =>
+          (workflowDagResultWithLoadingStatus.result?.artifacts ?? {})[
+            artifactId
+          ]
+      )
+      .filter((artf) => !!artf);
+  const inputs = mapArtifacts(operator.inputs);
+  const outputs = mapArtifacts(operator.outputs);
 
   return (
     <Layout user={user}>
--- a/src/ui/common/src/components/pages/metric/id/index.tsx
+++ b/src/ui/common/src/components/pages/metric/id/index.tsx
@@ -1,27 +1,19 @@
-import { faChevronRight } from '@fortawesome/free-solid-svg-icons';
-import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import { CircularProgress, Link, List, ListItem } from '@mui/material';
-import Accordion from '@mui/material/Accordion';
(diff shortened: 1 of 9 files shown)


src/ui/common/src/components/pages/metric/id/index.tsx
```diff
@@ -125,134 +117,18 @@ const MetricDetailsPage: React.FC<MetricDetailsPageProps> = ({
...
-              marker: { color: 'red' },
-            },
-          ]}
-          layout={{ width: '100%', height: '100%' }}
-        />
-        <PaginatedTable data={historicalData} />
-      </Box>
-    );
-  }
+  const inputs = operator.inputs
```
eunice-chan, line 120: We can make a helper function for the inputs and outputs variables.
```
const filterArtifacts = (artifacts) => {
artifacts
    .map(
      (artifactId) =>
        (workflowDagResultWithLoadingStatus.result?.artifacts ?? {})[artifactId]
    )
    .filter((artf) => !!artf);

const inputs = filterArtifacts(operator.inputs);
const outputs = filterArtifacts(operator.outputs);
};

**Review by eunice-chan — APPROVED** [review]
Thanks for component-ifying everything! Looks very neat now

File: src/ui/common/src/components/workflows/artifact/content.tsx [hunk 2]
```diff
@@ -0,0 +1,110 @@
...
+}) => {
+  if (!artifact.result) {
+    return (
+      <Typography variant="h5" component="div" marginBottom="8px">
+        No result to show for this artifact.
+      </Typography>
+    );
+  }
+
+  if (artifact.result.content_serialized != null) {
```
agiron123, line 30: !==

src/ui/common/src/components/workflows/artifact/content.tsx
```diff
@@ -0,0 +1,110 @@
...
+        <Alert title="Cannot parse table data.">
+          {err}
+          {contentWithLoadingStatus.data}
+        </Alert>
+      );
+    }
+  }
+  // TODO: handle images here
+  return (
+    <Typography variant="body1" component="div" marginBottom="8px">
```
agiron123, line 104: Is there a reason why we're using component="div" in all the typography components here? Haven't seen that being used in too many other places.

src/ui/common/src/components/workflows/artifact/metric/history.tsx
```diff
@@ -0,0 +1,76 @@
...
+      <Alert title="Failed to load historical data.">
+        {historyWithLoadingStatus.status.err}
+      </Alert>
+    );
+  }
+
+  const historicalData: Data = {
+    schema: {
+      fields: [
+        { name: 'status', type: 'varchar' },
```
agiron123, line 35: maybe we should make this fields array into a constant.

agiron123: something like const metricsHistoricalSchema could be a good name.

src/ui/common/src/components/workflows/artifact/metric/history.tsx
```diff
@@ -0,0 +1,76 @@
...
+  return (
+    <Box display="flex" justifyContent="center" flexDirection="column">
+      <Plot
+        data={[
+          {
+            x: timestamps,
+            y: values,
+            type: 'scatter',
+            mode: 'lines+markers',
+            marker: { color: 'red' },
```
agiron123, line 66: See if we can use the same dark blue color that we're using throughout our app here. Might be a nice touch 🧑‍🎨

agiron123 (approved):
Nice work! Excited to see these changes go in.

src/ui/common/src/components/workflows/artifact/content.tsx
likawind: let me add a comment, this is intentional as it makes sure the field is neither `null` nor `undefined`

src/ui/common/src/components/workflows/artifact/content.tsx
likawind: I don't know this is copied from your original version of artf detail page ;)

src/ui/common/src/components/workflows/artifact/metric/history.tsx
likawind: which string should I put here?

vsreekanti:
This may be out of scope for this PR, but we should do a quick design review on these new pages. There's some rough edges that can be ironed out. e.g., there shouldn't be colons in headings, and it's not clear why these two cards are different heights or have dividers underneath them:
 
image

Similarly the height of the data table should be adaptive so when there's less data, there shouldn't be dead space at the bottom: 
image

likawind:
@vsreekanti yes, these pages are expected in more rough state for now. Once all pages are ready (which should happen soon) we will ask feedbacks from the team and polish these places altogether

agiron123:
@likawind there's a number that you can set here for rows per page. If it's less than 10, we should use that number. Currently the minimum number of entries per page is set to 5, but that's something that we can adjust.

likawind:
Ahh figured. Though I do agree with Vikram that the size should be adaptable if there are only 5 or 10 rows (now I also need to scroll for just 10 rows which may not be ideal as well).

Let's revisit this as a separate problem, I think we need to play with the component a bit more

likawind:
https://linear.app/aqueducthq/issue/ENG-1752/make-data-tables-size-more-adaptive

likawind:
https://linear.app/aqueducthq/issue/ENG-1751/refinements-for-input-output-list

src/ui/common/src/components/workflows/artifact/metric/history.tsx
```diff
@@ -0,0 +1,80 @@
...
+  return (
+    <Box display="flex" justifyContent="center" flexDirection="column">
+      <Plot
+        data={[
+          {
+            x: timestamps,
+            y: values,
+            type: 'scatter',
+            mode: 'lines+markers',
+            // colors are 'blue.900'. Plot doesn't seem to accept color from theme.
```
likawind, line 68: @agiron123 I don't think plotly like `blue.900` so I put the hex here. lmk if there are better ways to solve this

agiron123: Yeah, import the theme object and use template literals. This is how we normally deal with react components that do not have the sx prop or our theme context.

```
color: `${theme.blue900}`
```
or however it should be indexed.

PR #514, opened 2022-09-27
Allow user functions to have multiple output artifacts

kenxu95:
The user can not specify whether they want an @op to return multiple output artifacts, by using `@op(num_outputs)`.

@cw75 as lead reviewer.

ENG-455

## Loom demo (if any)
https://www.loom.com/share/1277c3c87b5a4394bea08728d8e9455f

[EDIT] the bug shown in the video is fixed by this: https://github.com/aqueducthq/aqueduct/pull/515

### DIFF
--- a/integration_tests/sdk/multiple_output_test.py
+++ b/integration_tests/sdk/multiple_output_test.py
@@ -1,5 +1,4 @@
 import pytest
-
 from aqueduct.error import AqueductError
 from utils import run_flow_test
 
--- a/integration_tests/sdk/multiple_output_test.py
+++ b/integration_tests/sdk/multiple_output_test.py
@@ -1,0 +1,35 @@
+from aqueduct import op
+from utils import run_flow_test
+
+
+def test_multiple_outputs(client):
+    @op(num_outputs=2)
+    def generate_two_outputs():
+        return "hello", 1234
+
+    @op
+    def append_to_str(input_str):
+        return input_str + " world."
+
+    @op
+    def double_number(num):
+        return 2 * num
+
+    str_artifact, int_artifact = generate_two_outputs()
+    assert str_artifact.get() == "hello"
+    assert int_artifact.get() == 1234
+
+    str_output = append_to_str(str_artifact)
+    int_output = double_number(int_artifact)
+    assert str_output.get() == "hello world."
+    assert int_output.get() == 2468
+
+    run_flow_test(client, artifacts=[str_output, int_output], delete_flow_after=False)
+
+
+def test_multiple_outputs_user_failure(client):
+    @op(num_outputs=3)
+    def generate_two_outputs():
+        return "hello", 1234
+
+    generate_two_outputs()
--- a/integration_tests/sdk/multiple_output_test.py
+++ b/integration_tests/sdk/multiple_output_test.py
@@ -1,3 +1,6 @@
+import pytest
+
+from aqueduct.error import AqueductError
 from utils import run_flow_test
 
 from aqueduct import op
@@ -33,4 +36,5 @@ def test_multiple_outputs_user_failure(client):
     def generate_two_outputs():
         return "hello", 1234
 
-    generate_two_outputs()
+    with pytest.raises(AqueductError):
+        generate_two_outputs()
--- a/integration_tests/sdk/multiple_output_test.py
+++ b/integration_tests/sdk/multiple_output_test.py
@@ -1,5 +1,6 @@
-from aqueduct import op
 from utils import run_flow_test
+
+from aqueduct import op
 
 
 def test_multiple_outputs(client):
--- a/src/golang/lib/workflow/dag/workflow_dag.go
+++ b/src/golang/lib/workflow/dag/workflow_dag.go
@@ -2,6 +2,7 @@ package dag
 
 import (
 	"context"
+
 	db_artifact "github.com/aqueducthq/aqueduct/lib/collections/artifact"
 	"github.com/aqueducthq/aqueduct/lib/collections/artifact_result"
 	"github.com/aqueducthq/aqueduct/lib/collections/notification"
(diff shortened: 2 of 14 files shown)


src/golang/lib/workflow/dag/workflow_dag.go
```diff
@@ -164,14 +152,41 @@ func NewWorkflowDag(
...
+
+		for _, inputArtifactID := range dbOperator.Inputs {
+			opIDs, ok := opIDsByInputArtifact[inputArtifactID]
+			if !ok {
+				opIDs = make([]uuid.UUID, 0, 1)
+			}
+			opIDsByInputArtifact[inputArtifactID] = append(opIDs, dbOperator.Id)
+		}
+	}
+
```
kenxu95, line 172: The `workflow_dag` constructor is getting a little unwieldy. I think we need to do a little follow-up refactor to think through the minimum set of maps that we need, like we do in the SDK's dag representation.

vsreekanti:
Very cool! The demo makes sense to me. This does make me think that we should add functionality to allow users to customize artifact names:
```python
artf = fn()
artf.set_name('featurized data')
```

src/golang/lib/workflow/dag/workflow_dag.go
```diff
@@ -64,29 +64,20 @@ type workflowDagImpl struct {
 // by the artifact's original ID.
 func computeArtifactSignatures(
 	dbOperators map[uuid.UUID]db_operator.DBOperator,
-	dbArtifacts map[uuid.UUID]db_artifact.DBArtifact,
+	opIDsByInputArtifact map[uuid.UUID][]uuid.UUID,
```
likawind, line 67: Curious what's the motivation here?

kenxu95: Shoot I'm having a hard time recalling the exact motivation, since this was done a while back. I think I wanted to move the construction of all the dag maps up into the constructor, instead of the helpers, in an effort to centralize all the data structures we needed in one place, so that future refactors would be easier.

likawind: lol sg, as long as it works let's worry about that later. Two questions just for this change -
* does `numArtifacts` the same as the size of map `opIDsByInputArtifact`?
* Does this change only modify the interface but not the behavior in terms of how things get hashed?

src/golang/lib/workflow/dag/workflow_dag.go
```diff
@@ -111,33 +102,31 @@ func computeArtifactSignatures(
 			bytesToHash = append(bytesToHash, []byte(currOp.Spec.Param().Val)...)
 		}
 
-		if len(currOp.Outputs) > 1 {
-			return nil, errors.Newf("Multiple output artifacts is currently unsupported.")
-		}
-
 		// Compute that signature for each output artifact.
 		// The assumption is that there is only one output.
```
cw75, line 106: remove this comment.

src/golang/lib/workflow/dag/workflow_dag.go
```diff
@@ -111,33 +102,31 @@ func computeArtifactSignatures(
...
-		for _, nextOpID := range opIDsByInputArtifact[outputArtifactID] {
-			nextOp := dbOperators[nextOpID]
-
-			depsComputed := true
-			for _, inputArtifactID := range nextOp.Inputs {
-				if _, ok := processedArtifactIds[inputArtifactID]; !ok {
-					depsComputed = false
-					break
+		for _, outputArtifactID := range currOp.Outputs {
+			bytesToHash = append(bytesToHash, []byte(outputArtifactID.String())...)
```
cw75, line 108: maybe doesn't affect correctness, but every loop I think we should be appending each output artifact id to the `bytesToHash` in line 102. In the current implementation for example the second output artifact's signature seems to contain the first output artifact's id, which is a bit weird.

integration_tests/sdk/multiple_output_test.py
```diff
@@ -0,0 +1,39 @@
...
+
+    run_flow_test(client, artifacts=[str_output, int_output], delete_flow_after=False)
+
+
+def test_multiple_outputs_user_failure(client):
+    @op(num_outputs=3)
+    def generate_two_outputs():
+        return "hello", 1234
+
+    with pytest.raises(AqueductError):
```
cw75, line 38: I think based on the implementation this should raise a `InvalidUserArgumentException`? Not sure why it worked.

cw75 (approved):
Some nits inline but LGTM!

src/golang/lib/workflow/dag/workflow_dag.go
kenxu95: That map is expected to have <= numArtifacts. Terminal artifacts are not included, since they do not have any operators operating on them. The reason this works is because it is legal to loop over `opIDsByInputArtifact[terminalArtifactID]` with range, even if no entry exists, as it will simply not enter the for loop.

This does not change the behavior of how things get hashed at all, the signature of an artifact is still `hash(input artifact signatures +  Optional[parameter value string] + output artifact id])`, it's just been extended to handle multiple output artifact ids.

integration_tests/sdk/multiple_output_test.py
kenxu95: It should be failing at preview execution, which leads to an AqueductError. Where in the code did you expect `InvalidUserArgumentException` to be thrown? We can't check the number of outputs until we actually run the code on the backend.

src/golang/lib/workflow/dag/workflow_dag.go
kenxu95: oh yeah you're totally right - will fix.

kenxu95:
Task for `set_name()`: https://linear.app/aqueducthq/issue/ENG-1761/allow-users-to-customize-artifact-names

integration_tests/sdk/multiple_output_test.py
cw75: Oh I misread the code. Ignore this.

PR #515, opened 2022-09-27
Fix bug where tuples were being mishandled

kenxu95:
Tuples and lists are serialized into the same json string representation. To prevent confusion, we'll need to include the `artifact_type` in the deserialization process, so that we can cast tuple types appropriately before returning.

This allows us to continue viewing these objects on the UI without too much work.

Adds `deserialize()` as the method everyone should call to deserialize bytes into their python objects.

ENG-1748


kenxu95:
On the UI, a tuple still looks like a list. I don't know if that matters at all. We can have the UI detect such cases and convert them into "(...)", but I don't know if that's important.

image

cw75 (approved):

PR #517, opened 2022-09-27
Fixes issue where Parameter and Check operators were marked as "Cancelled" incorrectly

kenxu95:
@likawind as lead reviewer. @cw75 tagged to keep in the loop, since we previously discussed writing to the parameter output path in the golang orchestrator instead of the python operator.
@vsreekanti and @agiron123 for any product-related issues/questions about the coloring of nodes.

It used to be that, if an artifact is failed, we'd mark it as cancelled. However, this shouldn't necessarily be the case for checks and parameters. Parameters that fail type checking should be marked failed, as should checks that return false.

The solution proposed here is to decide whether an artifact is cancelled based on whether it has been "computed". Whether an artifact is "computed" simply means whether the artifact's content has been written to storage. This can happen even if the operator fails (once again, see parameters and checks).

This makes it so that failed check and parameter artifacts can be marked as red, instead of grey. This is especially noticeable for parameters - see the following parameterized flow who's parameter has failed a type check.

Before:
Screen Shot 2022-09-27 at 3 36 59 PM

After:
Screen Shot 2022-09-27 at 3 52 23 PM

ENG-1715

### DIFF
--- a/src/golang/lib/workflow/artifact/artifact.go
+++ b/src/golang/lib/workflow/artifact/artifact.go
@@ -33,11 +33,11 @@ type Artifact interface {
 	// Finish is an end-of-lifecycle hook meant to do any final cleanup work.
 	Finish(ctx context.Context)
 
-	// Computed indicates whether this artifact has been computed or not.
-	// An artifact is only considered "computed" if its results have been written to storage.
+	// Computed indicates whether this artifact's contents have been computed or not.
+	// An artifact is only considered "computed" if its content has been written to storage.
 	// This is *NOT* the same as having the operator's execution state == SUCCEEDED. For example,
-	// for check operators, the artifact could have been computed even if the check operator did not
-	// pass (returned false).
+	// for check operators, the artifact is computed even if the check operator does not pass
+	// (returned false).
 	Computed(ctx context.Context) bool
 
 	// GetMetadata fetches the metadata for this artifact.
@@ -126,11 +126,11 @@ func (a *ArtifactImpl) Name() string {
 }
 
 func (a *ArtifactImpl) Computed(ctx context.Context) bool {
-	// An artifact is only considered computed if its results have been written.
+	// An artifact is only considered computed if its contents have been written.
 	res := utils.ObjectExistsInStorage(
 		ctx,
 		a.storageConfig,
-		a.execPaths.ArtifactMetadataPath,
+		a.execPaths.ArtifactContentPath,
 	)
 	return res
 }
--- a/src/golang/lib/workflow/operator/base.go
+++ b/src/golang/lib/workflow/operator/base.go
@@ -283,6 +283,8 @@ func (bo *baseOperator) PersistResult(ctx context.Context) error {
 		// operator either never ran or did run but hit a user-generated exception.
 		// System-generated errors from things like the type checking of parameters will
 		// still generate downstream artifacts, so those will continue to be marked as "failed".
+		// Invariant: if an artifact is marked as failed, it's operator must also be marked failed,
+		// with the same error message and context.
 		artifactExecState := *execState
 		if !outputArtifact.Computed(ctx) {
 			artifactExecState.Status = shared.CanceledExecutionStatus
--- a/src/golang/lib/workflow/operator/base.go
+++ b/src/golang/lib/workflow/operator/base.go
@@ -279,10 +279,12 @@ func (bo *baseOperator) PersistResult(ctx context.Context) error {
 	)
 
 	for _, outputArtifact := range bo.outputs {
-		// If the operator failed, for now, we assume the downstream artifact is never generated.
-		// In this case, we will mark these artifacts as cancelled.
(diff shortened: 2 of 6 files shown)


vsreekanti:
@kenxu95, this mostly looks good to me. The only thing that I find a little confusing is that there are now two errors that ostensibly say the same thing. "Error creating num" and "Error executing num". Neither really make sense from the user's perspective -- `num` is a parameter that's definitely not executed and not even really created? At a minimum, I think we should only have one error, and I'd like to maybe rewrite the content to make it clearer that it's not about execution. Thoughts?

src/golang/lib/workflow/artifact/artifact.go
```diff
@@ -126,11 +126,11 @@ func (a *ArtifactImpl) Name() string {
 }
 
 func (a *ArtifactImpl) Computed(ctx context.Context) bool {
-	// An artifact is only considered computed if its results have been written.
+	// An artifact is only considered computed if its contents have been written.
 	res := utils.ObjectExistsInStorage(
 		ctx,
 		a.storageConfig,
-		a.execPaths.ArtifactMetadataPath,
```
likawind, line 133: are we still using `metadataPath`?

likawind (approved):
Looks great!

likawind:
I agree with @vsreekanti 's suggestions, would let you decide if we should address in a separate PR.

src/golang/lib/workflow/operator/base.go
```diff
@@ -279,10 +279,12 @@ func (bo *baseOperator) PersistResult(ctx context.Context) error {
...
-		// If the operator failed, for now, we assume the downstream artifact is never generated.
-		// In this case, we will mark these artifacts as cancelled.
+		// If the downstream artifact was never generated, we mark it as "cancelled", since the
+		// operator either never ran or did run but hit a user-generated exception.
+		// System-generated errors from things like the type checking of parameters will
+		// still generate downstream artifacts, so those will continue to be marked as "failed".
 		artifactExecState := *execState
-		if execState.Status == shared.FailedExecutionStatus {
+		if !outputArtifact.Computed(ctx) {
 			artifactExecState.Status = shared.CanceledExecutionStatus
```
likawind, line 288: to hide duplicated error messages, we could copy the execState and set `artifactExecState.Error = nil` here

src/golang/lib/workflow/artifact/artifact.go
kenxu95: Yeah, it just won't be used to checking if artifact was "computed". The metadata path and output path used to be written in tandem, but in certain edge cases (eg. failed parameter operator), it's possible that the contents are written but no the metadata path.

src/golang/lib/workflow/operator/base.go
kenxu95: What duplicated error messages are we talking about?  We should definitely keep the "Error executing..." msg, but not the other one.

kenxu95: oh I see what you mean, hmm

kenxu95:
@vsreekanti  @andre it currently seems like we don't expect artifact's to have status==Failed? I wonder if the best way to fix the duplicated error message is by getting rid of this case: https://github.com/aqueducthq/aqueduct/blob/main/src/ui/common/src/components/workflows/StatusBar.tsx#L386

In such cases, the operator should have errored right? So we don't need this duplication. @agiron123 can you verify?

src/golang/lib/workflow/operator/base.go
kenxu95: answered in the comment below, lemme know what you think

kenxu95:
Fixed the duplication concern by removing the FAILED status case from artifact:

image

likawind:
I think there are two cases - 1) artifact canceled due to operator failure and 2) artifact failed as a result of op failure

For 1 I feel the canceled artf should not carry duplication of failure message. It should just appear as 'canceled'. We could probably provide some global context in UI for all canceled op/artf that certain failure result in cancellation.

For 2 I feel we could just assume the failure context is identical to upstream op. In this case I guess we should just remove the `Error` and `Log` field for artifact and read it from upstream operator when rendering. If we go this way, the removal can be done in backend and there should be minimum fix required in UI since it sees no duplicated messages.

kenxu95:
hmm I'm not completely sure that we can solve this completely on the backend. I don't think this particular fix needs to be blocked on this discussion though, so I'll make follow-up task and we can discuss when @likawind is back in office.

vsreekanti:
Agree that we don't need to fix the canceled state right now. However, would it be possible to tidy up the error message to change the "Error executing" text?

kenxu95:
@vsreekanti does a simple change to `Error with num` instead of  `Error executing`, just for failing parameters work? I think the rest of the text makes sense.

vsreekanti:
Yep, that would be perfect. It's just the headline that's a little confusing. The rest of the message makes sense.

PR #530, opened 2022-10-03
Provide better guidance when there is a version mismatch between SDK and server

kenxu95:
Does three things:
1) Throws an exception with appropriate message if SDK version != server version. Uses the server's get_version endpoint.

This is what happens when the `aqueduct-sdk` package is outdated.
image

2) Adds the server version number in the banner printed when starting the server.
3) Also fixes our logger configuration in the SDK. I noticed this was broken while debugging this task.

ENG-1537


sdk/aqueduct/api_client.py
```diff
@@ -209,46 +224,53 @@ def construct_full_url(self, route_suffix: str, use_https: Optional[bool] = None
...
-    def _test_url(self, url: str) -> None:
-        """Perform a get on the url with default headers, raising an error if anything goes wrong.
+        Raises:
+             ClientValidationError:
+                If the server cannot be found, or if there is a versioning mismatch between server and sdk.
 
-        We don't are about the value of the response, as long as the request succeeds.
         """
+        try:
+            v = self._get_version(use_https=use_https)
```
likawind, line 253: nit: from context should we rename this to `_get_server_version()` together with the variable name?

likawind (changes requested):
The code looks great! My main concern though is the message.

This particular task comes from a customer call (or maybe debugging with Will) where the user forgot to kill aqueduct server before they run `pip install`. In this way if we just ask the user to upgrade, they may become confused as it's likely something they've already done so.

So I feel we should provide more specifics on how to perform the upgrade. For example:
```
The SDK is outdated, it is using version %s, while the server is of version %s. Please update your `aqueduct-sdk` package to the appropriate version:
* run <command>
* If running within a Jupyter notebook, restart your jupyter kernel

The server is outdated ... , please update your server or downgrade your SDK
To update your server:
* kill any running server process.
* run <command>
* run aqueduct start (with `--expose` if necessary)
To downgrade your sdk:
...
```
I also feel there might be some risk to support SDK downgrade as we never test it. So not sure if we should just cut that from the message.

Curious about how others think as well @vsreekanti @cw75

vsreekanti:
I'm not super worried about this. If you think about an analog like Jupyter, it's very unlikely that a user would update Jupyter and then forget to restart the server. Also, if you run `pip3 install --upgrade aqueduct-ml`, we will automatically upgrade your SDK and server together, and if you don't restart the server, it will detect the version mismatch the next time you create an SDK client. This also feels like a safer check because the SDK user might not have control over server versions if there's a separate admin. 

One thing we might want to do in the future is notify the user on the UI if there's a new version released, so they can upgrade. We should also link to a guide on Gitbook for updating Aqueduct regardless.

@kenxu95, looking at your screenshot, there seem to be some spacing issues with missing spaces between words (likely because of a multiline string?).

kenxu95:
@vsreekanti yeah the screenshot is outdated - I also noticed that and fixed it in the code. I can make a task for the notification on the UI + link to gitbook.

Has the concern about the messages been addressed @likawind?

likawind (approved):

likawind:
yeah I still feel some more detailed guide would be helpful, and as you said gitbook would be a better place to put them.

kenxu95:
https://linear.app/aqueducthq/issue/ENG-1785/add-gitbook-guide-for-updating-aqueduct

PR #532, opened 2022-10-03
Accept both singular and multiple artifacts in publish_flow()

kenxu95:
Both `client.publish_flow(artifacts=[output])` and `client.publish_flow(artifacts=output)` are possible now. Both arguments cannot be set or else an exception is raised.

Edited the example notebooks to use the singular version where possible.

ENG-1735


vsreekanti:
@kenxu95, I'm concerned that including or dropping the `s` is a subtle difference. Would it make sense to just do `artifacts=output` or `artifacts=[output1, output2]`? That way, it's the same argument with a union-typed value.

kenxu95:
I was concerned [multiple]=[singular] would be confusing, but that does seem like the lesser of two evils. Can fix.

vsreekanti:
I agree. Maybe we should support all combinations? `artifact=[list]`, `artifacts=[list]`, `artifact=obj`, `artifacts=obj`. If a user sets both, we throw an error. Otherwise, we accept either.

vsreekanti:
Maybe that's too complicated...

kenxu95:
yeah that seems like too much tbh

kenxu95:
I wonder if leaving it like this is fine, but we throw a helpful error message if they put the wrong type in?

vsreekanti:
I'd prefer we just accept both under `artifacts`. Generally, if we can predict things that a user might do wrong, we should address those things rather than notify them to do it differently.

kenxu95:
sounds good, this PR was been updated with the change.

cw75 (approved):
LGTM!

PR #537, opened 2022-10-04
Eng 1322 add UI for showing user what service is used for metadata storage

eunice-chan:
Display the storage layer type.

Adding the integration name will be a bigger effort and we should make a new task for that if we want to pursue that: https://aqueducthq.slack.com/archives/C01U6RREC7M/p1664834954509439

Eng 1322

## Loom demo (if any)
image


agiron123:
Nice work. Curious why this is a dropdown. Can the user change the service that is being used from this UI?

If they can't, then this should probably be text.

eunice-chan:
@agiron123 Currently, we do not allow the user to change the service used by the storage layer but this is planned feature and has a task on linear. Since it is not currently an implemented feature, should I comment out the dropdown and replace it with a text for now?

src/ui/common/src/components/workflows/storageSelector.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { RootState } from '../../stores/store';
+import { StorageType, StorageTypeNames } from '../../utils/storage';
+
+export const StorageSelector: React.FC = () => {
+  const workflow = useSelector((state: RootState) => state.workflowReducer);
+  const dag = workflow.selectedDag;
+  let selected = 'file';
+  if (dag) {
+    selected = dag.storage_config.type;
+    console.log('selected dag', dag.storage_config.type);
```
saurav-c, line 18: nit: remove logs

saurav-c (approved):
LGTM, so for now we will just show the type of the storage instead of the config or integration name?

agiron123:
> @agiron123 Currently, we do not allow the user to change the service used by the storage layer but this is planned feature and has a task on linear. Since it is not currently an implemented feature, should I comment out the dropdown and replace it with a text for now?

@eunice-chan this is fine for now. Let's go with this design. I have a task to clean up the workflow header and make it more responsive, so I might look into using a text instead then. But for now, I think what you have should be fine.

src/ui/common/src/components/workflows/storageSelector.tsx
```diff
@@ -0,0 +1,59 @@
...
+  if (dag) {
+    selected = dag.storage_config.type;
+  }
+
+  const getMenuItems = () => {
+    return (
+      Object.values(StorageType) as Array<
+        typeof StorageType[keyof typeof StorageType]
+      >
+    ).map((r, _) => {
```
agiron123, line 25: I think you can just omit the second parameter of .map here.

Try doing:
```typescript
.map((r) => ...
```

agiron123: Could you also use a more descriptive name here please? Something like storageType should suffice.

src/ui/common/src/components/workflows/storageSelector.tsx
```diff
@@ -0,0 +1,59 @@
...
+  };
+
+  const menuItems = getMenuItems();
+
+  return (
+    <Box sx={{ display: 'flex', alignItems: 'center' }}>
+      Storage Type:{' '}
+      <FormControl disabled sx={{ m: 1, minWidth: 120 }} size="small">
+        <Select
+          sx={{ maxHeight: 50 }}
```
agiron123, line 47: Can you use a unit of 8 here like 48 or 56 please?

agiron123 (approved):

src/ui/common/src/components/workflows/storageSelector.tsx
eunice-chan: I used the same structure & styling as the VersionSelector (except I added `m: 1` for this component.) I've updated both to be units of 8 (maxHeight: 48).

It doesn't seem to have much of a change to the UI:
image

PR #539, opened 2022-10-04
Eng 1760 workflow version switching bug

agiron123:
This PR fixes a bug which caused the app to crash when the user switches workflow versions using the version dropdown in the workflow header.

ENG-1760

## Loom demo (if any)
https://www.loom.com/share/78f11760e7b34b2bbe205d67523d2707

### DIFF
--- a/src/ui/common/src/components/workflows/ReactFlowCanvas.tsx
+++ b/src/ui/common/src/components/workflows/ReactFlowCanvas.tsx
@@ -58,6 +58,7 @@ const ReactFlowCanvas: React.FC<ReactFlowCanvasProps> = ({
     const metricArtifactNodes = [];
 
     // first find all check operators.
+    console.log('checking dagPositionState.result');
     if (dagPositionState.result) {
       const nodes = dagPositionState.result.nodes;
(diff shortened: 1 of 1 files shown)


src/ui/common/src/components/workflows/ReactFlowCanvas.tsx
```diff
@@ -146,7 +160,12 @@ const ReactFlowCanvas: React.FC<ReactFlowCanvasProps> = ({
       });
     });
 
+    console.log('filtered nodes: ', filteredNodes);
```
agiron123, line 163: remove logs.

eunice-chan (approved):
Is the change just making the default value of enrichedNodes as []?

agiron123:
> Is the change just making the default value of enrichedNodes as []?

That's one part of it. The error that was being printed to console had to do with result being undefined, so I needed to add a null check to the array index for both the boolean and metric artifacts. This code is similar for both:

```
 const boolArtifactResult =
                  artifactResults[boolArtifactNodes[i].id]?.result?.data;
```

PR #540, opened 2022-10-04
Requires workflow executor to check for paused workflow execution

saurav-c:
This PR requires the workflow executor to acquire a read lock at the beginning of each workflow run execution. This allows us to pause new workflow runs if there are certain system maintenance jobs that need to be performed, such as metadata store migration (this work will be in a later PR).

The PR uses a `filelock` to perform synchronization across processes, since each `executor` is its running in its own process. Once we turn these processes into goroutines, we can trade the `filelock` for a generic `mutex`.

ENG 1779

### DIFF
--- a/src/golang/cmd/executor/executor/workflow.go
+++ b/src/golang/cmd/executor/executor/workflow.go
@@ -70,7 +70,12 @@ func (ex *WorkflowExecutor) Run(ctx context.Context) error {
 	if err := lock.RLock(); err != nil {
 		return err
 	}
-	defer lock.RUnlock()
+	defer func() {
+		unlockErr := lock.RUnlock()
+		if unlockErr != nil {
+			log.Errorf("Unexpected error when unlocking execution lock: %v", unlockErr)
+		}
+	}()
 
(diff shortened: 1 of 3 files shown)


src/golang/cmd/executor/executor/workflow.go
```diff
@@ -63,6 +64,14 @@ func NewWorkflowExecutor(spec *job.WorkflowSpec, base *BaseExecutor) (*WorkflowE
 }
 
 func (ex *WorkflowExecutor) Run(ctx context.Context) error {
+	// First, ensure that workflow execution is not paused
+	lock := filelock.New(engine.ExecutionLock)
+	// The following will block until the RLock can be acquired
+	if err := lock.RLock(); err != nil {
+		return err
+	}
+	defer lock.RUnlock()
```
saurav-c, line 73: This is safe even if say the process crashes before it reaches the `defer` statement. This is because the underlying filelock is automatically released when the process exits.

kenxu95 (approved):
This makes sense to me!

PR #542, opened 2022-10-04
Implement script to add new tables to demo db for the next release

cw75:
Tested that it works with both fresh install and package upgrade.

### DIFF
--- a/src/python/bin/aqueduct
+++ b/src/python/bin/aqueduct
@@ -229,7 +229,7 @@ def update_server_version():
             f.write(requests.get(s3_path).content)
         os.chmod(data_script_path, 0o755)
         subprocess.check_call([sys.executable, data_script_path])
-        print("Successfully added new tables to the demo db.")
+        print("Successfully added new tables to the demo db as part of release v0.1.")
 
 
 def update():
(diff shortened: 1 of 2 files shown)


src/python/bin/aqueduct
```diff
@@ -246,8 +263,6 @@ def update():
             with open(os.path.join(server_directory, "db/demo.db"), "wb") as f:
                 f.write(requests.get(os.path.join(s3_server_prefix, "db/demo.db")).content)
 
-            update_server_version()
```
cw75, line 249: realized we can consolidate this to line 282 below.

src/resources/demo_db/sqlite/v0.1/create_tables.py
```diff
@@ -0,0 +1,56 @@
+from os.path import join
```
cw75, line 1: This script is here for reference; It will be fetched from S3 during the update process.

likawind: In future, should each version with new table contains a `create_tables.py` in that version's directory?

src/python/bin/aqueduct
```diff
@@ -214,6 +218,19 @@ def update_server_version():
         ]
     )
 
+    if parse_version(current_version) < parse_version("0.1") and parse_version(package_version) >= parse_version("0.1"):
```
likawind, line 221: Probably fine for now, but we like need it to be a helper so that it's easier to add new tables in future.

likawind (approved):

src/python/bin/aqueduct
cw75: Yup when we have many of these, we can package each of them into a helper function called from this main function. I'll add a comment as a reminder.

PR #543, opened 2022-10-04
[UI redesign] Updates design of metrics page

vsreekanti:

* Makes the titles of the components more uniform and adds divider.
* Fixes margins on Plotly plot to remove dead space; adds light grey background.
* Fixes formatting of timestamp by converting to TS `Date` and using `toLocaleString`.
* Replaces table with success-color-coded divs -- we'll probably want to render this in a paged view in the future, but I prefer this presentation because I found the table headers confusing. Open to feedback on improvements here!

## Loom demo (if any)
image

### DIFF
--- a/src/ui/common/src/components/workflows/operator/summaryList.tsx
+++ b/src/ui/common/src/components/workflows/operator/summaryList.tsx
@@ -19,12 +19,6 @@ type Props = {
   initiallyExpanded: boolean;
 };
 
-const listStyle = {
-  width: '100%',
-  maxWidth: 360,
-  bgcolor: 'background.paper',
-};
-
 const SummaryList: React.FC<Props> = ({
   title,
   workflowId,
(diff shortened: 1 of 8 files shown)


agiron123:
Looks good! Is there a way you can format the dates so that there is a 0 in the front for single digit months like September?

e.g. 09/15/2022

The check symbol seems to drift a little when rendering Sept and October.

Might also be a good idea to move the check before the date.

src/ui/common/src/components/workflows/operator/summaryList.tsx
```diff
@@ -19,20 +19,13 @@ type Props = {
   initiallyExpanded: boolean;
 };
 
-const listStyle = {
-  width: '100%',
-  maxWidth: 360,
-  bgcolor: 'background.paper',
-};
-
 const SummaryList: React.FC<Props> = ({
```
agiron123, line 22: How similar are the artifact and operator SummaryList components?

Perhaps we should combine the two components into one if they are similar enough.

agiron123 (approved):

vsreekanti:
@agiron123, I moved the checkmark before the datestring: 
image

Unfortunately, Javascript doesn't have a simple datestring formatter, so we'd have to hack the string together ourselves, which I'd prefer to avoid doing in this PR.

PR #549, opened 2022-10-05
Implements storage migration

saurav-c:
This PR adds support for migrating all artifact result content to the new metadata store when a user connects a new storage integration and sets it as the metadata store.

ENG 1786


src/golang/lib/workflow/artifact/migrate.go
```diff
@@ -0,0 +1,101 @@
...
+		return err
+	}
+	defer database.TxnRollbackIgnoreErr(ctx, txn)
+
+	dags, err := dagReader.GetWorkflowDags(ctx, nil, txn)
+	if err != nil {
+		return err
+	}
+
+	for _, dag := range dags {
```
kenxu95, line 53: If the a migration fails in the middle, and we try to run it again, we'll be able to detect already-migrated dags by checking it's storage config column right? We can then skip migrating those dags.

src/golang/lib/workflow/artifact/migrate.go
```diff
@@ -0,0 +1,101 @@
...
+				return err
+			}
+
+			// For each artifact result, move the content from `oldStore` to `newStore`
+			for _, artifactResult := range artifactResults {
+				val, err := oldStore.Get(ctx, artifactResult.ContentPath)
+				if err != nil {
+					return err
+				}
+
```
kenxu95, line 76: What happens if something already exists at this content path? Will it simply overwrite whats there?

src/golang/lib/workflow/artifact/migrate.go
```diff
@@ -0,0 +1,101 @@
...
+		for _, artifact := range artifacts {
+			artifactResults, err := artifactResultReader.GetArtifactResultsByArtifactId(ctx, artifact.Id, txn)
+			if err != nil {
+				return err
+			}
+
+			// For each artifact result, move the content from `oldStore` to `newStore`
+			for _, artifactResult := range artifactResults {
+				val, err := oldStore.Get(ctx, artifactResult.ContentPath)
```
kenxu95, line 72: A "migration" makes it sound like we should be deleting the data from the old store, once we've successfully written to the new store + updated the dag. Is this something we want to do, or are we ok leaving orphaned copies of data in the old store - I'm worried it'll cause confusion somehow.

kenxu95 (commented):
Can you make a task for writing an integration test for this migration? We don't have the ability to do this right now, but we can once our test suite gets more robust at handling third party integrations.

kenxu95:
Another thing that really confused me is that this we aren't migrating metadata right? We're migrating actual computed artifact data. The metadata stays in the Aqueduct-managed metadata store, which is, as of now, always on the server machine right?

src/golang/lib/workflow/artifact/migrate.go
saurav-c: Yes, we'll overwrite it. When a user sets something as the metadata store, they are giving us permission to do so.

src/golang/lib/workflow/artifact/migrate.go
saurav-c: The migration is atomic, so even if the content has been copied over, the DB records will not have been updated, so we simply run the entire migration again.

kenxu95 (approved):

saurav-c:
@kenxu95 You're right. We use storage for a few different reasons:
1. Temporary metadata about operator execution. This is simply for communicating between the Aqueduct server and actual operator execution. We persist this into the database once execution is complete.
2. Persisting artifact result content.
3. Persisting operator code (i.e. function or check code)
4. Vault secrets, such as integration config, but right not the vault the storage aren't completely unified and have their own configs.

src/golang/lib/workflow/artifact/migrate.go
saurav-c: @vsreekanti Are we ok with orphaned data or do we want to clean this up explicitly. I would rather not delete as part of the migration just in case something goes wrong. We wouldn't want to lose customer data.

vsreekanti: I'd prefer not to leave stuff lying around that's then orphaned because we'll lose the context for it, no? Maybe we can do it in two stages: Move everything first with a verification step, then delete?

saurav-c: Added as part of #561

PR #560, opened 2022-10-06
Chunk SQL insert for sqlite3 integration

cw75:
Due to SQLITE's variable number limit, we need to chunk the data insertion operation so that the number of variables is less than 32766. Verified that this fixes the issue on the wine workflow.

### DIFF
--- a/src/python/aqueduct_executor/operators/connectors/data/sqlite.py
+++ b/src/python/aqueduct_executor/operators/connectors/data/sqlite.py
@@ -1,14 +1,23 @@
+import sqlite3
+
 import pandas as pd
 from aqueduct_executor.operators.connectors.data import config, load, relational
 from aqueduct_executor.operators.utils.enums import ArtifactType
+from packaging.version import parse as parse_version
 from sqlalchemy import create_engine, engine
 
 # https://www.sqlite.org/limits.html#max_variable_number
 SQLITE_MAX_VARIABLE_NUMBER = 32766
+SQLITE_MAX_VARIABLE_NUMBER_EARLIER_VERSION = 999
 
 
 class SqliteConnector(relational.RelationalConnector):
     def __init__(self, config: config.SqliteConfig):
+        if parse_version(sqlite3.sqlite_version) >= parse_version("3.32.0"):
+            self.variable_number_limit = SQLITE_MAX_VARIABLE_NUMBER
+        else:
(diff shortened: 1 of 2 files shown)


src/python/aqueduct_executor/operators/connectors/data/sqlite.py
```diff
@@ -1,12 +1,35 @@
-from aqueduct_executor.operators.connectors.data import config, relational
+import pandas as pd
+from aqueduct_executor.operators.connectors.data import config, load, relational
+from aqueduct_executor.operators.utils.enums import ArtifactType
 from sqlalchemy import create_engine, engine
 
+SQLITE_MAX_VARIABLE_NUMBER = 32766
```
kenxu95, line 6: where did this number come from again? Can we add a comment about that here

src/python/aqueduct_executor/operators/connectors/data/sqlite.py
```diff
@@ -1,12 +1,35 @@
...
+        # callable function for `method` that does bulk loading.
+        # See: https://pandas.pydata.org/docs/user_guide/io.html#io-sql-method
+        df.to_sql(
+            params.table,
+            con=self.engine,
+            if_exists=params.update_mode.value,
+            index=False,
+            method="multi",
+            # We need to specify chunksize due to sqlite3's variable number limit.
+            chunksize=int(SQLITE_MAX_VARIABLE_NUMBER / len(df.columns)),
```
kenxu95, line 30: For my understanding (and maybe we could also add a comment about this calculation), it's one variable per column, so this restricts the number of rows such that the total number of variables is under the limit?

kenxu95 (approved):
Make sense, some requests to preserve a little more context in the codebase.

src/python/aqueduct_executor/operators/connectors/data/sqlite.py
cw75: yup, will add a comment.

cw75:
@kenxu95 added comments.

cw75:
@kenxu95 Yeah in theory, but from my observation the `sqlite3` package that comes with Python3 usually have versions after `3.32.0`, so I expect this to be a very rare case (with 999, for a table of 16 columns like this one you can only insert 62 rows at a time :-)). Anyway I can still add a branching logic to account for this case.

cw75:
@kenxu95 updated.

PR #573, opened 2022-10-10
Update k8s integration UI when the server runs in an existing cluster

cw75:
When the Aqueduct server is already running in a k8s cluster, when connecting to a k8s integration we want to give users the option to use the same cluster to launch their operators. We implement this by setting an environment variable in the helm chart template and adding an endpoint to the server that returns the value of this env var to the UI. Based on its value, the UI decides whether to show the check box to run the compute from within the same cluster.

There are a few more fixes on the backend side that I'll patch in a subsequent PR, but this PR completes the UI portion.

## Loom demo (if any)
https://www.loom.com/share/f68f10ae548e44199e1b43975bcb1fe8

### DIFF
--- a/src/golang/cmd/server/handler/get_server_environment.go
+++ b/src/golang/cmd/server/handler/get_server_environment.go
@@ -1,0 +1,49 @@
+package handler
+
+import (
+	"context"
+	"net/http"
+	"os"
+
+	log "github.com/sirupsen/logrus"
+)
+
+const inK8sClusterEnvVarName = "AQUEDUCT_IN_K8S_CLUSTER"
+
+type getServerEnvironmentResponse struct {
+	// Whether the server is running within a k8s cluster.
+	InK8sCluster bool `json:"inK8sCluster"`
+}
+
+// Route: /api/environment
+// Method: GET
+// Request:
+//
+//	Headers:
+//		`api-key`: user's API Key
+//
+// Response: Aqueduct server's environment variables.
+type GetServerEnvironmentHandler struct {
+	GetHandler
+}
+
+func (*GetServerEnvironmentHandler) Name() string {
+	return "GetServerEnvironment"
+}
+
+func (*GetServerEnvironmentHandler) Prepare(r *http.Request) (interface{}, int, error) {
+	return nil, http.StatusOK, nil
+}
+
+func (h *GetServerEnvironmentHandler) Perform(ctx context.Context, interfaceArgs interface{}) (interface{}, int, error) {
+	inCluster := false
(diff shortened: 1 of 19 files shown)


src/ui/common/src/components/integrations/dialogs/kubernetesDialog.tsx
```diff
@@ -1,43 +1,102 @@
...
+        headers: {
+          'api-key': apiKey,
+        },
+      });
+
+      const responseBody = await environmentResponse.json();
+      setInK8sCluster(responseBody['inK8sCluster']);
+    };
+
+    fetchEnvironment().catch(console.error);
```
agiron123, line 47: neat, didn't know you could do this actually. I usually just wrap the async stuff with a try/catch statement and handle errors from inside the catch block.

src/golang/cmd/server/handler/get_server_environment.go
```diff
@@ -0,0 +1,49 @@
...
+	return nil, http.StatusOK, nil
+}
+
+func (h *GetServerEnvironmentHandler) Perform(ctx context.Context, interfaceArgs interface{}) (interface{}, int, error) {
+	inCluster := false
+	if os.Getenv(inK8sClusterEnvVarName) == "1" {
+		inCluster = true
+	}
+
+	log.Info("Incluster is %s", inCluster)
```
agiron123, line 44: was this log for debugging, or do we need it?

agiron123 (approved):
LGTM nice work!

src/golang/cmd/server/handler/get_server_environment.go
cw75: removed

likawind:
As I wasn't closely follow this feature, I was initially a bit confused what it mean by 'using the same k8s cluster that the server is running on' until you explain it. Not sure if user will share this confusion. Should we use 'Run workflows using the same ...' ?

Another thing, maybe it's a separate task - is there something in integrations page showing which engine we are currently running workflows?

cw75:
@likawind Thx for bringing this up! When registering a k8s integration, the goal is for the user to provide credential to a k8s cluster, so I think "Use the same Kubernetes cluster that the server is running on." seems to make sense. I propose we keep it this way and we can always tweak the message if users find it confusing.

As for showing which engine we are currently running workflows, this is a workflow-dependent property: a user may register a k8s integration and a lambda integration, and choose to run workflow A on k8s and workflow B on lambda, so I think this info should probably be shown on the workflow page instead of the integration page. I will create a task about this.

PR #576, opened 2022-10-11
Fix a number of bugs in the helm chart

cw75:
There are two main fixes:
1. Since our executor attempts to create jobs in the `aqueduct` namespace, in the helm chart we need to define role and rolebinding in the `aqueduct` namespace.
2. In our Golang code we have logic to create namespace `aqueduct` if not exist. Since namespace is a cluster resource, we need to use clusterrole and clusterrolebinding for this.

### DIFF
--- a/src/aqueduct-helm/Chart.yaml
+++ b/src/aqueduct-helm/Chart.yaml
@@ -22,4 +22,3 @@ version: 0.0.16
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
 appVersion: "0.0.16"
-namespace: aqueduct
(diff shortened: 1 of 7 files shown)


src/aqueduct-helm/Chart.yaml
```diff
@@ -22,4 +22,3 @@ version: 0.0.16
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
 appVersion: "0.0.16"
-namespace: aqueduct
```
cw75, line 25: This doesn't seem to have any effect, so I'm dropping this. From the documentation it seems that we can't specify namespace in `Chart.yaml`, instead we have to use `--namespace` when calling `helm install`.

hsubbaraj-spiral (approved):

PR #580, opened 2022-10-11
Fix mypy errors that resulted from numpy version > 1.19

kenxu95:
Once this merges, we can once again upgrade our numpy to whatever version and run `run_linters.py` without fear.

Turns out all we had to do to fix this was to make `Union[int, float, np.number]` its own type. Don't ask me why that works...

The other errors were stomped out why asserting that data was of the correct type before returning.

ENG-1754

### DIFF
--- a/sdk/aqueduct/artifacts/bool_artifact.py
+++ b/sdk/aqueduct/artifacts/bool_artifact.py
@@ -74,19 +74,19 @@ class BoolArtifact(BaseArtifact):
                     "Parameterizing historical artifacts is not currently supported."
                 )
 
-        if parameters is None and self._get_content() is not None:
-            return self._get_content()
+        content = self._get_content()
+        if parameters is not None or content is None:
+            previewed_artifact = artifact_utils.preview_artifact(
+                self._dag, self._artifact_id, parameters
+            )
+            content = previewed_artifact._get_content()
+            assert isinstance(content, bool) or isinstance(content, np.bool_)
 
-        previewed_artifact = artifact_utils.preview_artifact(
-            self._dag, self._artifact_id, parameters
-        )
-        content = previewed_artifact._get_content()
+            # If the artifact was previously generated lazily, materialize the contents.
+            if parameters is None and self._get_content() is None:
+                self._set_content(content)
+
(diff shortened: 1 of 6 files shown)


sdk/aqueduct/artifacts/bool_artifact.py
```diff
@@ -74,19 +74,19 @@ def get(self, parameters: Optional[Dict[str, Any]] = None) -> Union[bool, np.boo
...
 
-        if parameters is None and self._get_content() is not None:
-            return self._get_content()
+        content = self._get_content()
+        if parameters is not None or content is None:
+            previewed_artifact = artifact_utils.preview_artifact(
+                self._dag, self._artifact_id, parameters
+            )
+            content = previewed_artifact._get_content()
+            assert isinstance(content, bool) or isinstance(content, np.bool_)
```
cw75, line 83: Do we have to do this just for `self._set_content(content)` to not complain? Because it seems that we are doing the same assertion in line 89 before returning anyway.

likawind, line 83: is this necessary?

cw75 (commented):
@kenxu95 LGTM. Left a question inline. Also, it seems we are slightly changing the structure in bool and numeric artifacts, so should we do the same for other artifact classes?

sdk/aqueduct/artifacts/bool_artifact.py
likawind: should line 89 also cover this?

likawind (approved):

sdk/aqueduct/artifacts/bool_artifact.py
kenxu95: actually yeah, I initially wanted to put this here just so we wouldn't ever set the content of this artifact to a non-bool data type. However, that seems unnecessary now that I think about it - these assertions are really just for type-checking, since it content value should never be a bool.

kenxu95: yeah it isn't (see answer below), removing.

PR #593, opened 2022-10-12
Adds Persistent Volume to Aqueduct in-cluster Deployment

hsubbaraj-spiral:
Adds Persistent Volume to Aqueduct in-cluster Deployment.
https://linear.app/aqueducthq/issue/ENG-1801/add-persistent-volume-to-helm-chart
## Loom demo (if any)
```
(base) ubuntu@ip-172-31-46-200:~$ kubectl get pv -n aqueduct
NAME          CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS   CLAIM                        STORAGECLASS   REASON   AGE
aqueduct-pv   10Gi       RWO            Retain           Bound    aqueduct/aqueduct-pv-claim   manual                  5m50s
```
in description of the pod:
```
Volumes:
  aqueduct-pv-storage:
    Type:       PersistentVolumeClaim (a reference to a PersistentVolumeClaim in the same namespace)
    ClaimName:  aqueduct-pv-claim
    ReadOnly:   false
```


src/aqueduct-helm/templates/persistentvolume.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: aqueduct-pv
+spec:
+  storageClassName: aqueduct
+  capacity:
+    storage: 5Gi
+  accessModes:
+    - ReadWriteOnce
+  hostPath:
+    path: "/root/.aqueduct"
```
cw75, line 12: It seems from here that `hostPath` is only used for dev and testing for single-node cluster? Does this work for multi-node clusters?

cw75 (commented):
@hsubbaraj-spiral Left a couple questions inline. Have we tested that if we kill the pod and restart another one, data in `~/.aqueduct` is persisted?

src/aqueduct-helm/templates/persistentvolume.yaml
hsubbaraj-spiral: Yep I tested this on multi-node clusters as well.

src/aqueduct-helm/templates/persistentvolumeclaim.yaml
```diff
@@ -3,8 +3,9 @@ kind: PersistentVolumeClaim
 metadata:
```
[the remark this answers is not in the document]
hsubbaraj-spiral: The persistent volume claim is sufficed by the persistent volume that we generate. There isn't a direct link, that resource just needs to be available to claim.

cw75 (approved):
LGTM! Can you also add a link to the documentation of how dynamic provisioning works?

Another thing is that I wonder if we should mount just `/root/.aqueduct` or the whole `/root/*`. Because the server pod may have installed some Python dependencies that's not gonna be stored in `/root/.aqueduct`, so if we only mount this dir we will lose these dependency installations.

PR #606, opened 2022-10-14
Fix bug where artifact preview for non tabular types aren't working properly

cw75:
@agiron123 @likawind I implemented the fix by adding the missing serialization field to the Go response, plus bringing some missing pieces from the original implementation back.

## Loom demo (if any)
https://www.loom.com/share/6c7b18e19f404f62a39f610063a4ac02


src/ui/common/src/components/workflows/artifact/content.tsx
```diff
@@ -56,57 +56,83 @@ const ArtifactContent: React.FC<Props> = ({
...
   if (!contentWithLoadingStatus.data) {
-    return (
-      <Typography variant="h5" component="div" marginBottom="8px">
-        No result to show for this artifact.
-      </Typography>
-    );
-  }
-
-  if (
-    artifact.type === ArtifactType.Bytes ||
```
likawind, line 67: I feel this makes the switch statement below slightly hard to read. I would consider doing the following instead:
* put Bytes and Picklable inside the switch below so that this switch cover all types. For these two types we don't have to check if the data field exists
* For those we expected to have data filled (e.g. images, table), if the data doesn't exists, we show an alert since this should be considered as an error case.

likawind (approved):
LGTM, thanks for fixing this! Some suggestions on rendering based on serialization type

src/ui/common/src/components/workflows/artifact/content.tsx
cw75: The checking of whether data field is empty happens before the switch statement, so if I put these two cases inside the switch statement they will never be hit right?

likawind: Yeah I mean this way we will completely remove the empty field check in line 58 and add those for each case in the switch. Since you already build error handling for image and table, it should be fine to just add that to json case

PR #610, opened 2022-10-14
[7/7] Syncs vault with storage at server start

saurav-c:
This PR syncs the vault with storage by moving the vault into the current storage layer. This operation will only happen once at server start time if it has not happened in the past.

This is necessary because currently the vault is found in the local filesystem at `aqueduct_path/vault`, whereas storage can be either in a different directory locally or it could be set to cloud storage, i.e. S3 or GCS.

### DIFF
--- a/src/golang/cmd/server/server/aqueduct_server.go
+++ b/src/golang/cmd/server/server/aqueduct_server.go
@@ -193,8 +193,7 @@ func (s *AqServer) Init() error {
 		readers.IntegrationReader,
 		db,
 	); err != nil {
-		db.Close()
-		log.Fatalf("Unable to sync vault with storage: %v", err)
+		return err
 	}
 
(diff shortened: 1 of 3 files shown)


likawind (commented):
My understanding of the current approach - it seems that we mainly check if we've migrated from the default vault based on the default file path?
Assuming that, I'm curious if we need to handle migrating from non-default vault. For example, if the user want to switch from gcs to s3, or from s3 back to file, etc.

saurav-c:
@likawind This is a special one off case where we are just trying to move the local vault from `.aqueduct/server/vault` to `.aqueduct/server/storage/vault`. Any subsequent migrations are not going to be using the code in this PR, they will be using the implementation found in #549

likawind (approved):
Understood, I think we can ship these PRs

PR #615, opened 2022-10-17
Add tfrs.Model type support

kenxu95:
Adds type support for any non-pickleable classes that inherit from tf.keras.Model. `tf.keras.Model` objects themselves are actually picklable, and so are already supported by our typing system.

In the example notebook torn pulled from https://www.tensorflow.org/recommenders, we can actually get it working with the existing system if `train_model` returns the user and movie models directly, instead of the `MoveLensModel`, which is a `tfrs.Model` (which inherits from `tf.keras.Model`).

If `train_model()` returns a `MovieLensModel`, we get the following error:
```
ValueError: Model <__main__.train_model.<locals>.MovieLensModel object at 0x7f1f20436100> cannot be saved either because the input shape is not available or because the forward pass of the model is not defined.To define a forward pass, please override `Model.call()`. To specify an input shape, either call `build(input_shape)` directly, or call the model on actual data using `Model()`, `Model.fit()`, or `Model.predict()`. If you have a custom training step, please make sure to invoke the forward pass in train step through `Model.__call__`, i.e. `model(inputs)`, as opposed to `model.call()`.
```
I'm not sure I have the context to fix this before asking ya'll. Googling around, I see a lot of people saying you have to call `model.fit()` beforehand, but we're doing exactly that in this notebook...

ENG-1822

### DIFF
--- a/sdk/aqueduct/enums.py
+++ b/sdk/aqueduct/enums.py
@@ -151,6 +151,7 @@ class ArtifactType(str, Enum, metaclass=MetaEnum):
     BYTES = "bytes"
     IMAGE = "image"  # corresponds to PIL.Image.Image type
     PICKLABLE = "picklable"
+    TF_KERAS = "tensorflow-keras-model"
 
 
 class SerializationType(str, Enum, metaclass=MetaEnum):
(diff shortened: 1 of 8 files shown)


kenxu95:
@cw75 from [this](https://stackoverflow.com/questions/48295661/how-to-pickle-keras-model) it sounds like tf keras models are supposed to be picklable "for now", but that they recommend using `model.save()`? I still like our ordering though because of that bug in the description.

vsreekanti (approved):

PR #616, opened 2022-10-17
[6/7] Implements GCS vault

saurav-c:
This PR implements a GCS vault.

ENG 1797

### DIFF
--- a/src/golang/lib/storage/gcs.go
+++ b/src/golang/lib/storage/gcs.go
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"context"
 	"io"
+	"path"
+	"strings"
 
 	"cloud.google.com/go/storage"
 	"github.com/aqueducthq/aqueduct/lib/collections/shared"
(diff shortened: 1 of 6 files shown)


src/golang/lib/storage/gcs.go
```diff
@@ -20,23 +22,44 @@ func newGCSStorage(gcsConfig *shared.GCSConfig) *gcsStorage {
 	}
 }
 
+// parseBucketAndKey uses a bucket in the form of bucket/path and a key and
+// returns the bucket name and the full key.
+func (g *gcsStorage) parseBucketAndKey(key string) (string, string) {
+	parts := strings.Split(g.gcsConfig.Bucket, "/")
```
likawind, line 28: would it be easier to ask user to provide a `bucket` and a `path` separately? Or this is the canonical way to specify a gcs location?

likawind (approved):
Looks great besides a minor question!

src/golang/lib/storage/gcs.go
saurav-c: As of right now, we just allow specifying a bucket. The reason we need this method is for internal purposes because we want to put the vault in a special path under `/vault`. Any user facing changes would be out of scope for this PR.

PR #618, opened 2022-10-18
ENG-1852 Adds back missing operator content download button

likawind:
This PR adds back missing operator content download button in both operator sidesheets and operator details page. This results in two buttons of the same function in sidesheet mode, but I think it should be fine?

We also fixed some minor alignment issues and other bugs when working on this feature:
* the details page overflows under sidesheet mode due to the padding of the sidesheet container. We reduced the width under sidesheet mode to be `sidesheetWidth - 2 * padding`. Previously it was just `sidesheetWidth`.
* The file content wasn't showing up in file viewer.
* the sidesheet scroll bar always appears and makes right / bottom of sidesheet looks a bit wierd.

ENG-1852

## Loom demo (if any)
Both sidesheet buttons and detail page buttons works. We also verified download files are valid. Verified things working both during clicking around and after refresh.
https://www.loom.com/share/4187dab8ccd347ff93e4d53ed6791502


src/ui/common/src/components/MultiFileViewer/index.tsx
```diff
@@ -44,17 +44,14 @@ const MultiFileViewer: React.FC<Props> = ({
...
+  const hasFiles = files && Object.keys(files).length > 0;
   let selected = files;
   if (hasFiles) {
     const pathList = selectedFile.split('/').splice(1);
 
     pathList.forEach((section) => {
       if (Object.keys(selected).includes(section)) {
         selected = selected[section];
-      } else {
-        hasFiles = false;
```
likawind, line 57: @agiron123 I'm not sure why you added this originally but this has caused the viewer not showing any content when there's no file selected

src/ui/common/src/components/pages/operator/id/index.tsx
```diff
@@ -176,18 +184,8 @@ const OperatorDetailsPage: React.FC<OperatorDetailsPageProps> = ({
       }
     }
 
-    // TODO: only call this when we are inside a function operator. not an extract or load operator.
```
likawind, line 179: slightly clean up this code by directly render the spec using code viewer. Thus, there's no need to set file for extract here

src/ui/common/src/components/pages/operator/id/index.tsx
```diff
@@ -225,30 +223,52 @@ const OperatorDetailsPage: React.FC<OperatorDetailsPageProps> = ({
...
+    if (operator?.spec?.type === OperatorType.Extract) {
       return (
         <Box>
           <Typography variant="h6" fontWeight="normal" mb={1}>
-            Code Preview
+            Query Preview
           </Typography>
-          <MultiFileViewer files={files} defaultFile={operator.name || ''} />
+          <CodeBlock language="sql">
+            {operator.spec.extract.parameters.query}
```
likawind, line 260: render query directly

src/ui/common/src/components/pages/operator/id/index.tsx
```diff
@@ -225,30 +223,52 @@ const OperatorDetailsPage: React.FC<OperatorDetailsPageProps> = ({
...
-        <CodeBlock language="sql">{fileContent}</CodeBlock>
-      </Box>
-    );
+    return null;
   };
 
   return (
     <Layout breadcrumbs={breadcrumbs} user={user}>
-      <Box width={'800px'}>
+      <Box width={sideSheetMode ? SidesheetContentWidth : '100%'}>
```
likawind, line 271: makes the width more adaptive when not under sidesheet mode.

src/ui/common/src/components/MultiFileViewer/index.tsx
agiron123: I was trying to get this to work with showing the SQL query that was used in an operator, but that ended up taking a bit too long, so i chose another approach. 

I think I just forgot to take this back out.

src/ui/common/src/utils/sidesheets.tsx
```diff
@@ -35,7 +35,7 @@ export function getDataSideSheetContent(
         px={'16px'}
         maxWidth="800px"
         height="100vh"
-        sx={{ overflowY: 'scroll' }}
+        sx={{ overflowY: 'hidden' }}
```
agiron123, line 38: Why is this changed? Are users still able to scroll vertically on the sidesheet if there is content to scroll to?

My hopes of doing things like this was to have the root layout handle showing scroll bars.

agiron123 (approved):
Approved with comments.

src/ui/common/src/utils/sidesheets.tsx
agiron123: @eunice-chan this may be related to the work that you are doing here: 
https://linear.app/aqueducthq/issue/ENG-1873/enable-scrolling-in-side-panel

likawind: I misunderstood what 'hidden' means, updated to use 'auto'. I also added some minor adjustments so that scroll bar position and margin looks better:

Screen Shot 2022-10-19 at 5 24 17 PM

PR #619, opened 2022-10-18
ENG-1657: Clean up Linter Warnings

agiron123:
This PR cleans up frontend linter warnings.

ENG-1657

### DIFF
--- a/src/ui/common/src/components/Search/index.tsx
+++ b/src/ui/common/src/components/Search/index.tsx
@@ -82,7 +82,7 @@ export const filteredList = (
   matchOn: (item: any) => string,
   listItems: (item: any, idx: number) => JSX.Element,
   noItemsMessage: JSX.Element
-) => {
+): JSX.Element => {
   if (allItems.length === 0) {
     return noItemsMessage;
   }
--- a/src/ui/common/src/components/Search/index.tsx
+++ b/src/ui/common/src/components/Search/index.tsx
@@ -18,11 +18,7 @@ type Props = {
   setSearchTerm: (v: string) => void;
 };
 
-export const SearchBar: React.FC<Props> = ({
-  options,
-  getOptionLabel,
-  setSearchTerm,
-}) => {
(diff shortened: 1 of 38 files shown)


src/ui/common/src/components/Search/index.tsx
```diff
@@ -86,7 +82,7 @@ export const filteredList = (
   matchOn: (item: any) => string,
   listItems: (item: any, idx: number) => JSX.Element,
   noItemsMessage: JSX.Element
-) => {
+): JSX.Element => {
```
vsreekanti, line 85: Out of curiosity what is `JSX.Element` vs `React.Element`?

src/ui/common/src/components/notifications/NotificationListItem.tsx
```diff
@@ -39,39 +34,40 @@ export const NotificationListItem: React.FC<Props> = ({
...
-              {notification.workflowMetadata.name}
-            </Typography>
-          </Typography>
-        </Box>
-      ) : (
-        <Box />
-      );
-      break;
-    }
+  // NOTE: Looks like title (and all the code in this switch statement) was unused and was giving us a linter warning :)
```
vsreekanti, line 37: If it's unused, can we just delete?

vsreekanti (approved):
👍 One question, one small thing inline.

src/ui/common/src/components/Search/index.tsx
agiron123: It's a more generic version of React.Element.

More info, and probably more than you'd like to know about these two here:
https://stackoverflow.com/questions/58123398/when-to-use-jsx-element-vs-reactnode-vs-reactelement

As for which one to use? I'm not 100% sure, they both seem to do the job and I don't really have strong opinions here.

src/ui/common/src/components/notifications/NotificationListItem.tsx
agiron123: Sure can. Good catch.

PR #624, opened 2022-10-19
[1/7] Allows for blocking requests during system maintenance

saurav-c:
This PR adds a couple of different things to the server struct:
1. An `atomic` bool is added as a flag to indicate when the server should block all new requests. An example of this use case is when performing storage migration.
2. It is not sufficient to just block new requests, sometimes the system maintenance job can only be performed when we are certain that there are no more active requests. One approach is shown in this PR by having a mutex that should be acquired and released by every single request. Each request will acquire a read lock, so requests will not block other requests, but when a system maintenance job needs to performed it can simply acquire an exclusive lock and be confident that there are no more requests active in the server, as otherwise the lock acquisition would still be blocked.

ENG 1871

### DIFF
--- a/src/golang/cmd/server/server/execute_handler.go
+++ b/src/golang/cmd/server/server/execute_handler.go
@@ -48,8 +48,11 @@ func HandleSuccess(
 	handlerObj.SendResponse(w, resp)
 }
 
-func ExecuteHandler(server Server, handlerObj handler.Handler) func(w http.ResponseWriter, r *http.Request) {
+func ExecuteHandler(server *AqServer, handlerObj handler.Handler) func(w http.ResponseWriter, r *http.Request) {
 	return func(w http.ResponseWriter, r *http.Request) {
+		server.RequestMutex.RLock()
+		defer server.RequestMutex.RUnlock()
+
 		args, statusCode, err := handlerObj.Prepare(r)
 		ctx := r.Context()
 		if err != nil {
--- a/src/golang/cmd/server/middleware/maintenance/maintenance.go
+++ b/src/golang/cmd/server/middleware/maintenance/maintenance.go
@@ -1,0 +1,23 @@
+package maintenance
+
+import (
+	"net/http"
+	"sync/atomic"
+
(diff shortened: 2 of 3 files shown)


src/golang/cmd/server/server/execute_handler.go
```diff
@@ -48,8 +48,11 @@ func HandleSuccess(
 	handlerObj.SendResponse(w, resp)
 }
 
-func ExecuteHandler(server Server, handlerObj handler.Handler) func(w http.ResponseWriter, r *http.Request) {
+func ExecuteHandler(server *AqServer, handlerObj handler.Handler) func(w http.ResponseWriter, r *http.Request) {
```
cw75, line 51: Curious what's the implication of changing this from an interface to a struct. I remember the initial goal for passing in an interface was so that the enterprise edition can reuse this function? If so, maybe we should implement interface APIs in `Server` for updating the atomic bool and mutex and keep the type to be `Server`.

cw75 (commented):
Looks good! Left a comment inline.

src/golang/cmd/server/server/execute_handler.go
saurav-c: We need access to the AqServer's specific fields for locking and releasing the mutex. It feels odd to be exposing a low-level operation such as updating an atomic bool in the `Server` interface. Since we aren't immediately supporting the enterprise edition, can we create a task for now and come back to this? I think we also need to rethink our `Server` interface in general as part of the system internal refactoring push in the next few weeks.

cw75: I wonder if the enterprise version will have the same fields, so they are not gonna be exclusive to the OSS version, and in that case an interface method seems to make sense. But if we are going to refactor the server interface anyway I don't think this is a blocker for now. Just stamped.

cw75 (approved):

src/golang/cmd/server/server/execute_handler.go
saurav-c: True

likawind (approved):

PR #625, opened 2022-10-19
[2/7] Implements server pause and restart

saurav-c:
This Pr allows for pausing and restarting the server. This is done by separating out the server field initialization from the server constructor. This allows us to ensure that the necessary fields are reinitialized after restart, such as the server vault. The important thing is that the `chi.Router` is not reinitialized as this needs to be the same in order for the HTTP server to continue running.

ENG 1882

### DIFF
--- a/src/golang/cmd/server/server/aqueduct_server.go
+++ b/src/golang/cmd/server/server/aqueduct_server.go
@@ -298,3 +298,19 @@ func IndexHandler() func(w http.ResponseWriter, r *http.Request) {
 
 	return http.HandlerFunc(fn)
 }
+
+// Pause puts the server in system maintenance mode by blocking all new requests
+// and waits for all active requests to finish.
+// It is the responsibility of the caller to call s.Restart() to allow requests
+// to be processed again once the system maintenance is complete.
+func (s *AqServer) Pause() {
+	s.UnderMaintenance.Store(true)
+	s.RequestMutex.Lock()
+}
+
+// Restart restarts a server that was previously stopped via s.Pause().
+func (s *AqServer) Restart() {
+	// init again
+	s.RequestMutex.Unlock()
+	s.UnderMaintenance.Store(false)
+}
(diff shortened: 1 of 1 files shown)


src/golang/cmd/server/server/aqueduct_server.go
```diff
@@ -298,3 +310,21 @@ func IndexHandler() func(w http.ResponseWriter, r *http.Request) {
 
 	return http.HandlerFunc(fn)
 }
+
+// Pause puts the server in system maintenance mode by blocking all new requests
+// and waits for all active requests to finish.
+// It is the responsibility of the caller to call s.Restart() to allow requests
+// to be processed again once the system maintenance is complete.
+func (s *AqServer) Pause() {
```
likawind, line 318: how do we know that all active requests are finished?

src/golang/cmd/server/server/aqueduct_server.go
```diff
@@ -298,3 +310,21 @@ func IndexHandler() func(w http.ResponseWriter, r *http.Request) {
...
+	s.UnderMaintenance.Store(true)
+	s.RequestMutex.Lock()
+}
+
+// Restart restarts a server that was previously stopped via s.Pause().
+func (s *AqServer) Restart() {
+	if err := s.Init(); err != nil {
+		log.Fatalf("Unable to restart server: %v", err)
+	}
+	s.RequestMutex.Unlock()
```
likawind, line 328: why do we unlock before setting maintenance mode to false?

likawind (approved):
Is the main change in this file the implementations of `Pause()` and `Restart()` ? It seems that other changes are mostly moving some codes around

src/golang/cmd/server/server/aqueduct_server.go
saurav-c: Since each request holds a read lock for the duration of the request (https://github.com/aqueducthq/aqueduct/pull/624/files#diff-d54562197a8af971fc47b82fe5c116481a4ac0b4de1f360070a943d1b06f19aaR53), we know for a fact that all requests are finished if this method can acquire an exclusive lock.

src/golang/cmd/server/server/aqueduct_server.go
saurav-c: We're just doing it in the opposite order that they were acquired. As soon as we set maintenance mode to false, new requests will start coming in, but until we unlock the `RequestMutex` they won't be able to proceed.

saurav-c:
@likawind Yes, those are the main changes in this PR.

PR #632, opened 2022-10-20
Move requirements integration tests into their own test suite

kenxu95:
Create a new integration test suite at `integration_tests/no_concurrency/` that runs all integration tests that alter global state (eg. pip install/uninstall, or even migration of the storage engine).

This moves us away from the world where we have to put these @pytest.mark.<flag> decorators on specific test cases.

ENG-1840


cw75 (approved):
If I understand correctly, we don't want concurrency in these tests because if we do, there may be one test installing a package that's later removed by another test, causing the first test to throw an error? Do we know if concurrent pytests share the same Python env, or do they create a separate venv under the hood?

My guess is that they don't have any isolation at the python env or file system level, so if I run a pytest that installs a new package or creates a new file without GCing, after the test these side effects will persist. Is this correct?

kenxu95:
@cw75 yeah I think that's right - I don't think there is isolation at the python env level, but I don't remember if I checked that or not, I'll do that now. Either way, having a separate suite for this could be useful for testing things like migrations (migrate metadata to S3 storage, then back, etc).

kenxu95:
yep just checked, it doesn't play well with concurrency.

PR #635, opened 2022-10-21
Invokes storage migration once storage integration is connected

saurav-c:
This PR puts together the final backend pieces of performing storage and vault migration.
The migration is invoked as an asynchronous goroutine once all other steps for connecting an integration have succeeded. These steps include validating the credentials of the integration.

The goroutine will do the following:
1. Stop all new requests to the server and respond with a server unavailable error code.
2. Wait for all active requests to complete.
3. Wait for all active workflow runs to complete.
4. Perform storage migration.
5. Perform vault migration.
6. Delete all storage and vault content from the old store and vault.
7. Update the storage config to the new config.
8. Unblock workflow runs.
9. Unblock server requests.

There are also a handful of various bug fixes. More details inline or in comments.

## Loom demo (if any)
TEST: Tested a full storage migration from local to S3. Demo coming soon...

### DIFF
--- a/src/golang/cmd/server/handler/connect_integration.go
+++ b/src/golang/cmd/server/handler/connect_integration.go
@@ -15,7 +15,6 @@ import (
 	"github.com/aqueducthq/aqueduct/config"
 	"github.com/aqueducthq/aqueduct/lib/airflow"
 	"github.com/aqueducthq/aqueduct/lib/collections/artifact"
-	db_artifact "github.com/aqueducthq/aqueduct/lib/collections/artifact"
 	"github.com/aqueducthq/aqueduct/lib/collections/artifact_result"
 	"github.com/aqueducthq/aqueduct/lib/collections/integration"
 	"github.com/aqueducthq/aqueduct/lib/collections/operator"
@@ -60,7 +59,7 @@ type ConnectIntegrationHandler struct {
 	JobManager job.JobManager
 
 	WorkflowDagReader    workflow_dag.Reader
-	ArtifactReader       db_artifact.Reader
+	ArtifactReader       artifact.Reader
 	ArtifactResultReader artifact_result.Reader
 	OperatorReader       operator.Reader
 	IntegrationReader    integration.Reader
--- a/src/golang/cmd/server/handler/connect_integration.go
+++ b/src/golang/cmd/server/handler/connect_integration.go
@@ -31,6 +31,7 @@ import (
 	"github.com/aqueducthq/aqueduct/lib/workflow/utils"
 	"github.com/dropbox/godropbox/errors"
 	"github.com/google/uuid"
+	log "github.com/sirupsen/logrus"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


src/golang/cmd/server/handler/connect_integration.go
```diff
@@ -159,27 +163,53 @@ func (h *ConnectIntegrationHandler) Perform(ctx context.Context, interfaceArgs i
...
+		// the connect integration request has succeeded and that the migration is now
+		// under way.
+		go func() {
+			// Wait until the server is paused
+			h.PauseServer()
+			// Makes sure that the server is restarted
+			defer h.RestartServer()
+
+			// Wait until there are no more workflow runs in progress
+			lock := utils.NewExecutionLock()
```
saurav-c, line 184: We need to move this lock acquisition to this higher level function, because workflow runs should continue to remain blocked until the actual storage config has been modified. This happens at the end of the `setIntegrationAsStorage` function.

src/golang/lib/vault/s3.go
```diff
@@ -13,7 +11,15 @@ const (
 
 func newS3Vault(s3StoreConf shared.S3Config, key string) Vault {
 	// The S3 vault stores secrets under the ../vault path
-	s3StoreConf.Bucket = path.Join(s3StoreConf.Bucket, s3VaultDir)
+	// The S3 bucket is in the form of s3:// so we can't use path.Join, because
```
saurav-c, line 14: This fixes a bug with how we were appending `vault` to the s3 bucket name.

src/golang/cmd/server/server/vault.go
```diff
@@ -44,7 +44,7 @@ func syncVaultWithStorage(
 		return err
 	}
 
-	if err := utils.MigrateVault(
+	if _, err := utils.MigrateVault(
```
saurav-c, line 47: This call can ignore the list of keys returned, because a few lines down we delete the entire outdated vault directory anyways.

src/golang/cmd/server/handler/connect_integration.go
```diff
@@ -159,27 +163,53 @@ func (h *ConnectIntegrationHandler) Perform(ctx context.Context, interfaceArgs i
...
+	if args.SetAsStorage {
+		// This integration should be used as the new storage layer.
+		// In order to do so, we need to migrate all content from the old store
+		// to the new store. This requires pausing the server and then restarting it.
+		// All of this logic is performed asynchronously so that the user knows that
+		// the connect integration request has succeeded and that the migration is now
+		// under way.
+		go func() {
+			// Wait until the server is paused
+			h.PauseServer()
```
likawind, line 179: Are we pausing all scheduled workflows in `PauseServer()`?

likawind (approved):
I think the changes generally makes sense. I'm curious if we plan to inform user in UI by:
* notify the user that the server WILL be under maintenance mode once the connect the integration
* even more, we can try to redirect user to some maintenance screen under maintenance mode.

src/golang/cmd/server/handler/connect_integration.go
saurav-c: `PauseServer` can only pause HTTP requests. For the workflows, since that is a cron job we are running, we have to use a different mechanism. This is why we are acquiring the `ExecutionLock` in this function a few lines before.

At the start of each workflow run the executor now has to acquire a read lock for this mutex: see https://github.com/aqueducthq/aqueduct/blob/main/src/golang/cmd/executor/executor/workflow.go#L68. This allows us to block workflow runs.

likawind: I see, so once this lock is acquired by the server, it will prevent any new workflow to be scheduled? In this case, what will happen to the workflow? I assume the migration time is short and newly scheduled workflow will eventually acquire the lock before the acquisition time-out?

PR #636, opened 2022-10-21
Eng 1740 allow users to chain together sql queries

likawind:
This PR implements e2e experience for chaining sql queries based on design discussed in https://linear.app/aqueducthq/issue/ENG-1740/allow-users-to-chain-together-sql-queries-by-creating-temp-tables . We added a `queries` field in relational DB params. When multiple query exists, we compiles it using `WITH` clause and the following steps simply reuse what we have for single query. This field is also rendered in UI using multiple codeblocks.

ENG-1740

## Loom demo (if any)
We run a workflow with the following queries in SDK (same as added integration test):
```
client.create_param("nationality", default=" United Kingdom ")
warehouse = client.integration(name="aqueduct_demo")
reviews = warehouse.sql(["""
    SELECT * FROM hotel_reviews
""",
" SELECT review, review_date from $ where reviewer_nationality ='{{nationality}}'",
" SELECT review from $"])
```

Verified everything works including UI:
Screen Shot 2022-10-21 at 1 37 00 PM

One minor design question is if we should provide more information when showing such operator in Integration Detail Page, when we show workflows and operators associated with a given integration:
Screen Shot 2022-10-21 at 1 35 52 PM


src/python/aqueduct_executor/operators/connectors/data/extract.py
```diff
@@ -16,15 +18,48 @@ class RelationalParams(models.BaseParams):
...
+    queries: Optional[List[str]] = None
 
     # TODO: Consider not including github as part of relational params when it is JSON marshalled
     github_metadata: Optional[Any]
 
-    def expand_placeholders(
-        self,
-        parameters: Dict[str, str],
-    ) -> None:
+    def _compile_chain(self, queries: List[str]) -> str:
```
cw75, line 27: Let's add a sentence or two that describe at a high level what this function accomplishes.

src/ui/common/src/components/operators/operatorParametersOverview.tsx
```diff
@@ -28,11 +28,20 @@ const OperatorParametersOverview: React.FC<Props> = ({
 
     // These checks tries to distinguish googlesheet vs relational
     // extracts based on the fields of type union exParams.
-    if ('query' in exParams) {
+    if ('query' in exParams || 'queries' in exParams) {
+      const relationalParams = exParams as RelationalDBExtractParams;
+      if (!!relationalParams.queries) {
+        return (
+          <Typography variant="body2" color={textColor}>
+            {`${relationalParams.queries.length} chained queries.`}
```
cw75, line 36: It'll be more informative if we can find a way to nicely show the content on all chained queries.

cw75 (approved):
Looks good! Left some nits inline, mainly about commenting.

kenxu95 (commented):
I'll look at the implementation next, but some comments about the surfaces based on the description and images:
- I noticed that the first query is indented on the UI sidesheet. Can we make it consistent?
- Can we add a tooltip to the `Extract Parameters` header that notes '$' means?
- Is "Extract Parameters" the right term? I personally find this confusing since the concept of a parameter in our system is different. Could we say "Extract Statements"? (I assume its not limited to SQL?) 
- The above also applies to the workflow Integration Detail Page. The "Parameters" column confuses me - it seems like what we're actually displaying is something like "Additional Info".

I think it would be good to have the name of the integration on extract operators on the Integration Detail page as a separate column.

sdk/aqueduct/integrations/sql_integration.py
```diff
@@ -150,32 +152,59 @@ def sql(
...
+                extract_params = RelationalDBExtractParams(
+                    queries=extract_params,
+                )
+        elif isinstance(
+            extract_params, RelationalDBExtractParams
+        ):  # query is a RelationalDBExtractParams object
+            if int(bool(extract_params.query)) + int(bool(extract_params.queries)) != 1:
+                raise Exception(
+                    "For a RelationalDBExtractParams object, exactly one of .query or .queries fields should be set."
+                )
```
kenxu95, line 175: I have a slight preference for having a single field on RelationalDBExtractParams for queries with `Union[str, List[str]]` type, since that's what we've been doing throughout the SDK - see `artifacts` parameter in `publish_flow()`.

sdk/aqueduct/integrations/sql_integration.py
```diff
@@ -150,32 +152,59 @@ def sql(
...
-                    raise InvalidUserArgumentException(
-                        "The parameter `%s` must be defined as a string. Instead, got type %s"
-                        % (param_name, param_artifact.type)
-                    )
-
-                sql_input_artifact_ids.append(param_artifact.id)
+
+        if extract_params.queries is not None:
+            for q in extract_params.queries:
+                matches.extend(re.findall(TAG_PATTERN, q))
```
kenxu95, line 186: To simplify this, can we make a separate list variable containing all the queries, and just have one loop over that? In the single element case, this would just be a list of size 1.

src/python/aqueduct_executor/operators/connectors/data/extract.py
```diff
@@ -16,15 +18,66 @@ class RelationalParams(models.BaseParams):
...
+
+                    # there are more subqueries to append in this 'WITH' clause
+                    if idx < len(queries) - 2:
+                        with_clause += ",\n"
+                    prev_table_name = cur_table_name
+                # otherwise, append the last query to 'WITH' clause
+                else:
+                    return f"{with_clause}\n{normalized_query}"
+
+        return queries[0]
```
kenxu95, line 78: Can we put this in an else statement? This prevents the if statement above from forgetting to return something.

src/python/aqueduct_executor/operators/connectors/data/extract.py
```diff
@@ -34,23 +87,38 @@ def expand_placeholders(
...
+
+    def compile(self, parameters: Dict[str, str]) -> None:
+        assert (
+            int(bool(self.query)) + int(bool(self.queries)) == 1
+        ), "Exactly one of .query and .queries fields should be set."
+        query = ""
+        if self.query:
+            query = self.query
+            print(f"Compiling query {query} .")
+        # this check mainly bypasses linter. The `assert` block above
```
kenxu95, line 112: nit: Capitalize first word?

src/golang/lib/collections/operator/connector/extract_params.go
```diff
@@ -11,6 +11,7 @@ type ExtractParams interface {
 type RelationalDBExtractParams struct {
 	GithubMetadata *gh_types.GithubMetadata `json:"github_metadata"`
 	Query          string                   `json:"query"`
+	Queries        []string                 `json:"queries"`
```
kenxu95, line 14: I assume we made them two fields instead of one list field so that we wouldn't have to do a migration. How hard would a migration be? This seems like unnecessary tech debt to me, I'd much rather reason about a single list.

src/ui/common/src/utils/operators.ts
```diff
@@ -74,7 +74,8 @@ export type ExtractParameters =
   | GoogleSheetsExtractParams;
 
 export type RelationalDBExtractParams = {
-  query: string;
+  query?: string;
+  queries?: string[];
```
kenxu95, line 78: My argument for making this a single list is that this stuff propagates to every one of our surfaces, and would rather pay the upfront migration cost now.

kenxu95 (changes requested):

kenxu95:
- Ok, I think adding a link in the tooltip should be good enough. I still think we should supply a tooltip though, since there could be users that just want to browse around the UI? I don't want to assume too much about what users know. In either case, I think this could be a good test case to put in https://www.notion.so/aqueducthq/UI-Testing-2f2c644e03394cf994fd199ba83cb7e1 (verify that the tooltip link to the SQL documentation is not broken).

- Yeah, I think your suggestions about 'parameter' term are steps in the right direction. As long as we get rid of the term, and display useful integration-specific information I'm happy.

likawind:
@kenxu95 and myself discussed offline on migration:
Given the solution is not finalized (https://linear.app/aqueducthq/issue/ENG-1892/allow-user-to-view-intermediate-tables-when-a-sql-is-chained) and we don't have much up-coming changes on query processing, we decided to go with current way to save some migration time. We will add some comments to clarify the use of two fields (`query` and `queries`)

src/python/aqueduct_executor/operators/connectors/data/extract.py
likawind: I decided to swap so that we have `if len(queries)==1: return queries[0]`

vsreekanti:
What do you mean by the "current way", @likawind? I agree that the word parameters is confusing and that the use of `$` is a little opaque.

cw75:
If I understand correctly "current way" refers to keeping the `query` and `queries` fields separate to avoid the need for a migration. @likawind could you confirm?

likawind:
@cw75 yup

likawind:
@vsreekanti @kenxu95 I updated tooltips and section names based on our discussion:
Screen Shot 2022-10-24 at 9 17 49 PM
Screen Shot 2022-10-24 at 9 10 17 PM

@vsreekanti we need a way to specify the placeholder for 'intermediate tables'. I don't see how we could do much better than using `$`, welcome to suggestions.

vsreekanti:
This is good, @likawind!

I think you can cut the explanation of the parameters. The explanation of the `$` syntax makes sense to me. 

Let's make sure to make the styling of the icon button consistent with the rest of our UI.

kenxu95:
@likawind should the tooltip message be ".. refers to the output of the previous query"? (singular)

sdk/aqueduct/integrations/sql_integration.py
```diff
@@ -122,7 +124,8 @@ def sql(
 
         Args:
             query:
-                The query to run.
+                The query to run. When a list is provided, we run the list
+                in a chain and returns the result of the final query.
```
kenxu95, line 128: grammar nit: "return the result"

src/python/aqueduct_executor/operators/connectors/data/extract.py
```diff
@@ -16,15 +18,71 @@ class RelationalParams(models.BaseParams):
     # any user-defined tags like `{{today}}`.
     query_is_usable: Optional[bool] = False
 
-    query: str
+    # exactly one of 'query' and 'queries' will be set.
+    # `query` represents a single query.
+    query: Optional[str] = None
+    # `queries` represents a chain of queries. We must first run _compile_chain
+    # to compile it to a single query before further processing.
+    queries: Optional[List[str]] = None
```
kenxu95, line 26: nit: capitalize the first word in each comment

src/python/aqueduct_executor/operators/connectors/data/extract.py
```diff
@@ -16,15 +18,71 @@ class RelationalParams(models.BaseParams):
...
+                cur_table_name = f"aqueduct_{uuid.uuid4().hex}"
+                with_clause += f"{cur_table_name} AS (\n{normalized_query}\n)"
+
+                # there are more subqueries to append in this 'WITH' clause
+                if idx < len(queries) - 2:
+                    with_clause += ",\n"
+                prev_table_name = cur_table_name
+
+        # returns `WITH` clause with the normalized final query.
+        return f"{with_clause}\n{normalized_query}"
```
kenxu95, line 83: Same capitalization comment for this section.

kenxu95 (approved):
A couple cosmetic comments, but otherwise fine with me. 

I added a tooltip test case to https://www.notion.so/aqueducthq/UI-Testing-2f2c644e03394cf994fd199ba83cb7e1, but feel free to add anymore than you think would be useful - maybe we should check the "details" column when there are chained queries?

src/python/aqueduct_executor/operators/connectors/data/extract.py
kenxu95: I guess I meant for the "Exactly" comment. But yeah you're right, I think the capitalization should be consistent when we're using the backticks syntax.

PR #639, opened 2022-10-24
Add positional, variable-length support for all functions

kenxu95:
While looking at the task for adding **kwargs support, I noticed that we didn't consider all the possibilities for *args support.

This PR allows us to support user functions that are defined with variable-length parameters:
```
def foo(*args):
    ...
```
This involves disallowing implicit parameters in such cases. For now, it's easier to make this an unsupported case, since we don't really know what name to assign such parameters. If they want to use parameters for such functions, they should create them explicitly.

Here is the error message:

Screen Shot 2022-10-24 at 1 49 39 PM

NOTE: supporting **kwargs is actually a bit more involved than originally intended, as we have to do some stuff on both the sdk and backend side.

### DIFF
--- a/sdk/aqueduct/decorator.py
+++ b/sdk/aqueduct/decorator.py
@@ -234,7 +234,7 @@ def _convert_input_arguments_to_parameters(
         if not isinstance(artifact, BaseArtifact):
             if implicit_params_disallowed:
                 raise InvalidUserArgumentException(
-                    """Input %d to function is not an artifact. Creating an Aqueduct parameter implicitly for a """
+                    """Input at index %d to function is not an artifact. Creating an Aqueduct parameter implicitly for a """
                     """function that takes in variable-length parameters (eg. *args or **kwargs) is currently unsupported."""
                     % idx
                 )
--- a/sdk/aqueduct/decorator.py
+++ b/sdk/aqueduct/decorator.py
@@ -221,6 +221,8 @@ def _convert_input_arguments_to_parameters(
     """
     # KEYWORD_ONLY parameters are allowed, since they are guaranteed to have a name.
     # Note that we only accept them after "*" arguments, since we error out on VAR_POSITIONAL (eg. *args).
+    # For example, `foo(*, positional_arg)` is allowed, but `foo(*args, positional_arg)` is not.
+    # See https://peps.python.org/pep-0362/#parameter-object for a description of each parameter kind.
     disallowed_kinds = [inspect.Parameter.VAR_POSITIONAL, inspect.Parameter.VAR_KEYWORD]
     implicit_params_disallowed = any(
         param.kind in disallowed_kinds for param in func_params.values()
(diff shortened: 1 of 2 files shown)


vsreekanti:
The error message/restriction makes sense to me!

sdk/aqueduct/decorator.py
```diff
@@ -208,21 +208,42 @@ def _type_check_decorated_function_arguments(
...
     """
     Converts non-artifact inputs to parameters.
+
+    Errors if the implicitly-created parameter collides with an existing one. Also errors if the function has a
+    variable-length parameter, since we don't know what name to attribute to those.
+
+    `func_params` maps from parameter name to an `inspect.Parameter` object containing additional information.
     """
+    # KEYWORD_ONLY parameters are allowed, since they are guaranteed to have a name.
+    # Note that we only accept them after "*" arguments, since we error out on VAR_POSITIONAL (eg. *args).
```
cw75, line 223: Not sure I understand what this sentence means. Are you saying if we have `foo(*args, param)`, then we support implicitly creating parameter for `param`? Then it seems to contradict with line 225, since we have `*`, `implicit_params_disallowed` will be true, which prevents the creation of implicit parameters.

sdk/aqueduct/decorator.py
```diff
@@ -208,21 +208,42 @@ def _type_check_decorated_function_arguments(
...
+    disallowed_kinds = [inspect.Parameter.VAR_POSITIONAL, inspect.Parameter.VAR_KEYWORD]
```
cw75, line 224: I'm also a bit confused by what each of these option means. Maybe at least giving a link to the documentation that lists all options would be helpful.

cw75 (commented):
Looks good overall. Left some questions inline.

sdk/aqueduct/decorator.py
kenxu95: ah, so `*args` is different from `*` (I know, right). If you have `foo(*, param)`, then `param` will be of type `KEYWORD_ONLY`, which is ok! `implicit_params_disallowed` will not be true in such a case, and we can create the param implicitly. The `*` wildcard is just there to force any arguments behind it to be keyword arguments (https://peps.python.org/pep-3102/).


I can add https://peps.python.org/pep-0362/#parameter-object to the comment

cw75: Since we have this line:
```
  implicit_params_disallowed = any(
      param.kind in disallowed_kinds for param in func_params.values()
  )
```
So for `foo(*, param)`, won't `*` make `implicit_params_disallowed` to be true due to the use of `any`?

cw75: I guess I'm not super familiar with what `*` means in `foo(*, param)`. If I call this function with (x, y), where x is a non-artifact and y is an artifact, we will attempt to create an implicit parameter for x, and since x doesn't have a name (because the first argument to `foo` is `*`), things will fail? Maybe we should grab a time to chat online to clarify this. I'm also having some trouble understanding the diff between `foo(*, param)` and `foo(*args, param)`. :)

PR #653, opened 2022-10-31
Add ability for SDK tests to work against various data integrations

kenxu95:
We can start to take small but important steps in laying the groundwork for running the SDK test suite against various data integrations. This refactors all the tests to take in the `data_integration` fixture. Currently, this only works with `aqueduct_demo`, but is easily extendible to something like `S3`.

It's important to start getting this framework down so we can use the same pattern with different execution engines for the various projects that are coming up.

The assumption is currently that the server is already set up with an integration. In the future, we can look at specifying the data integrations to test against in the command line. And eventually to automatically connecting integrations from a yaml file.

ENG-1836

### DIFF
--- a/integration_tests/sdk/conftest.py
+++ b/integration_tests/sdk/conftest.py
@@ -31,7 +31,7 @@ def fetch_flags(pytestconfig):
 
 
 def all_data_integration_names() -> List[str]:
-    """We currently only support the demo database, but this can eventually be configured through the command line."""
+    """We currently only support the demo database, but this can eventually be configured arbitrarily."""
     return ["aqueduct_demo"]
 
 
--- a/integration_tests/sdk/conftest.py
+++ b/integration_tests/sdk/conftest.py
@@ -1,4 +1,5 @@
 import os
+from typing import List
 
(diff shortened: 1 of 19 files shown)


integration_tests/sdk/conftest.py
```diff
@@ -36,6 +30,16 @@ def fetch_flags(pytestconfig):
     yield
 
 
+def all_data_integration_names() -> List[str]:
+    """We currently only support the demo database, but this can eventually be configured arbitrarily."""
+    return ["aqueduct_demo"]
+
+
+@pytest.fixture(scope="session", params=all_data_integration_names())
```
hsubbaraj-spiral, line 38: How is the `params` parameter used here? I'm not too familiar with the usage of this within fixtures.

hsubbaraj-spiral (approved):
Looks good, just left a question for my understanding!

integration_tests/sdk/conftest.py
kenxu95: it's referenced by `request.param` in the body

PR #657, opened 2022-11-02
Surface requirement installation error during operator execution

cw75:
Previously, dependency installation errors were ignored and we proceed to executing the operator, which can cause confusing errors. Now, if any dependency installation fails, we surface the error by writing the installation error log to the operator metadata file and mark the operator as "failed". The implementation involves changing the `prune_requirements.py` to `install_requirements.py` where it does both the pruning and requirement installation, and uses our utils library to write the error to the operator metadata file.


src/python/aqueduct_executor/operators/function_executor/install_requirements.py
```diff
@@ -0,0 +1,63 @@
...
+    with open(requirements_path, "r") as f:
+        required = f.read().split("\n")
+
+    missing = []
+    for r in required:
+        # Remove any @ file because we may not have those files local to the user's device in our file system.
+        if r not in local_req and "@ file" not in r:
+            missing.append(r)
+
+    if len(missing) > 0:
```
likawind, line 46: I think we should use an if-else to handle `len(missing) == 0` in `main()`, otherwise the path will be empty and `run()` would fail.

cw75: I think in the start script we are always inputting the `missing_path` argument, so the path will never be empty? Also the missing dependency installation happens only when `len(missing) > 0`, so it seems that it won't attempt to touch that missing file when there's no missing dependency?

likawind: The issue is if `len(missing) > 0` the file is not touched it will not be created at all. This line is fine, but on 14 it will error since the path is never created. We should check if the path exists or if `len(missing) > 0` before call `install_missing_packages()`.

src/python/aqueduct_executor/start-function-executor.sh
```diff
@@ -11,13 +11,9 @@ if [ $EXIT_CODE != "0" ]; then exit $(($EXIT_CODE)); fi
 if test -f "$FUNCTION_EXTRACT_PATH/op/requirements.txt"
 then
       python3 -m pip freeze >> "$FUNCTION_EXTRACT_PATH/op/local_deps.txt"
-      python3 -m aqueduct_executor.operators.function_executor.prune_requirements --local_path="$FUNCTION_EXTRACT_PATH/op/local_deps.txt" --requirements_path="$FUNCTION_EXTRACT_PATH/op/requirements.txt" --missing_path="$FUNCTION_EXTRACT_PATH/op/missing.txt"
+      python3 -m aqueduct_executor.operators.function_executor.install_requirements --local_path="$FUNCTION_EXTRACT_PATH/op/local_deps.txt" --requirements_path="$FUNCTION_EXTRACT_PATH/op/requirements.txt" --missing_path="$FUNCTION_EXTRACT_PATH/op/missing.txt" --spec "$JOB_SPEC"
       EXIT_CODE=$?
       if [ $EXIT_CODE != "0" ]; then exit $(($EXIT_CODE)); fi
-      if test -f "$FUNCTION_EXTRACT_PATH/op/missing.txt"
```
likawind, line 17: @cw75 this is where we check if the file exists, I don't think you have the equivalent check after your change

src/python/aqueduct_executor/operators/function_executor/install_requirements.py
likawind: Oh actually I realized you call `install_missing_packages()` under this block, so everything should work!

likawind (approved):

PR #658, opened 2022-11-03
[4/4] Eng 1802 add mongodb integration - UI

likawind:
## Describe your changes and why you are making these changes
This PR adds UI changes to properly render mongo integrations and workflows. Main changes includes:
* integration cards for connecting and rendering mongo
* additional `switch` statements in 'workflow on integration' section and operator detail page to show parameters for `find()`.
* To accomplish above, we moved some bulk of code to its own component in operator detail page.

## Related issue number (if any)
ENG-1802

## Loom demo (if any)
Integration mangement: https://www.loom.com/share/47cbe3032af644cab65963077e399573
Workflow management: https://www.loom.com/share/6d6305ca0acc45f6b67b26a25e661bdc

## Checklist before requesting a review
- [ ] I have created a descriptive PR title. The PR title should complete the sentence "This PR...".
- [ ] I have performed a self-review of my code.
- [ ] I have included a small demo of the changes. For the UI, this would be a screenshot or a Loom video.
- [ ] If this is a new feature, I have added unit tests and integration tests.
- [ ] I have run the integration tests locally and they are passing.
- [ ] I have run the linter script locally (See `python3 scripts/run_linters.py -h` for usage).
- [ ] All features on the UI continue to work correctly.
- [ ] Added one of the following CI labels:
    - `run_integration_test`: Runs integration tests
    - `skip_integration_test`: Skips integration tests (Should be used when changes are ONLY documentation/UI)

PR #659, opened 2022-11-03
[3/4] Eng 1802 add mongodb integration - Sdk

likawind:
This PR implements SDK APIs to interact with mongo integration. The user should use `mongo_integration.collection()` to access the collection and use `.find()` method similar to what's exposed by mongo API: https://www.mongodb.com/docs/manual/tutorial/query-documents/ .

Full API examples:
```
client.create_param("nationality", default="United Kingdom")
mongo = client.integration(name="mongo_sample")
mongo_reviews = mongo.collection("reviews").find(
    {"reviewer_nationality": " {{nationality}} "},
    projection={"_id": 0, "hotel_name": 1, "review": 1},
    name = "my_mongo_query",
)
mongo_reviews.get()

mongo_reviews.save(config=mongo.config(table="review_copied", update_mode=aqueduct.enums.LoadUpdateMode.REPLACE))
# or
mongo_reviews.save(config=mongo.review_copied.config(update_mode=aqueduct.enums.LoadUpdateMode.REPLACE))
```

More demos see #658 
## Loom demo (if any)
see #658 


sdk/aqueduct/integrations/mongodb_integration.py
```diff
@@ -0,0 +1,171 @@
...
+    _dag: DAG
+
+    def __init__(self, dag: DAG, metadata: IntegrationInfo, collection_name: str) -> None:
+        self._metadata = metadata
+        self._dag = dag
+        self._collection_name = collection_name
+
+    def find(
+        self,
+        *kargs: List[Any],
```
kenxu95, line 38: nit: `kargs` seems like a misnomer, shouldn't it be just `args`? It's not taking in keyword arguments. Also I think you just need to type it as `Any`, since that type will apply to every elem in the list.

sdk/aqueduct/integrations/mongodb_integration.py
```diff
@@ -0,0 +1,171 @@
...
+    OperatorSpec,
+    RelationalDBLoadParams,
+    SaveConfig,
+)
+from aqueduct.utils import artifact_name_from_op_name, generate_uuid
+
+from aqueduct import globals
+
+
+class MongoDbCollectionIntegration(Integration):
```
kenxu95, line 27: naming nit: Can we capitalize "DB" throughout?

sdk/aqueduct/integrations/mongodb_integration.py
```diff
@@ -0,0 +1,171 @@
...
+    def __getattr__(self, name: str) -> Any:
+        """
+        Overrided `__get_attr__` allows caller to access integration by
+        arbitrary attribute with similar experience as mongo API:
+
+        mongo_integration = client.integration("my_integration_name")
+        my_table_artifact = mongo_integration.my_table.find({})
+
+        """
+        return MongoDbCollectionIntegration(self._dag, self._metadata, name)
```
kenxu95, line 160: this is neat! It makes me wonder if we can do a similar thing when previewing operators with custom backends:
```
@op
def foo(sql):
    ...

foo.my_lambda_integration(input_artf)
```

kenxu95: Might be a bit too much magic though, now that I think about it, but it would be consistent with this. I'm curious if there was a reason you went for this proposal instead of:
```
mongo = client.integration(name="mongo_sample")
mongo.collection("reviews").find(...)
```
Since ^ actually seems more understandable to me.

sdk/aqueduct/integrations/sql_integration.py
```diff
@@ -54,6 +47,28 @@ def replace_today() -> str:
     return "'" + date.today().strftime("%Y-%m-%d") + "'"
 
 
+def find_parameter_names(query: str) -> List[str]:
+    matches = re.findall(TAG_PATTERN, query)
+    return [match.strip(" {}") for match in matches]
+
+
+def find_parameter_artifacts(dag: DAG, names: List[str]) -> List[ArtifactMetadata]:
```
kenxu95, line 55: The semantics are a little unexpected here, since `len(names) != len(returned artifacts)` in some cases. Can you put a comment here that not all names will have a corresponding artifact, and why that would be the case?

kenxu95 (changes requested):
Code-wise it looks good! Just had some nits/readability comments for it.

Blocking since I wanted to get product's perspective on the proposed API vs the alternative.  (See last comment in [mongodb_integration.py](https://github.com/aqueducthq/aqueduct/pull/659/files/92d00c14ef97e6ce4e65468aeed3c297155674cc#diff-c9a5937672466aa2fb205c959ef6c4f9ebd0699bdf33848414595404e1aa39a4))

sdk/aqueduct/integrations/mongodb_integration.py
likawind: This is how you access collection using a mongo's db object: https://www.mongodb.com/docs/manual/tutorial/query-documents/ so the overall experience is the same.

I don't think we should generalize this to other use cases as I myself find it confusing to learn. I also find what you recommend more intuitive. So it's more a question of consistent with Mongo (who's not consistent with other relational DB :) ) or consistent with us. I slightly prefer the former as we need to be mongo-consistent in the `.find()` part anyway

likawind: @vsreekanti would like to get your thoughts as well ^^

sdk/aqueduct/integrations/mongodb_integration.py
likawind: Actually I found we are using `Db` for all variable / class names throughout our codebase:

Screen Shot 2022-11-03 at 11 39 23 AM

kenxu95: yes but I think the canonical way to style it is `DB`, similar to how we should use `ID` and not `Id` (which I don't think is consistent in our codebase). Or do you want to hold off on the rename and do it all at one?

sdk/aqueduct/integrations/mongodb_integration.py
kenxu95: I had the same exact intuition - consistent for us vs consistent with Mongo. I actually have a slight preference for the latter, since I think objectively the pattern is cool but weird and I doubt its natural enough for users to try and generalize this pattern to other APIs. Also, I value keeping our API more consistent. But yea lets just chat at the sync.

sdk/aqueduct/integrations/mongodb_integration.py
likawind: based on https://learn.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/141e06ef(v=vs.71)?redirectedfrom=MSDN , `DB` is the way to go so I will move all mongo code to use `DB`. Let's migrate other ones separately.

sdk/aqueduct/integrations/sql_integration.py
```diff
@@ -54,6 +47,37 @@ def replace_today() -> str:
...
+    """
+    `find_parameter_artifacts` finds all parameter artifacts corresponding to given `names`.
+    parameters:
+        names: the list of names, repeating names are allowed.
+    returns:
+        a list of unique parameter artifacts for these names. Built-in names are omitted.
```
kenxu95, line 61: Curious what python commenting style you're following? I've been trying to follow the google style guide throughout the SDK (Args, Returns, Raises): https://google.github.io/styleguide/pyguide.html#244-decision

Again this is probably something we should standardize in our design doc.

kenxu95 (approved):

sdk/aqueduct/integrations/mongodb_integration.py
```diff
@@ -0,0 +1,169 @@
...
+    my_table_artifact = mongo_integration.collection("my_collection").find({})
+    """
+
+    def __init__(self, dag: DAG, metadata: IntegrationInfo):
+        self._dag = dag
+        self._metadata = metadata
+
+    def collection(self, name: str) -> Any:
+        """
+        `collection` returns a specific collection object to call `.find()` method.
```
kenxu95, line 152: 1) I think starting comments with `<name of function>` is a golang thing. Might be worth pattern matching what's in, for example, `aqueduct_client.py`?

2) Doesn't this return a `MongoDBCollectionIntegration`? I don't think the typing should be `Any`.

sdk/aqueduct/integrations/sql_integration.py
likawind: Cool that's good to know! Let's add them to our guide doc?

likawind:
skipping integration test now, we will do one altogether when merging to main

PR #661, opened 2022-11-03
[1/4] Eng 1802 add mongodb integration - Python Executor

likawind:
This PR implements executors for mongo integration. We call corresponding mongo APIs for all methods exposed on a relational integration. It'd be helpful to look at #659 for SDK inputs as a reference.

ENG-1802

## Loom demo (if any)
See #658 
We also verified `aqueduct install mongo` works.


src/python/aqueduct_executor/operators/connectors/data/extract.py
```diff
@@ -145,4 +147,22 @@ class S3Params(models.BaseParams):
     merge: Optional[bool]
 
 
-Params = Union[RelationalParams, S3Params]
+class MongoDbFindParams(models.BaseParams):
+    kargs: Optional[List[Any]] = None
```
kenxu95, line 151: another instance where it should be "args"? This isn't a MongoDB specific thing is it?

src/python/aqueduct_executor/operators/connectors/data/mongodb.py
```diff
@@ -0,0 +1,102 @@
...
+                replace = False
+                if exists:
+                    if params.update_mode == common.UpdateMode.FAIL:
+                        raise Exception(f"Specified table {params.table} already exists.")
+
+                    collection = db[params.table]
+                    if params.update_mode == common.UpdateMode.REPLACE:
+                        replace = True
+
+                if not exists:
```
kenxu95, line 83: nit: isn't this just an "else" statement?

kenxu95: I also don't think you need the "collection = None" statement, because of the assert on line 86.

kenxu95 (approved):
I don't know enough about Mongo to provide anything, but very surface-level critiques. I think the code makes sense though. 

Do you think our query-expanding code is hard to work with? I've always found it a bit of a weird pattern (since we have to remember to call `.usable()` everywhere).

likawind:
@kenxu95 thanks! I think the query expanding is fine for now as we are not using it a lot

PR #670, opened 2022-11-03
[3/N] Sets up testing for repos

saurav-c:
This PR sets up database integration tests for all repos.

The key decisions made are:
- Moving all tests to `repos/tests`
- Each repo should have a corresponding test file named, e.g. `workflow.go`. Inside this test file will be a test case for each repo method, such as `TestWorkflow_Exists`, `TestWorkflow_Get`.
- Any `seed` functions, such as `seedWorkflow` or `seedUser` or `seedDAGWithWorkflow` should be implemented in `seed.go`. These are essentially helper methods that each test can call to quickly generate some test data.
- The database will be wiped after EACH test case. This follows the current approach we have. This is because we just want to verify that each query is correct.

### DIFF
--- a/src/golang/lib/models/schema.go
+++ b/src/golang/lib/models/schema.go
@@ -1,0 +1,4 @@
+package models
+
+// SchemaVersion is the current database schema version.
+const SchemaVersion = 19
--- a/src/golang/lib/repos/tests/suite.go
+++ b/src/golang/lib/repos/tests/suite.go
@@ -22,32 +22,32 @@ type TestSuite struct {
 	// List of all repos
 	workflow repos.Workflow
 
-	db database.Database
+	DB database.Database
 }
 
 // SetupSuite is run only once before all tests. It initializes the database
 // connection and creates the repos. It initializes the database schema
 // to the latest version.
 func (ts *TestSuite) SetupSuite() {
-	db, err := database.NewSqliteInMemoryDatabase(&database.SqliteConfig{})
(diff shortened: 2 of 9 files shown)


src/golang/lib/repos/tests/suite.go
```diff
@@ -0,0 +1,100 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// TearDownTest is run after each test finishes.
+func (ts *TestSuite) TearDownTest() {
+	// Clear all of the tables
+	query := `
```
eunice-chan, line 78: Does this delete everything in the tables? I wonder if there is some way to just delete anything created from the tests..

eunice-chan (approved):

src/golang/lib/repos/tests/suite.go
saurav-c: Yeah, this deletes everything, which technically is only things created by the tests since the db is cleared after each test.

src/golang/lib/models/schema.go
```diff
@@ -0,0 +1,4 @@
+package models
+
+// SchemaVersion is the current database schema version.
```
kenxu95, line 3: Drive-by-comment from your slack message :P Can you put a big capital note here that this is the source of truth? (with whatever caveats there are)

saurav-c: Sure will add in a diff related PR I'm about to merge.

PR #678, opened 2022-11-07
Verified that the requirement detection hierarchy is already enforced

cw75:
I verified that the proposed requirement detection rules are already enforced. This PR mainly tidies up some comments.

The only functionality change I made is to add client's cloudpickle version to the function's requirements file. This is useful when we aren't using `pip freeze`, because when the user manually specifies the requirements, they are unlikely to include the cloudpickle entry, but for us, we need to ensure that the server installs the same version of cloudpickle as the client, so adding this entry will prevent potential serialization bugs due to cloudpickle version mismatch.

### DIFF
--- a/sdk/aqueduct/utils.py
+++ b/sdk/aqueduct/utils.py
@@ -13,6 +13,7 @@ from typing import Any, Callable, Dict, List, Mapping, Optional, Union
 import cloudpickle as pickle
 import multipart
 import numpy as np
+import pkg_resources
 import requests
 from aqueduct.config import (
     AirflowEngineConfig,
@@ -326,6 +327,8 @@ def _package_files_and_requirements(
     # Prune out any blacklisted requirements.
     _filter_out_blacklisted_requirements(packaged_requirements_path)
 
+    _add_cloudpickle_to_requirements(packaged_requirements_path)
+
     os.chdir(current_directory_path)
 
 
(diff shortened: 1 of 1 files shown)


likawind (approved):
lgtm. So for now we will just overwrite the server machine's cp version if there's a mismatch?

cw75:
Yes, unless the user registers the Conda integration. In that case we won't mess around with the server's default Python env.

sdk/aqueduct/utils.py
```diff
@@ -340,8 +344,26 @@ def _filter_out_blacklisted_requirements(packaged_requirements_path: str) -> Non
...
+    with open(packaged_requirements_path, "a") as f:
+        cloudpickle_requirement = (
+            "\ncloudpickle==%s" % pkg_resources.get_distribution("cloudpickle").version
+        )
+        f.write(cloudpickle_requirement)
+
+
 def _infer_requirements() -> List[str]:
-    """Obtains the list of pip requirements specifiers from the current python environment using `pip freeze`.
+    """
```
kenxu95, line 365: I think it's time we settle on a comment style in python. The previous comment follows the Google style guide: https://google.github.io/styleguide/pyguide.html#383-functions-and-methods, where the first comment is on the same line as `"""`. Can we keep this consistent? Either we keep it this way or change it all.

PR #682, opened 2022-11-07
Add ability for tests to run against multiple backing engines

kenxu95:
@hsubbaraj-spiral  as primary reviewer. @Fanjia-Yan this should help you test out your changes in a more automated fashion.

Adds an `engine` fixture to our SDK integration test suite. This value can be requested by any test:
```
def test_something(client, engine):
      ...
      client.publish_flow(..., engine=engine)
```

This `engine` value can be configured through the command line, and corresponds to the name of the compute integration we want to run the test suite against!

For example, if our server has connected to a K8s integration called "my_k8s_integration", I could run the test suite with:
```
 API_KEY=Q96DTSL0MNRCVO3YF2IJX1WAKPEH8U45 SERVER_ADDRESS=localhost:8080 pytest . -rP -vv --engine "my_k8s_integration" -n 8
```
And it will run all relevant tests agains the K8s compute backend.

If `--engine` cmd line arg is not supplied, we will default to `aqueduct_demo`.

## Testing
Tested that this works normally when `--engine` is not supplied. Tested that supplying the wrong integration name to `engine` will cause the tests to fail at publish time.

The next step in my work is to test the suite against the K8s integration, which will set the groundwork for testing CPU and Memory limits.

ENG-1952

### DIFF
--- a/integration_tests/sdk/conftest.py
+++ b/integration_tests/sdk/conftest.py
@@ -11,7 +11,6 @@ import aqueduct
 # The option variable can be accessed through utils.flags during tests.
 # One can also mark test to be triggered only when `--{flag}` is turned on through
 # @pytest.mark.<flag_name>
-from sdk.constants import AQUEDUCT_ENGINE
 
 FLAGS = []
 
--- a/integration_tests/sdk/conftest.py
+++ b/integration_tests/sdk/conftest.py
@@ -2,33 +2,17 @@ import os
 from typing import List
 
 import pytest
-import utils
 from aqueduct.dag import DAG, Metadata
 
 import aqueduct
 
-# Usage: add a <flag> in FLAGS which will enable `--{flag}` in test cmd options.
-# The option variable can be accessed through utils.flags during tests.
-# One can also mark test to be triggered only when `--{flag}` is turned on through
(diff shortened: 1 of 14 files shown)


integration_tests/sdk/conftest.py
```diff
@@ -2,34 +2,20 @@
...
-        parser.addoption(f"--{flag}", action="store_true", default=False)
+    # We currently only support a single engine backend at a time.
+    parser.addoption(f"--engine", action="store", default=None)
 
 
 API_KEY_ENV_NAME = "API_KEY"
 SERVER_ADDR_ENV_NAME = "SERVER_ADDRESS"
 
 
-@pytest.fixture(autouse=True)
```
hsubbaraj-spiral, line 26: Why are we getting rid of this fixture?

hsubbaraj-spiral (approved):
left a question but otherwise looks good to me.

integration_tests/sdk/conftest.py
kenxu95: all this fixture did was stick the flags onto the global var in `utils.py`. We don't use this variable anymore so deleting.

PR #687, opened 2022-11-08
Eng 1921 Allow user to connect to conda

likawind:
This PR allows user to 'connect' to conda from UI. 'Connect' simply verify if the user has conda installed on their server and create a dummy integration for that user. This allows us to futher use conda to run operators.

ENG-1921

## Loom demo (if any)
The demo go over connect, edit, test-connect, and deletion of conda integration:
https://www.loom.com/share/00b0ba7e31564285bf36ecd1a5057967

### DIFF
--- a/src/golang/cmd/server/handler/connect_integration.go
+++ b/src/golang/cmd/server/handler/connect_integration.go
@@ -295,6 +295,11 @@ func ValidateConfig(
 		return validateLambdaConfig(ctx, config)
 	}
 
+	jobName := fmt.Sprintf("authenticate-operator-%s", uuid.New().String())
+	if service == integration.Conda {
+		return validateConda(ctx, jobManager, jobName)
+	}
+
 	// Schedule authenticate job
 	jobMetadataPath := fmt.Sprintf("authenticate-%s", requestId)
 
@@ -303,7 +308,6 @@ func ValidateConfig(
 		go utils.CleanupStorageFiles(ctx, storageConfig, []string{jobMetadataPath})
 	}()
 
-	jobName := fmt.Sprintf("authenticate-operator-%s", uuid.New().String())
 	jobSpec := job.NewAuthenticateSpec(
 		jobName,
 		storageConfig,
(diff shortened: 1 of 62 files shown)


src/golang/cmd/server/request/integration.go
```diff
@@ -43,7 +43,11 @@ func ParseIntegrationConfigFromRequest(r *http.Request) (string, map[string]stri
 
 // isUserOnlyIntegration returns whether the specified service is only accessible by the user.
 func isUserOnlyIntegration(svc integration.Service) bool {
-	userSpecific := []integration.Service{integration.GoogleSheets, integration.Github}
+	userSpecific := []integration.Service{
+		integration.GoogleSheets,
+		integration.Github,
+		integration.Conda,
```
cw75, line 49: I actually think it's possible in the future for users to share the Conda env created. But since now we're focusing on OSS, there is only one user, so this seems irrelevant for now.

src/golang/cmd/server/handler/connect_integration.go
```diff
@@ -567,3 +571,21 @@ func validateLambdaConfig(
 
 	return http.StatusOK, nil
 }
+
+func validateConda(
```
cw75, line 575: Since all we're doing is `conda --version`, I wonder if we have to go through the job manager to trigger this command. Unlike data integration validations, `conda --version` is a local operation that returns immediately, so it doesn't really make sense to fire the job and poll it. Also, we are already doing validations that bypass the job manager in our engine integrations above, so I don't think we have to restrict ourselves to using job manager here.

cw75 (commented):
Left some comments inline.

src/ui/Makefile
```diff
@@ -9,10 +9,12 @@ endif
 # This command can be used to temporarily link a version of the component
 # library that you've developed locally into the app UI.
 link-dev:
-	cd common && npm install && sudo npm link
+	cd app && npm unlink
```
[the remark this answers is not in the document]
likawind: no it's just some tricks so that I run fewer cmd to reboot my dev UI. There's no harm to get them in though

cw75:
@likawind reminder to get rid of the `aqueduct install conda` part as we discussed. Otherwise LGTM.

cw75 (approved):

PR #696, opened 2022-11-09
Enable Preview Using Different Engines

Fanjia-Yan:
This PR makes changes on SDK and backend. On backend, I modify the jobManager choice based on the incoming WorkflowDag. On SDK side, as users configure `aq.global_config({"engine": engine})`, the preview will be executed on the designated engines. 

After the implementation, as suggested by Kenny, I have also done some latency test on, the result can be found here:
https://www.notion.so/aqueducthq/Benchmark-Preview-6757fbdf3909436db5d0777dfd31b82f

## Testing
For k8s preview, I tested it by walking through the example notebooks. For each operators executed, there will be a preview request which will spawn a k8s pod with pod name "function-operator-xxx". If the pod completed, that means the preview is executed on k8s succesfully. Finally, compare it with the result executed by aqueduct engine.

For lambda preview, I did the same thing and the log of lambda function will be documented on aws console. By looking into the log the preview will be executed and written in S3. If that's successful, then the preview is executed on lambda successfully.  Finally, compare it with the result executed by aqueduct engine.

For integration test, I still need to make sure that the test works locally on lambda and k8s while the dockerfile issue has been slowing me down on that.

https://linear.app/aqueducthq/issue/ENG-1948/refactor-aqenginepreviewworkflow-to-allow-preview-using-different


src/golang/lib/engine/aq_engine.go
```diff
@@ -342,11 +342,18 @@ func (eng *aqEngine) PreviewWorkflow(
 	timeConfig *AqueductTimeConfig,
 ) (*WorkflowPreviewResult, error) {
 	// previewing workflows always happens using the ProcessJobManager
```
kenxu95, line 344: Let's remove this comment? It's wrong now right

sdk/aqueduct/artifacts/utils.py
```diff
@@ -49,6 +50,19 @@ def preview_artifacts(
         make_copy=True,
     )
 
+    if globals.__GLOBAL_CONFIG__.engine is not None:
+        engine = globals.__GLOBAL_CONFIG__.engine
+        if engine is None:
+            engine_config = EngineConfig()
+        else:
+            conntected_integrations = globals.__GLOBAL_API_CLIENT__.list_integrations()
```
kenxu95, line 58: spelling: "connected_integrations"

kenxu95 (commented):
This looks good to me! Before I stamp, can you elaborate on what steps you took to test this in the PR description?

Fanjia-Yan:
> This looks good to me! Before I stamp, can you elaborate on what steps you took to test this in the PR description?

Updated the PR description with testing detail!

sdk/aqueduct/artifacts/utils.py
```diff
@@ -49,6 +50,19 @@ def preview_artifacts(
...
+        if engine is None:
+            engine_config = EngineConfig()
+        else:
+            connected_integrations = globals.__GLOBAL_API_CLIENT__.list_integrations()
+            if engine not in connected_integrations.keys():
+                raise InvalidIntegrationException(
+                    "Not connected to compute integration %s!" % engine
+                )
+            engine_config = generate_engine_config(connected_integrations[engine])
+        subgraph.engine_config = engine_config
```
kenxu95, line 64: After rebasing, use `dag.set_engine_config()` here instead. That'll catch any invalid resource configuration errors too.

That PR also makes `engine_config` optional, but it has to be set before uploading to the server. You can see how this pattern works in `aqueduct_client.py`, but essentially we should be calling `set_engine_config()` regardless of whether the engine was set globally, and calling `assert subgraph.engine_config is not None` right before the api_client.preview() call as a safety check.

kenxu95 (changes requested):
For writing integration tests against K8s, can you rebase this onto https://github.com/aqueducthq/aqueduct/pull/701 and add your preview tests using the scaffolding I provided?

See the `resources_test.py` file for how to do this. I think we should:
1) Add a preview test against K8s in `preview_test.py` using global config.
2) Add a preview test against Lambda in `preview_test.py` using global config.
2) Add a preview step within `resources_test.py`. Essentially, remove all the `.lazy()` calls and after configuring globally :) 

As long as these new tests are working, we should be good.

Fanjia-Yan:
One thing I thought could improve testing the preview is to make sure that k8s pods/ lambda function are actually spawned and executed correctly. The current implementation of the integration test is based on the assumption that `aq.global_config` always behaves correctly. Maybe one way I have in mind is to use command line "kubectl" to locates the pods spawned? @kenxu95 Do you think that will be worthy or feasible to implement?

integration_tests/sdk/preview_test.py
```diff
@@ -109,3 +111,47 @@ def bad_return():
...
+    assert output_artifact._dag.engine_config.type == RuntimeType.LAMBDA
+    output_df = output_artifact.get()
+    assert list(output_df) == [
+        "hotel_name",
+        "review_date",
+        "reviewer_nationality",
+        "review",
+        "positivity",
+    ]
+    assert output_df.shape[0] == 100
```
kenxu95, line 157: Let's not duplicate the exact same test for different integrations. That's why we have this `enable_only_for_engine_types()` in the first place.

To get this merged into one test, just use:
`@pytest.mark.enable_only_for_engine_type(ServiceType.K8S, ServiceType.LAMBDA)`
And update the `enable_by_engine_type()` function in `conftest.py` to handle multiple engine types. Start by changing `enabled_engine_type = request.node.get_closest_marker("enable_only_for_engine_type").args[0]` to `enabled_engine_types = request.node.get_closest_marker("enable_only_for_engine_type").args` and dealing with the fallout in the rest of the method.

Let me know if you want to chat about this change, its pretty straightforward so I want you to take a crack at it.

kenxu95 (changes requested):
Thanks for the tests! Blocking due to two issues.

sdk/aqueduct/artifacts/utils.py
```diff
@@ -35,6 +36,20 @@ def preview_artifacts(
...
+            engine_config = EngineConfig()
+        else:
+            connected_integrations = globals.__GLOBAL_API_CLIENT__.list_integrations()
+            if engine not in connected_integrations.keys():
+                raise InvalidIntegrationException(
+                    "Not connected to compute integration %s!" % engine
+                )
+            engine_config = generate_engine_config(connected_integrations[engine])
+        dag.engine_config = engine_config
+        assert dag.engine_config is not None
```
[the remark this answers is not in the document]
kenxu95: Chatted with @Fanjia-Yan:
- The `set_engine_config()` comment still holds.
- It turns out that leaving the `engine_config` field on the DAG (SDK side) empty doesn't matter, since its backfilled to the default engine config on the backend: https://github.com/aqueducthq/aqueduct/blob/main/src/golang/cmd/server/request/dag.go#L76. @Fanjia-Yan could you add a comment on the field in `dag.py` that a None value means default aqueduct server?

sdk/aqueduct/dag.py
```diff
@@ -45,9 +45,10 @@ class DAG(BaseModel):
     # Is excluded from json serialization.
     operator_by_name: Dict[str, Operator] = {}
 
-    # The fields below must be set when publishing the workflow
+    # The fields below must be set when publishing the workflow.
```
kenxu95, line 48: nit: `engine_config` no longer needs to be set, so lets just say "This field must be ..."

kenxu95 (approved):
Thanks @Fanjia-Yan! 

Final small testing request, could you rebase onto latest main and add your preview check to the two tests recently committed in `resources_test.py` here: https://github.com/aqueducthq/aqueduct/pull/720/files 

Thanks! Instead of removing `.lazy()` it'll just be calling those methods without `.lazy()` and making sure they error (I think AqueductError is what is thrown)

PR #697, opened 2022-11-09
Fix start-function-executor mismatch bug between aqueduct_executor and dockerfiles

kenxu95:
Couple questions here:
1) Why didn't this fail before? This was changed 6 days. Is it because we didn't update the dockerhub image for 3.8? @cw75 
2) Why are these two dockerfiles separate at all?

### DIFF
--- a/src/dockerfiles/function/start-function-executor.sh
+++ b/src/dockerfiles/function/start-function-executor.sh
@@ -1,6 +1,5 @@
 #!/bin/bash
-#!/bin/bash
-echo $JOB_SPEC
+JOB_SPEC=$1
 FUNCTION_EXTRACT_PATH=$(python3 -m aqueduct_executor.operators.function_executor.get_extract_path --spec "$JOB_SPEC")
 EXIT_CODE=$?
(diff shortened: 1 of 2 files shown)


hsubbaraj-spiral (changes requested):

kenxu95:
@hsubbaraj-spiral I removed the job spec line. Do you want me to also try to consolidate the two scripts in this PR? It's not super clear to me how to do that without messing around with some assumptions. For example, if we chose the script in `aqueduct_executor`, we would no longer be able to build from the `dockerfiles/` directory, but instead would have to go one level up.

It seems to me the quickest way forward that makes sure this doesn't happen again is to but a banner comment at the top of each copy of `start-function-executor.sh` indicating that changes should be kept in sync? What do you think

cw75:
Maybe one way is to edit the docker build process to: 1) copy the start script from `aqueduct_executor` to the dockerfile directory, 2) build the image, and 3) clean up the temporarily copied start script. If we want to do this later in a separate PR, for now I think the easiest workaround is to clearly comment that these two files need to be kept in sync.

hsubbaraj-spiral (approved):

PR #701, opened 2022-11-11
Add ability to configure memory and number of CPUs on K8s backend

kenxu95:
@hsubbaraj-spiral as lead reviewer.

This PR allows for users to configure operators with a custom number of CPUs and memory allowances. The API looks like:
```
@op(resources={"num_cpus": 2, "memory": "5Gb"})
def foo():
    ...
```
The "memory" field can be either an integer or a string. If it's a string, only the "gb" and "mb" suffixes are allowed (case-insensitive). If it's an integer, we assume it's in MBs.
If a user attempts to publish this kind of operator in a flow that doesn't use K8s, we will throw an exception.

### Testing ###
We introduce a new `pytest` marker. This one will filter a test to run on only a single type of executor engine. For example:
```
# This will run when `--engine` flag is set to an integration that is Kubernetes.
@pytest.mark.enable_only_for_engine_type(ServiceType.K8s)
def test_k8s():
    ...
```
This allows us to not impede all our local server tests, while also writing down and committing test cases for K8s.

This PR will include the following tests:
- [x] Test that number of CPUs can be configured exactly.
- [x] Test the memory available to an operator can be configured exactly.
- [x] Test that OOM issues during runtime will fail the workflow.
- [x] Unit test the handling of the memory string: ` _convert_memory_string_to_mbs()`
- [x] Unit and Integration tests that K8s is the only engine that we can set custom resources for.

### Additional issues out of scope of this PR. ###
These mostly have to do with the user experience when something fails:
- [ ] Test that when the number of CPUs requested is more than the capacity, we fail quickly and gracefully. Currently,
we simply hang...
- [ ] Same as above, but for memory.
- [ ] Test that when we have a OOM error during execution, we surface this error message back to the user, instead of a generic "system error". 

We will create tasks for this.

ENG-1951

### DIFF
--- a/sdk/aqueduct/dag.py
+++ b/sdk/aqueduct/dag.py
@@ -3,8 +3,12 @@ from typing import Dict, List, Optional
 
 from aqueduct.artifacts.metadata import ArtifactMetadata
 from aqueduct.config import EngineConfig
-from aqueduct.enums import ArtifactType, OperatorType, TriggerType, RuntimeType
-from aqueduct.error import ArtifactNotFoundException, InternalAqueductError, InvalidUserArgumentException
+from aqueduct.enums import ArtifactType, OperatorType, RuntimeType, TriggerType
+from aqueduct.error import (
+    ArtifactNotFoundException,
+    InternalAqueductError,
+    InvalidUserArgumentException,
+)
 from aqueduct.operators import (
     Operator,
     OperatorSpec,
@@ -33,7 +37,6 @@ class Metadata(BaseModel):
     retention_policy: Optional[RetentionPolicy]
 
 
-
 class DAG(BaseModel):
     operators: Dict[str, Operator] = {}
     artifacts: Dict[str, ArtifactMetadata] = {}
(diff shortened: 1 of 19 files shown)


sdk/aqueduct/dag.py
```diff
@@ -50,6 +54,36 @@ class Config:
...
+            if op.spec.resources is not None:
+                if not allowed_customizable_resources["num_cpus"] and op.spec.resources.num_cpus:
+                    raise InvalidUserArgumentException(
+                        "Operator %s cannot configure the number of cpus, since it is not supported when running on %s."
+                        % (op.name, self.engine_config.type)
+                    )
+                if not allowed_customizable_resources["memory"] and op.spec.resources.memory_mb:
+                    raise InvalidUserArgumentException(
+                        "Operator %s cannot configure the amount of memory, since it is not supported when running on %s."
+                        % (op.name, self.engine_config.type)
```
kenxu95, line 84: Now that I think about it, there's no reason this shouldn't go within `publish_flow()` directly as a helper. This `dag.verify()` stuff is confusing. Will move.

hsubbaraj-spiral (approved):

PR #710, opened 2022-11-14
Disable Conda from UI and remove Conda install option

cw75:
We will re-enable the Conda integration option once we finish the backend implementation.
The Conda install option needs to be removed anyway. This should have been done in the previous PR but we forgot to remove it.

### DIFF
--- a/src/ui/common/src/utils/integrations.ts
+++ b/src/ui/common/src/utils/integrations.ts
@@ -351,11 +351,14 @@ export const SupportedIntegrations: ServiceInfoMap = {
     activated: true,
     category: 'data',
   },
+  // TODO: enable this once we release conda
+  /*
   ['Conda']: {
     logo: 'https://aqueduct-public-assets-bucket.s3.us-east-2.amazonaws.com/webapp/pages/integrations/conda.png',
     activated: true,
     category: 'compute',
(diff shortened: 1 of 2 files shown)


src/ui/common/src/utils/integrations.ts
```diff
@@ -351,11 +351,14 @@ export const SupportedIntegrations: ServiceInfoMap = {
     activated: true,
     category: 'data',
   },
+  // TODO: enable this once we release conda
+  /*
   ['Conda']: {
     logo: 'https://aqueduct-public-assets-bucket.s3.us-east-2.amazonaws.com/webapp/pages/integrations/conda.png',
     activated: true,
```
vsreekanti, line 358: @cw75, can't we just set `activated` to `false` here?

cw75: If we set it to `false`, I think the UI will still show the Conda box but in a greyed-out state. If we're ok with this we can do that.

agiron123: Perhaps we can try to add a hidden flag in the future, but that would effectively do the same thing here.

agiron123 (approved):

PR #714, opened 2022-11-14
Handle OOM exceptions gracefully in Kubernetes

kenxu95:
@hsubbaraj-spiral and @cw75 as lead reviewers. 

@saurav-c and @likawind for any additional high level discussion/thoughts around this.

Currently, when an operator runs out of memory, we throw a generic system error that doesn't tell the user anything. However, we can actually detect OOM errors through the K8s interface, and present the user with a much more helpful error:
image

This PR is important not because of the bug (which can be argued is an edge-case), but for its broader application to the JobManager interface. Essentially, do we want the JobManager interface to return a special error type, encoding stack trace info + additional context, back to the orchestrator?

The current rules are:
1) If it is a user-caused issue (eg. OOM), then we should print a nice message for them, without the stack trace - since the stack trace points within our code.
2) If it is a system issue, we should(?) dump both the message and the stack trace context for debugging purposes on the UI. I think it is ok to be overly transparent here. We *are* currently already doing this, but, because we define our errors beforehand, this way: https://github.com/aqueducthq/aqueduct/blob/main/src/golang/lib/job/job.go#L11 we lose the real stack trace and instead get something like
```
ORIGINAL STACK TRACE:
github.com/aqueducthq/aqueduct/lib/engine.init
        /Users/kennethxu/work/aqueduct/src/golang/lib/engine/engine.go:25 +0x5477cee
runtime.doInit
        /usr/local/go/src/runtime/proc.go:6222 +0x4048005
runtime.doInit
        /usr/local/go/src/runtime/proc.go:6199 +0x4047f50
runtime.doInit
        /usr/local/go/src/runtime/proc.go:6199 +0x4047f50
runtime.main
        /usr/local/go/src/runtime/proc.go:233 +0x403b232
runtime.goexit
        /usr/local/go/src/runtime/asm_amd64.s:1571 +0x4068c80
```
which is much less useful than
```
ORIGINAL STACK TRACE:
github.com/aqueducthq/aqueduct/lib/job.(*k8sJobManager).Poll
        /Users/kennethxu/work/aqueduct/src/golang/lib/job/k8s.go:168 +0x544434a
github.com/aqueducthq/aqueduct/lib/workflow/operator.(*baseOperator).Poll
        /Users/kennethxu/work/aqueduct/src/golang/lib/workflow/operator/base.go:235 +0x5464ccd
github.com/aqueducthq/aqueduct/lib/engine.(*aqEngine).execute
        /Users/kennethxu/work/aqueduct/src/golang/lib/engine/aq_engine.go:754 +0x5474c0a
github.com/aqueducthq/aqueduct/lib/engine.(*aqEngine).ExecuteWorkflow
        /Users/kennethxu/work/aqueduct/src/golang/lib/engine/aq_engine.go:319 +0x546fe67
github.com/aqueducthq/aqueduct/cmd/executor/executor.(*WorkflowExecutor).Run
        /Users/kennethxu/work/aqueduct/src/golang/cmd/executor/executor/workflow.go:80 +0x547aefb
main.run
        /Users/kennethxu/work/aqueduct/src/golang/cmd/executor/main.go:101 +0x547cd39
main.main
        /Users/kennethxu/work/aqueduct/src/golang/cmd/executor/main.go:82 +0x547cb07
runtime.main
        /usr/local/go/src/runtime/proc.go:250 +0x403b271
runtime.goexit
```
The proposal for the future is that we get rid of these pre-defined errors in favor of the more flexible `JobError`, defined in this PR. All JobManager methods will return this `JobError`, which is just an error with more context.

ENG-2001

## Loom demo (if any)
https://www.loom.com/share/e974f718eed44f928ac7080e74ac4343

### DIFF
--- a/integration_tests/sdk/resources_test.py
+++ b/integration_tests/sdk/resources_test.py
@@ -75,6 +75,18 @@ def test_custom_num_cpus(client, engine):
 
 
 @pytest.mark.enable_only_for_engine_type(ServiceType.K8S)
+def test_too_many_cpus_requested(client, engine):
+    """Assumption: nodes in the k8s cluster have less then 20 CPUs."""
+
+    @op(requirements=[], resources={"num_cpus": 20})
+    def too_many_cpus():
+        return 123
+
+    output = too_many_cpus.lazy()
+    run_flow_test(client, [output], engine=engine, expect_success=False)
+
+
+@pytest.mark.enable_only_for_engine_type(ServiceType.K8S)
 def test_custom_memory(client, engine):
     """Assumption: nodes in the K8s cluster have more than 200MB of capacity.
 
@@ -110,3 +122,15 @@ def test_custom_memory(client, engine):
         engine=engine,
         expect_success=False,
     )
+
+
+@pytest.mark.enable_only_for_engine_type(ServiceType.K8S)
+def test_too_much_memory_requested(client, engine):
+    """Assumption: nodes in the k8s cluster have less then 100GB of memory."""
+
+    @op(requirements=[], resources={"memory": "100GB"})
+    def too_much_memory():
+        return 123
+
+    output = too_much_memory.lazy()
+    run_flow_test(client, [output], engine=engine, expect_success=False)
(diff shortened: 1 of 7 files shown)


hsubbaraj-spiral (approved):
Looks good as per our discussion earlier!

PR #717, opened 2022-11-15
Enables GPUs on K8s engine

hsubbaraj-spiral:
This PR enables users to run workflows on GPUS via the following:
```
@op(resources={"gpu_resource_name":"nvidia.com/gpu"})
def join(a, b) -> pd.DataFrame:
    return b
```


src/dockerfiles/gpu/py310_env.yml
```diff
@@ -0,0 +1,17 @@
...
+  - pip==22.0.4
+  - scipy==1.7.3
+  - cloudpickle==2.0.0
+  - pyarrow==7.0.0
+  - boto3==1.18.0
+  - pydantic==1.9.0
+  - pip:
+    - scikit_learn==1.0.2
+    - typing_extensions==4.3.0
+    - aqueduct-ml==0.1.4
```
kenxu95, line 17: These requirements are only needed for the docker build process right? The user's function requirements will overwrite these at execution time?

kenxu95 (changes requested):
I don't have any gripes with the code itself, but blocking on some higher level questions:

Can you remind me why we didn't go for
```
@op(resources={"use_gpus": True})
```
? Was it cus it's impossible to figure out what GPU types are attached to the cluster?

Couple questions around testing: 
- Is that going to be a follow up or included in this PR? 
- What happens if the user makes a typo/enters the wrong GPU resource string in the decorator.
- What happens if the user enters a value GPU resource, but that resource isn't available on the backend.

kenxu95 (approved):

PR #728, opened 2022-11-18
Fix Run Notebook Fail Resulted From Python Setup

Fanjia-Yan:
After some digging, I think I get what’s going on. So we run notebooks on ubuntu-latest but now there is a migration that changes ubuntu-latest  version from 20.04 to 22.04. https://github.com/actions/runner-images/issues/6399. However, there is no python version of 3.8.10 of linux 22.04 which causes the failure.

One way we can make this work is to either hard coded the ubuntu version or we can change the python version that has linux 22.04 version.

### DIFF
--- a/.github/workflows/run-notebooks.yml
+++ b/.github/workflows/run-notebooks.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Set up Python 3.8
         uses: actions/setup-python@v2
         with:
-          python-version: '3.8.10'
+          python-version: '3.8'
 
       - name: Set up GOPATH variable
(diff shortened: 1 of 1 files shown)


Fanjia-Yan:
Current Run: https://github.com/aqueducthq/aqueduct/actions/runs/3499900949/jobs/5861983185

kenxu95 (approved):
Can we just do a quick check over whether we're hard coding exact versions in any other GH actions?

Fanjia-Yan:
> Can we just do a quick check over whether we're hard coding exact versions in any other GH actions?

Checked. Other GH Actions all use 3.8 version.

PR #730, opened 2022-11-22
ENG-2032 Fix Workflows Back Navigation

agiron123:
When pressing the browser's back button on the workflow details page, the page would simply reload and not go back as expected.

ENG-2032

## Loom demo (if any)
https://www.loom.com/share/717dde5c32c1476c817a0f992c51ead6

### DIFF
--- a/src/ui/common/package-lock.json
+++ b/src/ui/common/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@aqueducthq/common",
-  "version": "0.1.3",
+  "version": "0.1.4",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@aqueducthq/common",
-      "version": "0.1.3",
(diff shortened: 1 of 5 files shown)


kenxu95 (approved):
Was this just missing a onClick?

agiron123:
> Was this just missing an onClick?

@kenxu95 couple things. Workflow card wasn't using the react router links to navigate to the workflow details page.

After fixing that, I still noticed that we had the navigation issue. Turns out that when the page is loaded and we navigate to a particular workflowDagId, calling navigate adds to the navigation stack. As a result, clicking back would just take you to the same page and add the same route back to the nav stack, creating an infinite loop :).

The real magic here is in the ```{ replace: true }```, which replaces the top of the nav stack instead of adding to it.

```
      navigate(
        `?workflowDagResultId=${encodeURI(workflow.selectedResult.id)}`,
        { replace: true }
      );
```

PR #731, opened 2022-11-22
Eng 1981 migrate workflow registration handler to use environments

likawind:
This PR adds changes to migration workflow registration handler to ensure enviroments associated with operators are also properly created. Specifically, we reuse `setupEnv` helper previously implemented for preview to create and register new conda env. And we additionally update DB operator with the new env's UUID so that they can be registered to DB.

We've already implemented the execution part to read env based on the `DBOperator`'s env ID. So the execution works automatically.

ENG-2011

## Loom demo (if any)
Screen Shot 2022-11-22 at 4 21 13 PM


cw75 (approved):
@likawind LGTM. The only nit is maybe we should move `setupExecEnv` to `lib/execution_environment/execution_environment.go` since it's shared across both handlers?

likawind:
@cw75 I just noticed the `setupExecEnv` helper takes a `dagSummary` object which depends on `request` package in `server`. So it makes better sense to me to defer moving it around in later refactor

PR #732, opened 2022-11-23
Builds many improvements into SDK integration test suite

kenxu95:
1) Introduces a `flow_name` fixture that allows the test suite to handle all the cleanup. The developer doesn't have to do any delete_flow() manipulation anymore. This fixture must be used as a function.
2) Replaces `run_flow_test()` with a better, updated version of the helper, `publish_flow_test()`. Also introduces a trigger counterpart `trigger_flow_test()`.
3) Adds the `name()` method to the Flow object.
4) Adds a `--keep-flows` flag to the command line, which will NOT delete flows that were created during the test. This helps preserve state when you're debugging an individual test, for example.
5) Adds an empty requirements file to `integration_tests/sdk`. This might help a little bit with perf on GH actions, but it should definitely help a lot when running against a backing engine like Kubernetes, for example. This requirements file is deleted in the periodic tests github action, to make sure that we're still covering inference.

Not supposed to be addressed by this PR, but was: ENG-2031

### DIFF
--- a/.github/workflows/periodic-integration-tests.yml
+++ b/.github/workflows/periodic-integration-tests.yml
@@ -19,6 +19,11 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      # For now, we want to always perform inference in our periodic tests for additional coverage.
+      - name: Remove the requirements.txt file.
+      - working-directory: integration_tests/sdk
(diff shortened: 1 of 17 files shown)


likawind (approved):
yay!

PR #740, opened 2022-11-30
Parallelize CreateLambdaFunction() via ErrGroup 

Fanjia-Yan:

- In this PR, I parallelize `CreateLambdaFunction()` using ErrGroup package, which creates a new go-routine for each  `CreateLambdaFunction()` using `sync.Waitgroup`.
- One thing worth noticing is that if running without `time.Sleep(x second)` ,connecting lambda through EC2 is fine but connecting lambda through Mac kept failing randomly at different spot. It seems like a concurrency issue and after I add the `time.Sleep`, it works fine. Might want to get more input on that.

## Performance Benchmark

- EC2: ~12min ->>> 3min 10 sec
- Mac: ~20min ->>> ~12min(11m47s, 13m26s,12m36s)

## Testing
I did multiple run on connecting to lambda through Mac and Ec2 and publish workflow for the lambda function.

### DIFF
--- a/src/golang/lib/engine/authenticate.go
+++ b/src/golang/lib/engine/authenticate.go
@@ -3,6 +3,8 @@ package engine
 import (
 	"context"
 
+	"golang.org/x/sync/errgroup"
+
 	"github.com/aqueducthq/aqueduct/lib/k8s"
 	lambda_utils "github.com/aqueducthq/aqueduct/lib/lambda"
 	"github.com/aqueducthq/aqueduct/lib/workflow/operator/connector/auth"
@@ -32,6 +34,8 @@ func AuthenticateLambdaConfig(ctx context.Context, authConf auth.Config) error {
 		return errors.Wrap(err, "Unable to parse configuration.")
 	}
 
+	errGroup := new(errgroup.Group)
+
 	functionsToShip := [10]lambda_utils.LambdaFunctionType{
 		lambda_utils.FunctionExecutor37Type,
 		lambda_utils.FunctionExecutor38Type,
@@ -46,10 +50,15 @@ func AuthenticateLambdaConfig(ctx context.Context, authConf auth.Config) error {
 	}
 
 	for _, functionType := range functionsToShip {
-		err := lambda_utils.CreateLambdaFunction(functionType, lambdaConf.RoleArn)
-		if err != nil {
-			return errors.Wrap(err, "Unable to Create Lambda Function")
-		}
+		lambdaFunctionType := functionType
+		errGroup.Go(func() error {
+			return lambda_utils.CreateLambdaFunction(lambdaFunctionType, lambdaConf.RoleArn)
+		})
 	}
+
+	if err := errGroup.Wait(); err != nil {
+		return errors.Wrap(err, "Unable to Create Lambda Function.")
+	}
+
 	return nil
 }
--- a/src/golang/lib/engine/authenticate.go
+++ b/src/golang/lib/engine/authenticate.go
@@ -2,6 +2,7 @@ package engine
 
 import (
 	"context"
+	"time"
 
 	"golang.org/x/sync/errgroup"
 
@@ -54,6 +55,7 @@ func AuthenticateLambdaConfig(ctx context.Context, authConf auth.Config) error {
 		errGroup.Go(func() error {
 			return lambda_utils.CreateLambdaFunction(lambdaFunctionType, lambdaConf.RoleArn)
 		})
+		time.Sleep(3 * time.Second)
 	}
 
 	if err := errGroup.Wait(); err != nil {
--- a/src/golang/go.mod
+++ b/src/golang/go.mod
@@ -23,7 +23,7 @@ require (
 	github.com/tklauser/go-sysconf v0.3.10 // indirect
 	github.com/yusufpapurcu/wmi v1.2.2 // indirect
 	golang.org/x/oauth2 v0.0.0-20220622183110-fd043fe589d2
-	golang.org/x/sync v0.0.0-20220601150217-0de741cfad7f // indirect
+	golang.org/x/sync v0.0.0-20220601150217-0de741cfad7f
 	google.golang.org/api v0.88.0
 	google.golang.org/grpc v1.48.0
 	gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c // indirect
--- a/src/golang/go.mod
+++ b/src/golang/go.mod
@@ -23,6 +23,7 @@ require (
 	github.com/tklauser/go-sysconf v0.3.10 // indirect
 	github.com/yusufpapurcu/wmi v1.2.2 // indirect
 	golang.org/x/oauth2 v0.0.0-20220622183110-fd043fe589d2
+	golang.org/x/sync v0.0.0-20220601150217-0de741cfad7f // indirect
 	google.golang.org/api v0.88.0
 	google.golang.org/grpc v1.48.0
 	gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c // indirect


src/golang/lib/engine/authenticate.go
```diff
@@ -32,6 +34,8 @@ func AuthenticateLambdaConfig(ctx context.Context, authConf auth.Config) error {
 		return errors.Wrap(err, "Unable to parse configuration.")
 	}
 
+	errGroup := new(errgroup.Group)
```
kenxu95, line 37: I think the canonical way to do it is to use `errgroup.WithContext(ctx)` or `WithCancel()` here. That way when the outside context gets cancelled the goroutines will too. In this case, I don't think our goroutines actually use the context so it might not be that useful. However, when used in combination with `errgroup.GoContext()` instead of `errgroup.Go()`, you can at least make sure that obvious errors in the first goroutine will fail all the goroutines immediately. This is especially useful if you're sleeping in the loop.

src/golang/lib/engine/authenticate.go
```diff
@@ -46,10 +50,16 @@ func AuthenticateLambdaConfig(ctx context.Context, authConf auth.Config) error {
...
 	for _, functionType := range functionsToShip {
-		err := lambda_utils.CreateLambdaFunction(functionType, lambdaConf.RoleArn)
-		if err != nil {
-			return errors.Wrap(err, "Unable to Create Lambda Function")
-		}
+		lambdaFunctionType := functionType
+		errGroup.Go(func() error {
+			return lambda_utils.CreateLambdaFunction(lambdaFunctionType, lambdaConf.RoleArn)
+		})
+		time.Sleep(3 * time.Second)
```
kenxu95, line 57: This is interesting. I have no problem with a small sleep but I wonder what the concurrency issue is. If you do this a bunch of times, does a pattern emerge (eg. failures when pulling from ECR)? I have a hard time believing its random.

kenxu95:
@Fanjia-Yan is this still being worked on?

Fanjia-Yan:
> @Fanjia-Yan is this still being worked on?

Yes, sorry for the wait. Just finished off some bug fix tasks yesterday and have resumed working on this today.

Fanjia-Yan:
Here is the update I have made after last discussion:

- Use `Errgroup.withcontext()` instead of `new Errgroup`
- Discard the 3 second timeout because it might work but is not sustainable
- Instead, we now create lambda function within three steps:
1. Authentication: authentication will only be performed once.
2. Pull Images: Images will be pulled on `MaxConcurrentDownload`(default to 3) basis(3 each time) to avoid pull timeout. 
3. Tag + Push Images + Create Lambda Functions: Images are pushed and made to be lambda function all at once using goruntine.

I have time each run and for each run I have deleted cached images using `docker system prune --all `, since authentication doesn't take much time, it is neglected. 

 EC2 as Server(Before):
Pulling : 230sec
Push: ~420sec
Total: 10-11mins

EC2 as Server(After):
Pull Finish: ~100sec
Push Time ~200sec
Total: 4-5 mins

Mac as Server(Before):
Pulling: 600 sec
Pushing: 900sec
Total: 24-25mins

Mac as Server(After):
Pulling: 720 sec
Pushing: 580sec
Total: 20-21 mins

It seems like my home's Internet bandwidth are causing the problems, which the pull time deviates a lot for several rounds of testing while EC2's results are consistent.

src/golang/lib/engine/authenticate.go
```diff
@@ -8,8 +8,12 @@ import (
 	"github.com/aqueducthq/aqueduct/lib/lib_utils"
 	"github.com/aqueducthq/aqueduct/lib/workflow/operator/connector/auth"
 	"github.com/dropbox/godropbox/errors"
+	"golang.org/x/sync/errgroup"
 )
 
+// The maximum numnber of concurrent download allowed by Docker and is default to 3.
```
kenxu95, line 14: nit: "and defaults to 3". Actually I think we can just get rid of the default comment since this is a constant.

src/golang/lib/engine/authenticate.go
```diff
@@ -42,11 +46,40 @@ func AuthenticateLambdaConfig(ctx context.Context, authConf auth.Config) error {
...
-	for _, functionType := range functionsToShip {
-		err := lambda_utils.CreateLambdaFunction(functionType, lambdaConf.RoleArn)
-		if err != nil {
-			return errors.Wrap(err, "Unable to Create Lambda Function")
+	errGroup, _ := errgroup.WithContext(ctx)
+
+	// Run authentication only once since the credentials will be memorized.
+	err = lambda_utils.AuthenticateDockerToECR()
+	if err != nil {
+		return errors.Wrap(err, "Unable to Create Lambda Function.")
```
kenxu95, line 54: Can we change err message to "Unable to authenticate Lambda Function"

src/golang/lib/engine/authenticate.go
```diff
@@ -42,11 +46,40 @@ func AuthenticateLambdaConfig(ctx context.Context, authConf auth.Config) error {
...
+	}
+
+	// Pull images on a currency of "MaxConcurrentDownload" to parallelize while avoiding pull timeout.
+	for i := 0; i < len(functionsToShip); i += MaxConcurrentDownload {
+		for j := 0; j < MaxConcurrentDownload; j++ {
```
kenxu95, line 59: This doubly-nested for-loop is a bit of a brittle construction, and is suboptimal since work is done in strict batches.

I think the canonical way to do this is to use Go channels. Put all the work to do in a single golang channel, to be consumed by some number of errgroup goroutines (workers). Each worker will keep going until 1) the input channel is empty 2) an error occurs or 3) the context is cancelled.

src/golang/lib/engine/authenticate.go
```diff
@@ -42,11 +46,40 @@ func AuthenticateLambdaConfig(ctx context.Context, authConf auth.Config) error {
...
+			return errors.Wrap(err, "Unable to Create Lambda Function.")
 		}
 	}
+
+	// Push the images and create lambda functions all at once.
+	for _, functionType := range functionsToShip {
+		lambdaFunctionType := functionType
+		errGroup.Go(func() error {
+			return lambda_utils.CreateLambdaFunction(lambdaFunctionType, lambdaConf.RoleArn)
+		})
```
kenxu95, line 77: I wonder if we should limit it to like 5 concurrent pushes at a time, so that this doesn't accidentally scale with the total number of function images. You'd do this in the same fashion as above, with workers.

src/golang/lib/engine/authenticate.go
```diff
@@ -42,11 +46,40 @@ func AuthenticateLambdaConfig(ctx context.Context, authConf auth.Config) error {
...
+		for j := 0; j < MaxConcurrentDownload; j++ {
+			if j+i < len(functionsToShip) {
+				lambdaFunctionType := functionsToShip[j+i]
+				errGroup.Go(func() error {
+					return lambda_utils.PullImageFromECR(lambdaFunctionType)
+				})
+			}
+		}
+		if err := errGroup.Wait(); err != nil {
+			return errors.Wrap(err, "Unable to Create Lambda Function.")
```
kenxu95, line 68: Can we make these error messages more specific? Lets note that the failure happened at the pulling stage at least.

src/golang/lib/engine/authenticate.go
```diff
@@ -42,11 +46,40 @@ func AuthenticateLambdaConfig(ctx context.Context, authConf auth.Config) error {
...
+	for _, functionType := range functionsToShip {
+		lambdaFunctionType := functionType
+		errGroup.Go(func() error {
+			return lambda_utils.CreateLambdaFunction(lambdaFunctionType, lambdaConf.RoleArn)
+		})
+	}
+
+	if err := errGroup.Wait(); err != nil {
+		return errors.Wrap(err, "Unable to Create Lambda Function.")
+	}
```
kenxu95, line 82: same error message comment here, maybe note that it happened at push time?

src/golang/lib/lambda/utils.go
```diff
@@ -23,15 +23,49 @@ type EcrAuth struct {
...
+	if err != nil {
+		log.Info(stdout.String())
+		log.Info(stderr.String())
+		return errors.Wrap(err, "Unable to authenticate docker client to ECR.")
+	}
+	return nil
+}
+
+func PullImageFromECR(functionType LambdaFunctionType) error {
+	// Pull the Image from public ECR Library.
```
kenxu95, line 67: Will the images be pulled onto the server machine? Should we worry about deleting the images locally after they are pushed? My impression was that they took up a ton of space. cc: @hsubbaraj-spiral

kenxu95 (changes requested):

agiron123:
@kenxu95 @hsubbaraj-spiral Is this PR still relevant? Seems like it's been open for a while.

Fanjia-Yan:
> @kenxu95 @hsubbaraj-spiral Is this PR still relevant? Seems like it's been open for a while.
@agiron123 Hey Andre, I have been distracted by the work authorization application process from the university so I haven't been working from then. I will continue working on this PR as soon as I got back!

Fanjia-Yan:
The update from last suggestions are the following:

1. Create a buffered channel and continuously add the lambda functions when the channel is not full until all the functions are created.
2. Set limit for ErrGroup with the amount of workers equal to`MaxConcurrentDownload = 3` or `MaxConcurrentUpload   = 5` and let them execute concurrently.
3. Error Handling through error group.

Performance Note: Implementing this method does not show any noticeable performance improve. I think the reason is that all lambda images are about the same size so batching using for loop vs. using channel didn't make much difference.

src/golang/lib/engine/authenticate.go
```diff
@@ -43,12 +49,46 @@ func AuthenticateLambdaConfig(ctx context.Context, authConf auth.Config) error {
...
-	for _, functionType := range functionsToShip {
-		err := lambda_utils.CreateLambdaFunction(functionType, lambdaConf.RoleArn)
-		if err != nil {
-			return errors.Wrap(err, "Unable to Create Lambda Function")
-		}
+	errGroup, _ := errgroup.WithContext(ctx)
+
+	pullImageChannel := make(chan lambda_utils.LambdaFunctionType, MaxConcurrentDownload)
+
+	pushImageChannel := make(chan lambda_utils.LambdaFunctionType, MaxConcurrentUpload)
```
kenxu95, line 56: I'm typically more of a fan of defining variables right before you need to use them, can we move these definitions down to where they are immediately needed? It keeps the code a little better module-ized.

src/golang/lib/engine/authenticate.go
```diff
@@ -43,12 +49,46 @@ func AuthenticateLambdaConfig(ctx context.Context, authConf auth.Config) error {
...
+	}
+
+	// Pull images on a concurrency of "MaxConcurrentDownload".
+	errGroup.SetLimit(MaxConcurrentDownload)
+	go lambda_utils.AddFunctionTypeToChannel(functionsToShip[:], pullImageChannel)
+	for lambdaFunction := range pullImageChannel {
+		lambdaFunctionType := lambdaFunction
+		errGroup.Go(func() error {
+			return lambda_utils.PullImageFromECR(lambdaFunctionType)
+		})
```
kenxu95, line 71: hmm this still doesn't seem quite like the right concurrency model. Let me know what you think about this:
- Typically in a situation where you have a maximum number of workers situation, you should spin up those workers explicitly, and have them constantly poll for new work once their old work is finished. Intuitively, this means that we should immediately spin up `MaxConcurrentDownload` workers with `errGroup.Go()`. Inside, each goroutine, you should fetch a function type from the FunctionType channel to start downloading. Then, that goroutine writes that processed FunctionType to a different, upload work channel, and goes back to polling on the input FunctionType channel until all work is completed.

- In the upload case, you would do the same thing, but poll on the output FunctionType channel from the above step. This allows a much more efficient pipeline where you can immediately start uploading pulled images to ECR, even while other pulls are taking place. Essentially, *all* the waiting should happen at the very end.

In order for all this to work, 1) `AddFunctionTypeToChannel` will have to happen synchronously in the very beginning. 2) You'll need one errgroup for uploads and one for downloads. 

Happy to chat about this! I agree it's kinda complicated, not super easy to test and iterate on, and might not be the highest priority thing over the other lambda tasks. Maybe we can try pairing on this sometime next week?

kenxu95 (changes requested):
About your "Performance Note", I remember you said that this was faster on EC2? (when there's more bandwidth).

src/golang/lib/engine/authenticate.go
Fanjia-Yan: Yeah works for me! I will experiment with the currency model you mention a bit before that.

kenxu95: Cool feel free to drop something on my calendar :)

Fanjia-Yan:
Ended up with one errgroup for both pulling and pushing! Performance is about the same on my PC and EC2. I saw some more `select` clause example. Some code segment have `case <-ctx.Done():
  return ctx.Err()`, the `ctx` is from the `createErrGroup` with context. I am wondering if we should add that in as well since we want to error out when context is closed.

kenxu95:
@Fanjia-Yan Yeah I think we should definitely add that! I thought errgroups would do that automatically but I guess all they can do is cancel the context, so we'll need to select on it. Based on quick reading of https://www.fullstory.com/blog/why-errgroup-withcontext-in-golang-server-handlers/

src/golang/lib/engine/authenticate.go
```diff
@@ -43,11 +49,69 @@ func AuthenticateLambdaConfig(ctx context.Context, authConf auth.Config) error {
...
+					lambdaFunctionType := functionType
+					err := lambda_utils.PullImageFromECR(lambdaFunctionType)
+					if err != nil {
+						return err
+					}
+					pushImageChannel <- functionType
+				case <-errGroupCtx.Done():
+					return errGroupCtx.Err()
+				default:
+					return nil
```
kenxu95, line 79: Can you add a comment here that this case should only be hit when the `pullImageChannel` is empty?

src/golang/lib/engine/authenticate.go
```diff
@@ -43,11 +49,69 @@ func AuthenticateLambdaConfig(ctx context.Context, authConf auth.Config) error {
...
+				}
+			}
+		})
+	}
+
+	// Create a signal channel to flag when all the lambda functions have been created.
+	signalChannel := make(chan lambda_utils.LambdaFunctionType, len(functionsToShip))
```
kenxu95, line 86: nit: Maybe a name like `outstandingWorkChannel` or `incompleteWorkChannel` might be a little more explanatory? I'd like to capture the fact that the channel is populated, but gradually drawn down to zero. Signal channel makes it sound like we're adding to the channel as the signaling mechanism.

kenxu95: Also, I think it's important to note here in a comment that the contents of this channel do not matter, only the number of elements in it :)

src/golang/lib/engine/authenticate.go
```diff
@@ -43,11 +49,69 @@ func AuthenticateLambdaConfig(ctx context.Context, authConf auth.Config) error {
...
-		err := lambda_utils.CreateLambdaFunction(functionType, lambdaConf.RoleArn)
-		if err != nil {
-			return errors.Wrap(err, "Unable to Create Lambda Function")
-		}
+	// Run authentication only once since the credentials will be recorded.
+	err = lambda_utils.AuthenticateDockerToECR()
+	if err != nil {
+		return errors.Wrap(err, "Unable to authenticate Lambda Function.")
+	}
+
```
kenxu95, line 57: Can we move the new pushing/pulling code to `lambda/utils.go`? This file is shared by a lot of other integrations.

kenxu95 (approved):
Thanks for slogging through everything! Left some mostly cosmetic comments but the functionality looks good.

I think I'm ok with us merging this, as long as you've manually tested the very latest version (after the cosmetic changes I've recommended)

src/golang/lib/engine/authenticate.go
Fanjia-Yan: Sounds good! Thanks for pushing this through. I will do another around of testing after that.

PR #753, opened 2022-12-02
Fix bug where metrics graph and history is not sorted by time

Fanjia-Yan:
This PR makes the metrics and check history sorted by time of creation on UI side. The graph displays metric left to right from oldest to newest. The history displays metric top to buttom from newest to oldest. 
Metric:
Screen Shot 2022-12-01 at 8 27 49 PM
Check:
Screen Shot 2022-12-01 at 8 27 55 PM

https://linear.app/aqueducthq/issue/ENG-1933/fix-bug-where-metrics-graph-and-history-is-not-sorted-by-time

### DIFF
--- a/src/ui/common/src/components/workflows/artifact/metric/history.tsx
+++ b/src/ui/common/src/components/workflows/artifact/metric/history.tsx
@@ -12,7 +12,7 @@ import Plot from 'react-plotly.js';
 
 import { ArtifactResultsWithLoadingStatus } from '../../../../reducers/artifactResults';
 import { theme } from '../../../../styles/theme/theme';
-import { Data, DataSchema } from '../../../../utils/data';
+import { Data, DataSchema, TableRow } from '../../../../utils/data';
 import ExecutionStatus, {
   isFailed,
   isInitial,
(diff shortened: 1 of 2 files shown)


src/ui/common/src/components/workflows/artifact/metric/history.tsx
```diff
@@ -62,10 +62,18 @@ const MetricsHistory: React.FC<Props> = ({ historyWithLoadingStatus }) => {
       }
     ),
   };
-
-  const dataToPlot = historicalData.data.filter(
-    (x) => !!x['timestamp'] && !!x['value']
+  const dataSortedByLatest = historicalData.data.sort(
+    (x, y) =>
+      Date.parse(y['timestamp'] as string) -
```
agiron123, line 67: What do y['timestamp'] and x['timestamp'] look like? Ideally it's just time in milliseconds since the unix epoch right? If so, we don't need to parse these dates out.

agiron123: I believe you could do something like x.getTime() and y.getTime() to get a number back that you can use for comparison:
image

Fanjia-Yan: Screen Shot 2022-12-06 at 11 24 07 AM
@agiron123 The timestamp here is the Datetime object converted to string I believe, which is why I converted back to Date. One way we can do it is simply store the timestamp as `Date` object. Is there any reason we cast it to a string over here?

vsreekanti:
@Fanjia-Yan @agiron123, what's the status of this PR? This is a pretty embarrassing bug, so let's please get this in before Monday's release.

src/ui/common/src/components/workflows/artifact/metric/history.tsx
agiron123: we convert to a string so that it shows nicely like 1/1/2023 4:00pm

agiron123 (approved):
Code here should be fine. Was just curious on how the sorting worked with parsed dates.

PR #760, opened 2022-12-02
Eng 2025 restrict delete integration to check for only active workflows

likawind:
This PR modifies the check in `delete integration` route against only active workflows. Previously, we block integration deletion even if there are historical runs on this integraiton.

ENG-2025

### DIFF
--- a/src/golang/cmd/server/handler/delete_integration.go
+++ b/src/golang/cmd/server/handler/delete_integration.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"net/http"
 
+	"github.com/aqueducthq/aqueduct/cmd/server/queries"
 	"github.com/aqueducthq/aqueduct/cmd/server/routes"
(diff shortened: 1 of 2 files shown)


cw75 (commented):
@likawind Should we put `validateNoActiveWorkflowOnIntegration` in the prepare phase, just as what we have currently?

cw75:
@likawind alright I merged all of my PRs to this branch. Once the comment above is addressed I think we can run the tests and merge to main?

likawind:
@cw75  Thanks! I actually feel `Perform` is the right place to put application logic. Ideally I would like `Prepare` to be just prepare, where we parse http requests. `Perform` should hold all application logics esp. those involves DB access. We are not following this in all handlers due to various reasons, so I think currently `Prepare` and `Perform` don't make much difference though.

cw75 (approved):

PR #761, opened 2022-12-04
Adds GC column to the env table

cw75:
This PR adds a GC column to the execution environment table, and modifies existing logic to update this column correctly.
The PR also adds an async GC routine after the user deletes an workflow, as after that there could be exec envs that are no longer used.


likawind (approved):

likawind (commented):
For the sake of readability, should we explicitly set GC column to false when we create the env?

cw75:
@likawind Since the column already has a property that defaults the entry to false, I think we are good here. I also verified that new columns are indeed populated with "false".

PR #768, opened 2022-12-06
Fixes bug where multiple save operators to the same integration crashes the UI

kenxu95:
This fails because of mismatches expectations between the SDK and backend. The SDK thinks that having multiple save operators with the same name is fine, as long as they are in different branches. However, this crashes when the UI attempts to display these kinds of DAGs.

The solution for now is to deduplicate save operators the same way we duplicate extract operators (with a suffix starting at "1" and gradually incrementing). However, save operators on the same artifact to the same integration are not duplicated. In those cases, we will simply overwrite the existing operator (this allows for updating of saves).

Of course, this means that you cannot currently save the same artifact to different places into the same integration.

Example workflow:
```
db = client.integration("aqueduct_demo")
output1 = db.sql("Select * from hotel_reviews;")
output2 = db.sql("Select * from hotel_reviews;")

output1.save(db.config("table_1", "replace"))
output2.save(db.config("table_2", "replace"))
```

### DIFF
--- a/sdk/aqueduct/artifacts/base_artifact.py
+++ b/sdk/aqueduct/artifacts/base_artifact.py
@@ -4,18 +4,21 @@ from abc import ABC, abstractmethod
 from typing import Any, Dict, Optional
 
 from aqueduct.dag import DAG
-from aqueduct.dag_deltas import (
-    AddOrReplaceOperatorDelta,
-    apply_deltas_to_dag,
-    find_duplicate_load_operator,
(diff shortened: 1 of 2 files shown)


kenxu95:
https://capture.dropbox.com/RTEudNQBmMJmWHBb

agiron123 (approved):

cw75 (approved):
LGTM

PR #769, opened 2022-12-06
[SDK] Update our Save API to be integration.save()

kenxu95:
## Describe your changes and why you are making these changes
Instead of the hard-to-read `output_artifact.save()`, we should use `integration.save(output_artifact, ..)` instead.

This PR also does the following:
- Deprecates the old path by logging a warning each time. but does not eliminate the path. All the places we should remove are tagged with TODO(ENG-2035).
- Updates a lot of the docstrings to give the user context for what their saves mean.

The testing portion is large and is in a follow-up PR.

cc: @vsreekanti since this is a user-facing.

## Related issue number (if any)
ENG-1947

## Checklist before requesting a review
- [ ] I have created a descriptive PR title. The PR title should complete the sentence "This PR...".
- [ ] I have performed a self-review of my code.
- [ ] I have included a small demo of the changes. For the UI, this would be a screenshot or a Loom video.
- [ ] If this is a new feature, I have added unit tests and integration tests.
- [ ] I have run the integration tests locally and they are passing.
- [ ] I have run the linter script locally (See `python3 scripts/run_linters.py -h` for usage).
- [ ] All features on the UI continue to work correctly.
- [ ] Added one of the following CI labels:
    - `run_integration_test`: Runs integration tests
    - `skip_integration_test`: Skips integration tests (Should be used when changes are ONLY documentation/UI)

PR #770, opened 2022-12-06
Update the integration tests to support the new Save API

kenxu95:
This is the testing follow-up of: https://github.com/aqueducthq/aqueduct/pull/769

Because of the Save API change, we should now abstract away our `save()` helper. This PR does the following:
- Adds a `save()` helper that should be used by the test cases (instead of directly using the SDK's Save API).
- Add a flag `--deprecated`, which allows us to continue to provide coverage to code paths that are deprecated, but need to be temporary support. See `save()` for how this switch works.
- Adds a `Validator` class instance (see `validator.py`), which can be included in tests that need it. This Validator can be used to check certain things about the flow after `publish_flow_test()` is called. The reason I made this a class is 1) it's easier to include as a fixture and 2) its easier to organize all our validators and 3) in the future, we can improve performance by caching common data that multiple validations need to use.

ENG-1947

### DIFF
--- a/integration_tests/sdk/delete_workflow_test.py
+++ b/integration_tests/sdk/delete_workflow_test.py
@@ -1,5 +1,5 @@
-import pytest
 import pandas as pd
+import pytest
 from aqueduct.error import InvalidRequestError
 from constants import SHORT_SENTIMENT_SQL_QUERY
 from utils import (
@@ -84,7 +84,9 @@ def test_delete_workflow_saved_objects(client, flow_name, data_integration, engi
     check_table_doesnt_exist(integration, table_name)
 
 
-def test_delete_workflow_saved_objects_twice(client, flow_name, data_integration, engine, validator):
+def test_delete_workflow_saved_objects_twice(
+    client, flow_name, data_integration, engine, validator
+):
     """Checking the successful deletion case and unsuccessful deletion case works as expected.
     To test this, I have two workflows that write to the same table. When I delete the table in the first workflow,
     it is successful but when I delete it in the second workflow, it is unsuccessful because the table has already
--- a/integration_tests/sdk/delete_workflow_test.py
+++ b/integration_tests/sdk/delete_workflow_test.py
@@ -1,4 +1,5 @@
 import pytest
+import pandas as pd
 from aqueduct.error import InvalidRequestError
 from constants import SHORT_SENTIMENT_SQL_QUERY
 from utils import (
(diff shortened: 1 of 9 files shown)


integration_tests/sdk/delete_workflow_test.py
```diff
@@ -39,12 +40,12 @@ def test_delete_workflow_invalid_saved_objects(client, flow_name, data_integrati
     client.flow(flow.id())
 
 
-def test_delete_workflow_saved_objects(client, flow_name, data_integration, engine):
+def test_delete_workflow_saved_objects(client, flow_name, data_integration, engine, validator):
```
likawind, line 43: I would suggest to put these as a single `context` class so that it's easier to add / change / rename all 'contexts' available to a test, without changing signatures of tests every single time.

likawind (approved):
Looks great! I dropped one comment to make test signatures easier to manage

integration_tests/sdk/delete_workflow_test.py
kenxu95: interesting - but that means every test is going to initialize all these fixtures, even if the test doesn't need them right?

kenxu95: My understanding of the philosophy behind pytest fixtures is that they are run only when requested by the test case. Test cases that don't need to, for example, publish a flow, wouldn't need to ask for `flow_name`. And I think this improves the readability of what a test is capable of, based on just glancing at the signature, since not all test cases require all the fixtures.

I'd be willing to discuss this with a broader audience perhaps? I'm curious what others think.

hsubbaraj-spiral: I agree with Kenny here, tests should only request fixtures that are required for the test to run.

hsubbaraj-spiral (approved):
Looks great to me, just a question about the `Validator` class naming.

integration_tests/sdk/validator.py
```diff
@@ -0,0 +1,78 @@
...
+from typing import Any, List, Tuple
+
+from aqueduct.integrations.integration import Integration
+from aqueduct.integrations.sql_integration import RelationalDBIntegration
+from utils import artifact_id_to_saved_identifier
+
+from aqueduct import Client, Flow, LoadUpdateMode
+
+
+class Validator:
```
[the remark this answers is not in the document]
kenxu95: I was imagining that this class would validate everything about a completed run. `publish_flow_test()` makes sure that the run terminates with some expected status. Then, we'd use the validator to verify any additional things about the run, like saved objects or the shape of the published flow's DAG.

PR #779, opened 2022-12-07
[15/N] Replaces custom queries with repos

saurav-c:
This PR takes all of the adhoc "custom" queries and turns them into the relevant `Views` and `Repos`. The motivation was to make these queries as clean as possible in terms of fitting into a Repo that made logical sense, rather than having a collection of adhoc queries. 

TODO: Writing the database integration tests.

ENG 1903

### DIFF
--- a/src/golang/lib/models/views/id.go
+++ b/src/golang/lib/models/views/id.go
@@ -1,0 +1,8 @@
+package views
+
+import "github.com/google/uuid"
+
+// ObjectID is a wrapper around any ID
+type ObjectID struct {
+	ID uuid.UUID `db:"id" json:"id"`
+}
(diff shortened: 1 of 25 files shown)


likawind (commented):
Questions about 'batch' naming principles. It seems that we should use `GetXByYAndZBatch` if `Y` and `Z` are lists. What if Y is a list and Z is a single ID?

src/golang/lib/models/views/id.go
```diff
@@ -0,0 +1,8 @@
+package views
+
+import "github.com/google/uuid"
+
+// ObjectID is a wrapper around any ID
+type ObjectID struct {
+	ID uuid.UUID `db:"id" json:"id"`
```
likawind, line 7: This can go to future refactors. But I think in these situations we should simply return id / list of ids from the query. If a struct is required to fetch the return value it should be an private view within that repo, rather than a public one

likawind (approved):

PR #781, opened 2022-12-08
Refactors the SDK code into layers

kenxu95:
I apologize - this PR was an absolute monster.. I recommend checking out this branch and poking around the SDK as a better way of understanding what's going on here.

Our SDK has gotten to the point where its flat file structure is causing more and more unexpected issues (mainly circular dependencies) when developing on it. It's in desperate need of some refactor, the PR enforces the following hierarchy:

- User-facing files (decorators, Flow, FlowRun, Client, etc.), these will continue to stay at the top-level.
- `integrations/` (already exists)
- `artifacts/` (already exists)
- `globals/`: Global variables
- `backend/`: The API Client
- `utils/`: serialization, dag_deltas, etc.
- `models/`: most of our pydantic structs, eg `dag`, `operator_spec`, etc.
- `constants/`: our enums, eg. `ArtifactType`
- logger, errors (these are also staying at in the top-level directory)

This hierarchy should make it clear: python files in each directory should import from the layers below, but not from above! 
That way, you should only need to worry about circular dependencies without your own submodule.

Adding a couple guidelines to the Python SDK Style Guide, which this PR attempts to follow: https://www.notion.so/aqueducthq/Python-Style-Guide-ccadeb6e47da4009bf446737e7a9872f


likawind:
Thanks for getting this started! This is a huge improvement of life!

For package named 'utils', I feel they are too vague to understand what's inside and could easily expand to unmaintainable size that requires reorg. So I think we should at least rethink the namings of newly added `utils` file. I'm fine to merge with current state. We can keep the discussion going and add these as separate PR. More thoughts on files I read:
* sdk/aqueduct/backend/utils -> response or request_response
* sdk/aqueduct/integration/utils -> this doesn't seem to relate to integration. Should it go to `dag` and we can name it like `graph_traverse`?
* sdk/aqueduct/utils/utils -> we should take some time to split this file into finer grained pieces

likawind:
I do find some external references also suggesting avoiding  'utils', for example: https://breadcrumbscollector.tech/stop-naming-your-python-modules-utils/

kenxu95:
Yep, I totally agree we should avoid utils.py (had originally added it as a bullet to https://www.notion.so/aqueducthq/Python-Style-Guide-ccadeb6e47da4009bf446737e7a9872f). I actually pulled out a ton of things from utils.py into better-named files, but had some trouble thinking of how to split up the remaining pieces of `aqueduct/utils/utils.py` - any ideas for this one?

likawind:
I need to take deeper look to the codebase, but from a glance I feel there are a few ways:
* if a helper is only used in a package, we can move it to that package rather than keeping it in `utils`
* we can add some arbitrary groupings and turn each group into one single file, like `uuid`, `parameter`, `engine_configs`, `pretty_string`, etc. I think it's acceptable that some grouping contains just one helper and some may need to reorg in the future

kenxu95:
Yep makes sense. I've already gone through all the stuff in utils/utils.py in this PR and pulled them out if they were only imported within packages. I can also do the second bullet, but this PR was already extremely large so I kept a small segment of auxilliary methods in utils.py as a compromise. I wonder if I can make that a follow-up task and we can merge this improvement as is?

likawind (approved):
Yes, those utils thing should def. go to another PR

sdk/aqueduct/integrations/utils.py
```diff
@@ -0,0 +1,100 @@
...
+        if colliding_op is None:
+            op_name = candidate_op_name  # break out of the loop!
+        default_op_index += 1
+
+    assert op_name is not None
+
+    return op_name
+
+
+def _get_checks_for_op(op: Operator, dag: DAG) -> List[Operator]:
```
hsubbaraj-spiral, line 44: Do these utils functions belong in this package? I think they would be better suited in the `aqueduct.artifacts` package.

hsubbaraj-spiral (approved):
Apart from moving a few of the utils functions, this looks great to me!

saurav-c (approved):
This looks great! Doesn't need to happen in this PR or anytime soon, but do we have a sense of when we should import an entire module vs a specific method of object from a module?

PR #787, opened 2022-12-09
Include Python type into artifact versions route's response

cw75:
This PR first adds the python type to the artifact result's `metadata` field, and then return the entire `metadata` field in the artifact versions route.

From the UI, for each `artifactVersion`, we can get the python type by doing `artifact_version.metadata.python_type`.

I still need to make a PR to backfill the python type for existing artifact results in order to pass the regression test.


src/golang/lib/repos/sqlite/operator_result.go
```diff
@@ -100,7 +100,6 @@ func (*operatorResultReader) GetCheckStatusByArtifactBatch(
 		`SELECT DISTINCT
 			workflow_dag_edge.from_id AS artifact_id,
 			operator.name AS operator_name,
-			operator_result.status,
```
cw75, line 103: @saurav-c I had to remove this, otherwise we will get a database scan error.

src/golang/lib/repos/sqlite/artifact.go
```diff
@@ -87,7 +87,7 @@ func (*artifactReader) GetIDsByDAGAndDownstreamOPBatch(
 	// Get all the unique `artifact_id`s with an outgoing edge to an operator specified by `operatorIds`
 	// from workflow DAGs specified by `workflowDagIds`.
 	query := fmt.Sprintf(
-		`SELECT DISTINCT from_id AS artifact_id 
+		`SELECT DISTINCT from_id AS id
```
cw75, line 90: @saurav-c I had to change the column name to match the struct, otherwise we will get a database scan error.

saurav-c: Got it thanks!

src/golang/cmd/server/handler/get_artifact_result.go
```diff
@@ -55,9 +55,10 @@ type artifactResultMetadata struct {
 	// `status` in the UI.
 	Status            shared.ExecutionStatus            `json:"status"`
 	ExecState         shared.ExecutionState             `json:"exec_state"`
-	Schema            []map[string]string               `json:"schema"`
-	SerializationType artifact_result.SerializationType `json:"serialization_type"`
-	ArtifactType      artifact.Type                     `json:"artifact_type"`
+	Schema            []map[string]string               `json:"schema,omitempty"`
```
agiron123, line 58: Quick question about this omitempty keyword. Does it mean that if schema or any of these keys are null that the value will not come back in the response?

Might be easier for me on the frontend to have the keys always available in the response and just get a null value in the payload.

likawind (approved):

PR #801, opened 2022-12-13
Add ability to set engine in the operator decorator

kenxu95:
@hsubbaraj-spiral as lead reviewer. @saurav-c for data model perspective.

Unlocks the ability to run a single operator on a custom compute integration like:
```
@op(engine="my_k8s")
def foo():
    ...
```
Adds the following hierarchy in the SDK:
1) @op(engine=...)
2) client.publish_flow(.., engine=...)
3) aq.global_config({"engine": "..."})

In the database, engine config can be set on both the DAG and the operator spec. The operator spec has higher precedence if customized.

Tested manually with a workflow where one of the operators used k8s. Passes integration tests too ofc.

Question for @saurav-c: I don't think a migration is necessary here right, since all I'm doing is adding a pointer field `engine_config` to OperatorSpec. For previous workflows, this will resolve to nil (meaning aqueduct engine) anyways right?

## Follow-Ups
[] Add this field to the UI data types. It's unused for now, but its important to document.

ENG-2039

### DIFF
--- a/src/golang/lib/models/dag.go
+++ b/src/golang/lib/models/dag.go
@@ -27,7 +27,7 @@ type DAG struct {
 	CreatedAt     time.Time            `db:"created_at" json:"created_at"`
 	StorageConfig shared.StorageConfig `db:"storage_config" json:"storage_config"`
 
-	// Sets the default engine for DAG execution. Can we overridden by the operator spec.
+	// Sets the default engine for DAG execution. Can be overridden by the operator spec.
 	EngineConfig shared.EngineConfig `db:"engine_config" json:"engine_config"`
 
 	/* Field not stored in DB */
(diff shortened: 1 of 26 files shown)


src/golang/lib/models/dag.go
```diff
@@ -26,7 +26,9 @@ type DAG struct {
 	WorkflowID    uuid.UUID            `db:"workflow_id" json:"workflow_id"`
 	CreatedAt     time.Time            `db:"created_at" json:"created_at"`
 	StorageConfig shared.StorageConfig `db:"storage_config" json:"storage_config"`
-	EngineConfig  shared.EngineConfig  `db:"engine_config" json:"engine_config"`
+
+	// Sets the default engine for DAG execution. Can we overridden by the operator spec.
+	EngineConfig shared.EngineConfig `db:"engine_config" json:"engine_config"`
```
saurav-c, line 31: nit: "Can be overridden..."

saurav-c (approved):
Looks good from data model changes perspective, yes the `EngineConfig` should be nil for existing Operator specs.

hsubbaraj-spiral (approved):
Yep, looks great, thanks @kenxu95 !

hsubbaraj-spiral:
Might be some repo org issues that are causing the tests to fail, but looks good to merge after those are fixed!

PR #803, opened 2022-12-13
[UI] Add engine_config field to operator spec

kenxu95:
This isn't used anywhere in the UI, but just adding it as important context once we start showing the engine type on the UI :) These fields are populated correctly once this PR merges.

ENG-2039

### DIFF
--- a/src/ui/common/src/utils/engine.ts
+++ b/src/ui/common/src/utils/engine.ts
@@ -7,9 +7,23 @@ export enum EngineType {
 
 export type EngineConfig = {
   type: EngineType;
+  aqueduct_config?: AqueductConfig;
   airflow_config?: AirflowConfig;
+  k8s_config?: K8sConfig;
+  lambda_config?: LambdaConfig;
 };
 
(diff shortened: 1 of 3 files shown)


agiron123:
Quick question here, changes seem fine on my end. However can we make the config more generic (can be done in another PR):

```
export type EngineConfig = {
  type: EngineType;
  airflow_config?: AirflowConfig;
};
```

airflow_config being inside the EngineConfig object is a little confusing to me. Perhaps we can make a more generic "config" key here and make it a union type of AirflowConfig | LambdaConfig, etc.

agiron123 (approved):

kenxu95:
Yeah we're currently just returning back the same objects we use on the backend, where EngineConfig has one field per config type (which isn't really a union type). I'm happy to add the additional config types to the struct too, but I think making this more streamlined/readable is a read endpoint task that we should probably have a broader discussion on?

agiron123:
> 

@kenxu95 yes, let's get some time to talk about building a read endpoint.

PR #810, opened 2022-12-15
Refactor job manager interface to always return a JobError

kenxu95:
Finishes what I started a few weeks ago when dealing with job manager errors coming from K8s and Lambda that were user-facing.

Additionally, we want to move away from the 
```
ErrPollJobTimeout = errors.New("Reached timeout waiting for the job to finish.")
```
pattern, since the stack trace that error captures is completely useless for our debugging (since stack trace capture happens at error creation time). By wrapping these errors on the fly in a JobError, we make it possible to both 1) record useful stack trace information when returning errors from the job manager and 2) perform any conditional checks on the specific type of error with `JobError.Code()`.

ENG-2009

### DIFF
--- a/src/golang/lib/job/errors.go
+++ b/src/golang/lib/job/errors.go
@@ -1,0 +1,75 @@
+package job
+
+import "github.com/dropbox/godropbox/errors"
+
+// JobErrorCode come from our JobManagers when they fail to properly guide
+// their a job through its proper lifecycle. Errors surfaced this way are propagated
+// outside of the python executor context. Their meaning is consistent across all
+// types of JobManagers.
+type JobErrorCode int
+
+const (
+	// System indicates an unexpected system issue that we cannot recover from.
+	System JobErrorCode = 0
+
+	// User error code indicates that the issue was the user's fault, and to surface that message
+	// to the user.
+	User = 1
+
+	// JobMissing indicates that the job manager could not find the specified job.
+	JobMissing = 2
+
+	// Noop indicates that the job manager does not have the context to make a definitive claim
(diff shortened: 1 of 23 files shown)


src/golang/lib/job/errors.go
```diff
@@ -0,0 +1,74 @@
...
+	// Noop indicates that the job manager does not have the context to make a definitive claim
+	// as to whether the job has succeeded or not. This is equivalent to saying "Do not ask me, I don't know.".
+	//
+	// For example, polling a Lambda JobManager will return this error because there is no concept of
+	// fetching a specific job's information from Lambda. The caller must figure out the job status
+	// through other means.
+	Noop = 2
+)
+
+type JobError interface {
```
saurav-c, line 31: When would we have a different implementation of this interface?

src/golang/lib/job/process.go
```diff
@@ -291,18 +291,18 @@ func (j *ProcessJobManager) Config() Config {
 }
 
 func (j *ProcessJobManager) Launch(
-	ctx context.Context,
+	_ context.Context,
```
saurav-c, line 294: nit: should we remove this param if its unused

saurav-c (approved):
LGTM left a couple comments

src/golang/lib/job/errors.go
kenxu95: It's mainly because it needs to implement the errors interface, and so that we can `return nil` as a job error.

src/golang/lib/job/process.go
kenxu95: I think it's used by the other job manager implementations though.

PR #817, opened 2022-12-16
[3/3] Adds DatabricksJobManager

hsubbaraj-spiral:
Adds DatabricksJobManager and related libraries
eng-2086


src/golang/lib/collections/integration/config.go
```diff
@@ -45,6 +45,12 @@ type LambdaIntegrationConfig struct {
 	RoleArn string `json:"role_arn" yaml:"roleArn"`
 }
 
+type DatabricksIntegrationConfig struct {
+	WorkspaceUrl         string `json:"workspace_url" yaml:"workspaceUrl"`
+	AccessToken          string `json:"access_token" yaml:"accessToken"`
+	S3InstanceProfileArn string `json:"s3_instance_profile_arn" yaml:"s3InstanceProfileArn"`
```
kenxu95, line 51: I'd like us to get in the habit of commenting the fields of configuration structs. Can we add a few sentences for each? Can also copy the descriptions over to the job manager config.

src/golang/lib/databricks/jobs.go
```diff
@@ -0,0 +1,100 @@
...
+						InstanceProfileArn: s3InstanceProfileArn,
+					},
+				},
+				SparkPythonTask: &jobs.SparkPythonTask{
+					PythonFile: pythonFilePath,
+				},
+				Libraries: []libraries.Library{
+					libraries.Library{
+						Pypi: &libraries.PythonPyPiLibrary{
+							Package: "aqueduct-ml==0.1.7",
```
kenxu95, line 69: Should we be hard-coding the version? It's another version number we have to remember to bump - can we do without it? Or use the server's version number?

src/golang/lib/databricks/jobs.go
```diff
@@ -0,0 +1,100 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "Unable to create Databricks client.")
+	}
+	return datatbricksClient, nil
+}
+
+func ListJobs(
+	ctx context.Context,
+	databricksClient *databricks.WorkspaceClient,
+) error {
```
kenxu95, line 31: nit: any reason we don't return the job list here? It's a little strange for a method called `ListJobs` to not return any job info. The caller can just ignore it if they want. Or, if its only used for validating the integration, we can name it as such.

src/golang/lib/job/databricks.go
```diff
@@ -0,0 +1,189 @@
...
+			return "", "", errors.Wrap(err, "Unable to encode spec.")
+		}
+		return databricksParamScript, specStr, nil
+
+	} else if spec.Type() == AuthenticateJobType ||
+		spec.Type() == LoadJobType ||
+		spec.Type() == ExtractJobType ||
+		spec.Type() == LoadTableJobType ||
+		spec.Type() == DeleteSavedObjectsJobType ||
+		spec.Type() == DiscoverJobType {
```
kenxu95, line 77: Are these expected to actually work? The issue is just data scale right.

src/golang/lib/job/databricks.go
```diff
@@ -0,0 +1,189 @@
...
+		return errors.Wrap(err, "Spec unexpectedly has no storage config.")
+	}
+	bucket := storageConfig.S3Config.Bucket
+	pythonFilePath := fmt.Sprintf("%s/%s", bucket, scriptFile)
+
+	jobId, err := databricks_lib.CreateJob(ctx, j.databricksClient, name, j.conf.S3InstanceProfileArn, pythonFilePath)
+	if err != nil {
+		return errors.Wrap(err, "Error creating job in Databricks.")
+	}
+	log.Info("SPEC:")
```
kenxu95, line 122: nit: "DATABRICKS SPEC: " might be more informative in the logs.

saurav-c: nit: We can combine the 2 log statements into: `log.Info("Databricks Spec: \n %v", specStr)`

src/golang/lib/job/databricks.go
```diff
@@ -0,0 +1,189 @@
+package job
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/aqueducthq/aqueduct/lib/collections/shared"
+	databricks_lib "github.com/aqueducthq/aqueduct/lib/databricks"
```
saurav-c, line 8: nit: Is there a reason for this alias if the sdk package already has an alias, which im assuming is to prevent package import clashes.

src/golang/lib/job/databricks.go
```diff
@@ -0,0 +1,189 @@
...
+		return err
+	}
+	storageConfig, err := spec.GetStorageConfig()
+	if err != nil {
+		return errors.Wrap(err, "Spec unexpectedly has no storage config.")
+	}
+	bucket := storageConfig.S3Config.Bucket
+	pythonFilePath := fmt.Sprintf("%s/%s", bucket, scriptFile)
+
+	jobId, err := databricks_lib.CreateJob(ctx, j.databricksClient, name, j.conf.S3InstanceProfileArn, pythonFilePath)
```
saurav-c, line 118: nit: rename to `jobID` to follow convention in https://www.notion.so/aqueducthq/Golang-Style-Guide-665fb1b16d3f4448afc385fbae903a98#d6234eef37e14e67a753cf23b4b22cf2

src/golang/lib/job/databricks.go
```diff
@@ -0,0 +1,189 @@
...
+	if err != nil {
+		return errors.Wrap(err, "Error creating job in Databricks.")
+	}
+	log.Info("SPEC:")
+	log.Info(specStr)
+	runId, err := databricks_lib.RunNow(ctx, j.databricksClient, jobId, specStr)
```
saurav-c, line 124: nit: rename to `runID` (and places below)

src/golang/lib/job/databricks.go
```diff
@@ -0,0 +1,189 @@
...
+
+	status := getRunResp.State.LifeCycleState
+	if status == jobs.RunLifeCycleStatePending ||
+		status == jobs.RunLifeCycleStateRunning ||
+		status == jobs.RunLifeCycleStateTerminating {
+		return shared.RunningExecutionStatus, nil
+	} else if status == jobs.RunLifeCycleStateInternalError {
+		return shared.FailedExecutionStatus, nil
+	} else if status == jobs.RunLifeCycleStateTerminated {
+		getTerminalRunResp, err := j.databricksClient.Jobs.GetRun(ctx, *getRunReq)
```
saurav-c, line 154: Isn't this call redundant since we already got the job runs on line 141?

src/golang/lib/job/databricks.go
```diff
@@ -0,0 +1,189 @@
...
+
+	getRunReq := &jobs.GetRun{
+		RunId: runId,
+	}
+	getRunResp, err := j.databricksClient.Jobs.GetRun(ctx, *getRunReq)
+	if err != nil {
+		return shared.UnknownExecutionStatus, errors.Wrap(err, "Unable to get run from databricks.")
+	}
+
+	status := getRunResp.State.LifeCycleState
```
saurav-c, line 146: nit: use a switch statement here instead, you can have multiple values for each case separated by commas

```
switch getRunResp.State.LifeCycleState {
case jobs.RunLifeCycleStatePending, jobs.RunLifeCycleStateRunning, jobs.RunlifeCycleStateTerminating:
   //
...
}
```

src/golang/lib/job/databricks.go
```diff
@@ -0,0 +1,189 @@
...
+
+}
+
+func (j *DatabricksJobManager) DeployCronJob(
+	ctx context.Context,
+	name string,
+	period string,
+	spec Spec,
+) error {
+	return nil
```
saurav-c, line 176: Can we change this to return an error that deploying cron jobs on databricks is not supported

src/golang/lib/databricks/jobs.go
```diff
@@ -0,0 +1,100 @@
...
+	s3InstanceProfileArn string,
+	pythonFilePath string,
+) (int64, error) {
+	createRequest := &jobs.CreateJob{
+		Name: name,
+		Tasks: []jobs.JobTaskSettings{
+			jobs.JobTaskSettings{
+				TaskKey: name,
+				NewCluster: &clusters.CreateCluster{
+					SparkVersion: "10.4.x-scala2.12",
```
likawind, line 56: should these also go to constants?

src/golang/lib/job/databricks.go
likawind: I also feel these logs can be moved to caller. Not required for this PR though

src/golang/lib/job/databricks.go
hsubbaraj-spiral: Both our package and the external one are named `databricks`, so I thought it'd be better to explicitly alias both of them to avoid confusion. I'll keep the `databricks` name for the external library and alias `lib/databricks` to `databricks_lib`.

src/golang/lib/job/databricks.go
hsubbaraj-spiral: These do work, they just don't read directly into spark so we don't solve the scale issue.

src/golang/cmd/server/handler/connect_integration.go
```diff
@@ -320,6 +320,12 @@ func ValidateConfig(
 		return validateLambdaConfig(ctx, config)
 	}
 
+	if service == integration.Databricks {
```
likawind, line 323: In https://www.notion.so/aqueducthq/In-memory-struct-interface-requirement-collections-brainstorming-be252894dcf2402ba27db28c980c79e6 I had one section to move these `validate` to interface and implementation pattern so that we can avoid these large section of `if` statements

src/golang/lib/collections/integration/constants.go
```diff
@@ -37,7 +38,7 @@ const (
 func ParseService(s string) (Service, error) {
 	svc := Service(s)
 	switch svc {
-	case Postgres, Snowflake, MySql, Redshift, MariaDb, SqlServer, BigQuery, GoogleSheets, Salesforce, S3, Athena, AqueductDemo, Github, Sqlite, Airflow, Kubernetes, GCS, Lambda, MongoDB, Conda:
+	case Postgres, Snowflake, MySql, Redshift, MariaDb, SqlServer, BigQuery, GoogleSheets, Salesforce, S3, Athena, AqueductDemo, Github, Sqlite, Airflow, Kubernetes, GCS, Lambda, MongoDB, Conda, Databricks:
```
likawind, line 41: @saurav-c does newer version of go allow us to this casting better?

likawind: If it's unclear, I think we can add it to https://www.notion.so/aqueducthq/WIP-Case-studies-for-go1-19-migration-ee1761b26e4049bd9784f65cf83a74c3 somewhere to do some research in future

src/golang/lib/job/config.go
```diff
@@ -40,6 +41,23 @@ type LambdaJobManagerConfig struct {
...
+	// WorkspaceURL is the full url for the Databricks workspace that
+	// Aqueduct operators will run on.
+	WorkspaceURL string `yaml:"workspaceUrl" json:"workspace_url"`
+	// AccessToken is a Databricks AccessToken for a workspace. Information on how
+	// to create tokens can be found here: https://docs.databricks.com/dev-tools/auth.html#personal-access-tokens-for-users
+	AccessToken string `yaml:"accessToken" json:"access_token"`
+	// Databricks needs an Instance Profile with S3 permissions in order to access metadata
+	// storage in S3. Information on how to create this can be found here:
+	// https://docs.databricks.com/aws/iam/instance-profile-tutorial.html
+	S3InstanceProfileARN string `yaml:"s3InstanceProfileArn" json:"s3_instance_profile_arn"`
```
likawind, line 54: This should be fine for now, but I think we can have a task to look into merging this with the S3 configs?

src/golang/lib/job/databricks.go
```diff
@@ -0,0 +1,180 @@
...
+	databricksFunctionScript = "aqscript.py"
+	databricksParamScript    = "paramScript.py"
+	databricksMetricScript   = "metricScript.py"
+	databricksDataScript     = "dataScript.py"
+)
+
+type DatabricksJobManager struct {
+	databricksClient *databricks.WorkspaceClient
+	conf             *DatabricksJobManagerConfig
+	runMap           map[string]int64
```
likawind, line 25: I just realized this - the `process` job manager uses a concurrent map. Do we need a concurrent map here as well?

likawind: If we do, I'm fine with doing it in another PR. This PR is already large and we won't release the whole feature anyway

likawind (approved):
Let's fix all linter issues before merging, other than that contents in this PR looks good!

I do have a few questions that may lead to some future work

PR #820, opened 2022-12-17
Eng 2104 [1/2][backend] fix a bug where mongo collection preview crashes

likawind:
This PR introduce backend fixes for mongo data preview crash. The root cause is our current `to_json` data frame serialization is not compatible with Mongo's python types, which relies on `bson`. This randomly cause json serialization to fail.

We introduced a new serialization type `bson_table` to cover this case. For now, the logic is -
* df extracted from mongo is with `bson_table` serialization type.
* operator with any `bson_table` inputs will have `bson_table` to serialize df outputs
It might work by introducing a new artf type, but the current solution allow us to fix the issue with narrower range of change.

I noticed some `enum` migration are not finished in the middle, so there are some package import change mixed up in this PR.

ENG-2104

## Loom demo (if any)
WIP

### DIFF
--- a/sdk/aqueduct/utils/serialization.py
+++ b/sdk/aqueduct/utils/serialization.py
@@ -7,7 +7,6 @@ from typing import Any, Callable, Dict, cast
 import cloudpickle as pickle
 import pandas as pd
 from aqueduct.constants.enums import ArtifactType, SerializationType
-from bson import json_util as bson_json_util
 from PIL import Image
 
 from .function_packaging import _make_temp_dir
@@ -25,7 +24,8 @@ def _read_table_content(content: bytes) -> pd.DataFrame:
 
 
 def _read_bson_table_content(content: bytes) -> pd.DataFrame:
-    return pd.DataFrame.from_records(bson_json_util.loads(content.decode(DEFAULT_ENCODING)))
+    from bson import json_util
+    return pd.DataFrame.from_records(json_util.loads(content.decode(DEFAULT_ENCODING)))
 
 
 def _read_json_content(content: bytes) -> Any:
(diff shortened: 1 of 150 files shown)


sdk/aqueduct/utils/serialization.py
```diff
@@ -147,10 +158,17 @@ def serialize_val(val: Any, serialization_type: SerializationType) -> bytes:
 
 
 def artifact_type_to_serialization_type(
-    artifact_type: ArtifactType, content: Any
+    artifact_type: ArtifactType,
+    # derived_from_bson specifies if the artifact is derived from a bson object
+    # and thus requires bson encoding.
+    # For now, it only applies to data frames extracted / transformed from Mongo.
```
cw75, line 164: "DataFrames"

sdk/aqueduct/utils/utils.py
```diff
@@ -105,7 +105,12 @@ def parse_user_supplied_id(id: Union[str, uuid.UUID]) -> str:
 
 
 def construct_param_spec(val: Any, artifact_type: ArtifactType) -> ParamSpec:
-    serialization_type = artifact_type_to_serialization_type(artifact_type, val)
+    serialization_type = artifact_type_to_serialization_type(
+        artifact_type,
+        # not derived from bson. This is irrelevant for now as we don't support table parameters.
```
cw75, line 110: We support parameters of any type, so I do think we support table parameters; we just don't expect the table parameter to be serialized in the bson form, so I think this change is still fine. If so we just need to update this comment to be more accurate.

cw75 (approved):
@likawind LGTM! Left a couple nits inline. I do think creating a separate artifact type for mongo table is cleaner, as it removes the need for the `derived_from_bson` argument which is not used for other artifact types. So if this fix works for the user, I think we should make an effort to clean this up. It will involve a data migration effort after merging this PR though.

likawind:
@cw75 I agree using a new artifact type could make a few codes cleaner here. This is the task: https://linear.app/aqueducthq/issue/ENG-2153/consider-making-mongo-artifacts-a-new-artf-type-instead-of-a-new

PR #821, opened 2022-12-17
Eng 2104 [2/2][ui] Fix a bug where mongo collection preview crashes

likawind:
This PR introduce necessary UI change to render `bson_table` serialization type in UI, as introduced in #820 .

ENG 2104

## Loom demo (if any)
Verified with a mongo workflow with extract, function, and load operator. We checked the table renders in following pages: table preview in integration page, artifact content preview, and artifact versions in data page.
table preview (does not reproduce original problem):
Screen Shot 2022-12-16 at 5 37 26 PM

artifact preview:
Screen Shot 2022-12-16 at 5 34 03 PM

data page (technically this is not affected by this change):
Screen Shot 2022-12-16 at 5 49 54 PM

### DIFF
--- a/src/ui/common/src/components/tables/PaginatedTable.tsx
+++ b/src/ui/common/src/components/tables/PaginatedTable.tsx
@@ -93,6 +93,9 @@ export const PaginatedTable: React.FC<PaginatedTableProps> = ({ data }) => {
                       const value = row[column.name];
                       let displayedValue = '';
                       if (!!value) {
+                        // when the column type is json or object,
+                        // the value parsed from backend API will be an arbitrary json object.
+                        // Here we need to serialize the object to render it properly.
                         if (
                           column.type === 'json' ||
                           column.type === 'object'
--- a/src/ui/common/src/components/tables/PaginatedTable.tsx
+++ b/src/ui/common/src/components/tables/PaginatedTable.tsx
@@ -91,12 +91,23 @@ export const PaginatedTable: React.FC<PaginatedTableProps> = ({ data }) => {
                   >
                     {columns.map((column, columnIndex) => {
                       const value = row[column.name];
+                      let displayedValue = '';
+                      if (!!value) {
+                        if (
+                          column.type === 'json' ||
+                          column.type === 'object'
+                        ) {
+                          displayedValue = JSON.stringify(value);
+                        } else {
+                          displayedValue = value.toString();
+                        }
+                      }
                       return (
                         <TableCell
                           key={`table-col-${columnIndex}`}
                           align={'left'}
                         >
-                          {value ? value.toString() : ''}
+                          {displayedValue}
                         </TableCell>
                       );
                     })}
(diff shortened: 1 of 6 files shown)


cw75:
@likawind Do you mind dropping some comments in the code and/or in this PR about why we need to add these changes to make mongo table previewable?

likawind:
@cw75 could you elaborate on where you find the code confusing? I agree the block in `PagenatedTable` file is a bit confusing since here the table cell contents are objects.

I personally feel other parts should be straight forward since they just adds changes to make the new `bson_table` serialization type working

src/ui/common/src/reducers/integration.ts
```diff
@@ -154,7 +154,17 @@ export const handleLoadIntegrationObject = createAsyncThunk<
...
-      return JSON.parse(serialized) as Data;
+      const rawData = JSON.parse(serialized);
+      if ('schema' in rawData) {
+        return rawData as Data;
+      }
+
+      // this is a bson_table
+      const rows = rawData as TableRow[];
+      return {
+        schema: inferSchema(rows),
```
cw75, line 165: Is the reason we need `inferSchema` that bson table doesn't have `schema` field populated?

src/ui/common/src/reducers/integration.ts
```diff
@@ -154,7 +154,17 @@ export const handleLoadIntegrationObject = createAsyncThunk<
       return thunkAPI.rejectWithValue(objectResponseBody.error);
     } else {
       const serialized = (objectResponseBody as ObjectPreviewResponse).data;
-      return JSON.parse(serialized) as Data;
+      const rawData = JSON.parse(serialized);
+      if ('schema' in rawData) {
```
cw75, line 158: Seems that here, we are using whether `schema` exist to identify whether it's a bson table, but in the code above we are explicitly checking the serialization type. Is this because here it's hard to get access to that property?

src/ui/common/src/components/tables/PaginatedTable.tsx
```diff
@@ -91,12 +91,23 @@ export const PaginatedTable: React.FC<PaginatedTableProps> = ({ data }) => {
                   >
                     {columns.map((column, columnIndex) => {
                       const value = row[column.name];
+                      let displayedValue = '';
+                      if (!!value) {
+                        if (
+                          column.type === 'json' ||
+                          column.type === 'object'
```
cw75, line 98: Is `column.type === 'object'` because in `infer_schema`, we hardcoded the type to be `object`? Also, why do we need `column.type === 'json'` then?

cw75 (commented):
@likawind I think I know what each block of code is doing, but I was a bit confused about why we need them, so I put some of my "guess" and assumptions inline. If they are correct, maybe we can put them into code comments.

src/ui/common/src/components/tables/PaginatedTable.tsx
likawind: Also, I believe `json` is not used in practice since our normal table schema depends on pandas `dtype` which will make all python objects column to `object`

cw75 (approved):

PR #833, opened 2022-12-23
[1/3] Run SDK Integration Tests against S3 [Backend Bugfix]

kenxu95:
Lead reviewers: @cw75 and @likawind 

A number of backend bugs were discovered while attempting to get our integration test suite working against S3.

- In `data/s3.py`, we were initialize a bytes buffer with non-bytes data, causing the code to crash.
- In `sqlite/operators.go`, we were writing read queries that assumed the load operators would be to relational databases. Since S3 was not relational, these queries would not detect such load operators, causing the `list_workflow_objects` endpoint to omit those operators incorrectly. Fixing up those queries necessitated a cascade of other downstream changes:
     - `list_workflow_objects` still returns a LoadOperator, but this struct contains a Load spec instead of flattened fields now, which generalizes to different integrations better. This means that changes had to be made to the SDK's `flow.list_saved_objects()`  user-facing method, as well as the UI [addressed in a follow-up PR].
     - `delete_workflow` had to be altered to take in this new Load spec variant, since it is expected that a user use the output of `list_saved_objects()` to feed into `delete_workflow`. This also has a UI component to it [addressed in a follow-up PR].

ENG-2110

### DIFF
--- a/sdk/aqueduct/backend/response_models.py
+++ b/sdk/aqueduct/backend/response_models.py
@@ -10,7 +10,7 @@ from aqueduct.constants.enums import (
 )
 from aqueduct.models.artifact import ArtifactMetadata
 from aqueduct.models.dag import Metadata
-from aqueduct.models.operators import Operator
+from aqueduct.models.operators import LoadSpec, Operator
 from aqueduct.utils.utils import human_readable_timestamp
 from pydantic import BaseModel
 
@@ -231,10 +231,7 @@ class SavedObjectUpdate(BaseModel):
     operator_name: str
     modified_at: str
     integration_name: str
-    integration_id: uuid.UUID
-    service: ServiceType
-    object_name: str
-    update_mode: str
+    spec: LoadSpec
 
 
 class ListWorkflowSavedObjectsResponse(BaseModel):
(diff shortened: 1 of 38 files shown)


src/golang/lib/collections/operator/connector/load.go
```diff
@@ -72,3 +73,7 @@ func (l *Load) UnmarshalJSON(data []byte) error {
 	l.Parameters = params
 	return nil
 }
+
+func (l *Load) Scan(value interface{}) error {
+	return utils.ScanJSONB(value, l)
+}
```
kenxu95, line 79: This was necessary to get the `json_extract(operator.spec, "$.load")` query working in sqlite/operators.go

sdk/aqueduct/backend/response_models.py
```diff
@@ -231,10 +231,7 @@ class SavedObjectUpdate(BaseModel):
     operator_name: str
     modified_at: str
     integration_name: str
-    integration_id: uuid.UUID
-    service: ServiceType
-    object_name: str
-    update_mode: str
+    spec: LoadSpec
```
likawind, line 234: Does this need to be `Optional`? I noticed the go struct is a pointer

likawind (approved):
LGTM! One minor nit comment

sdk/aqueduct/backend/response_models.py
kenxu95: Yeah good catch! Since the load spec must exist, I just made the field on a go struct a value instead of pointer.

PR #834, opened 2022-12-23
[2/3] Run SDK Integration Tests against S3 [Testing Changes]

kenxu95:
This PR makes all the changes necessary to run the full suite of SDK integration tests against the S3 data integration: eg. `pytest ... --data data_s3` passes for an S3 integration named "data_s3".

NOTE: an important assumption is that the S3 integration already has the following tables populated (from the demo db).
image
These dataframes are all serialized in parquet format. Diversity of data format is not what this test suite is responsible for.

There will eventually be a script that populates S3 automatically for you from the demo db if you don't have them there yet. Note that the table name in sqlite matches the filepath of the object in S3 exactly.

Notable changes:
- Introduce an `extract()` helper in our integration tests that returns a TableArtifact from any supported integration (that being Snowflake, Demo, and S3 right now).
- Introduce a `@pytest.mark.enable_only_for_data_integration_type(ServiceType.S3, ...)` marker for our tests. There is already one for engines, so this is very similar.

The rest of the PR is really just applying all these changes to the entire test suite.

ENG-2110

### DIFF
--- a/integration_tests/sdk/utils.py
+++ b/integration_tests/sdk/utils.py
@@ -228,23 +228,6 @@ def extract(
     raise Exception("Unexpected data integration type provided in test: %s", type(integration))
 
 
-def get_object_identifier_from_load_spec(load_spec: LoadSpec) -> str:
-    if isinstance(load_spec.parameters, RelationalDBLoadParams):
-        return load_spec.parameters.table
-    elif isinstance(load_spec.parameters, S3LoadParams):
-        return load_spec.parameters.filepath
-    raise Exception("Unsupported load parameters %s." % type(load_spec.parameters))
-
-
-def set_object_identifier_on_load_spec(load_spec: LoadSpec, new_obj_identifier: str) -> str:
-    if isinstance(load_spec.parameters, RelationalDBLoadParams):
-        load_spec.parameters.table = new_obj_identifier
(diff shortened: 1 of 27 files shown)


integration_tests/sdk/utils.py
```diff
@@ -199,28 +204,81 @@ def stop_condition(
...
+    if isinstance(integration, RelationalDBIntegration):
+        return integration.sql(query="SELECT * from %s" % obj_identifier, name=op_name, lazy=lazy)
+    elif isinstance(integration, S3Integration):
+        return integration.file(
+            obj_identifier, ArtifactType.TABLE, "parquet", name=op_name, lazy=lazy
+        )
+    raise Exception("Unexpected data integration type provided in test: %s", type(integration))
+
+
+def get_object_identifier_from_load_spec(load_spec: LoadSpec) -> str:
```
likawind, line 231: Is this method available for all types of `LoadSpec`s? Could we just add an `.object_identifier()` method to `LoadSpec` object?

likawind (approved):
Nice work!

integration_tests/sdk/utils.py
kenxu95: I like this idea. Yeah, I think every data integration must have some location identifier for the object.

PR #835, opened 2022-12-23
[3/3] Run SDK Integration Tests against S3 [UI Fix]

kenxu95:
The backend bug fixes in https://github.com/aqueducthq/aqueduct/pull/833 caused two backend routes used by the UI to change, specifically `delete_workflow.go` and `list_workflow_objects`. This PR updates the UI to be compatible with the new backend, and ensures that we can now display objects in S3 and delete them.

ENG-2110

## Loom demo (if any)
https://www.loom.com/share/4e581cde84d34856b05f23e42923c4d4

### DIFF
--- a/src/ui/common/src/components/workflows/WorkflowSettings.tsx
+++ b/src/ui/common/src/components/workflows/WorkflowSettings.tsx
@@ -50,6 +50,7 @@ import {
   getNextUpdateTime,
   PeriodUnit,
 } from '../../utils/cron';
+import { UpdateMode } from '../../utils/operators';
 import ExecutionStatus, { LoadingStatusEnum } from '../../utils/shared';
 import {
   getSavedObjectIdentifier,
@@ -491,7 +492,7 @@ const WorkflowSettings: React.FC<WorkflowSettingsProps> = ({
         [{integration}] <b>{name}</b>
       </Typography>
 
-      {/* Objects saved into S3 are currently expected to have update_mode === "replace". */}
+      {/* Objects saved into S3 are currently expected to have update_mode === UpdateMode.replace */}
       {sortedObjects && (
         <Typography
           style={{
@@ -504,7 +505,8 @@ const WorkflowSettings: React.FC<WorkflowSettingsProps> = ({
           Update Mode:{' '}
           {sortedObjects
             .map(
-              (object) => `${object.spec.parameters.update_mode || 'replace'}`
+              (object) =>
+                `${object.spec.parameters.update_mode || UpdateMode.replace}`
             )
             .join(', ')}
           {sortedObjects.length > 1 && ' (active)'}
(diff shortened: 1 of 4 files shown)


src/ui/common/src/components/workflows/WorkflowSettings.tsx
```diff
@@ -499,7 +502,11 @@ const WorkflowSettings: React.FC<WorkflowSettingsProps> = ({
           display="inline"
         >
           Update Mode:{' '}
-          {sortedObjects.map((object) => `${object.update_mode}`).join(', ')}
+          {sortedObjects
+            .map(
+              (object) => `${object.spec.parameters.update_mode || 'replace'}`
```
kenxu95, line 507: The 'replace' is currently hard coded because S3LoadParams don't have an `update_mode` field.

src/ui/common/src/utils/workflows.tsx
```diff
@@ -135,14 +141,21 @@ export type GetWorkflowResponse = {
   workflow_dag_results: WorkflowDagResultSummary[];
 };
 
+export function getSavedObjectIdentifier(savedObject: SavedObject): string {
+  if ('format' in savedObject.spec.parameters) {
```
kenxu95, line 145: I'm not sure if we have a recommended way of switching on a type? I tried the `as` syntax to check the type but it didn't do what I wanted.

likawind: Is there a `ServiceType` field we can use anywhere? That should be safer than examining fields. @agiron123 should know better on doing the casting part properly.

likawind (approved):
Nice demo! I don't think anything is blocking, let's see if we can improve on the type casting part.

src/ui/common/src/utils/workflows.tsx
kenxu95: Yeah that seems better - I'll check check `ServiceType` in the conditionals.

src/ui/common/src/utils/workflows.tsx
```diff
@@ -135,14 +143,22 @@ export type GetWorkflowResponse = {
   workflow_dag_results: WorkflowDagResultSummary[];
 };
 
+export function getSavedObjectIdentifier(savedObject: SavedObject): string {
+  const serviceString: string = savedObject.spec.service;
+  if (Object.values<string>(RelationalDBServices).includes(serviceString)) {
```
kenxu95, line 148: I followed on of the answers in the top answer here: https://stackoverflow.com/questions/43804805/check-if-value-exists-in-enum-in-typescript

It's a little roundabout imo, but can test to make sure.

src/ui/common/src/components/workflows/WorkflowSettings.tsx
agiron123: nit: Can we make a const enum for the update modes that we have. UPDATE_MODE.REPLACE would be nice to be able to use here.

src/ui/common/src/utils/workflows.tsx
agiron123: Looks fine to me. Read through the StackOverflow earlier today and a lot of this stuff boils down to how our code gets transpiled into JS.

Does look a little roundabout, but it is quite type safe in it's check :)

agiron123 (approved):

PR #837, opened 2023-01-02
Implement basic usage stats collection and reporting

cw75:
This PR enables basic usage stats collection and reporting. Currently, for each request, we collect 1) a privacy-preserving ID of the user's host machine, 2) the name of the route being hit (any uuid in the route is being stripped away), and 3) the latency of the request. We can add more stats as necessary. This usage data is then sent to a Loki server, which is a log aggregation system that I've been playing around with.

Usage data collection is enabled by default, and can be disabled by running `aqueduct start` with `--disable-usage-report`.

### DIFF
--- a/src/golang/cmd/server/middleware/usage/usage.go
+++ b/src/golang/cmd/server/middleware/usage/usage.go
@@ -4,30 +4,18 @@ import (
 	"bytes"
 	"encoding/json"
 	"net/http"
+	"strconv"
 	"strings"
 	"time"
 
 	"github.com/denisbrodbeck/machineid"
+	"github.com/go-chi/chi/middleware"
 	"github.com/google/uuid"
 	log "github.com/sirupsen/logrus"
 )
 
-const (
-	obfuscated string = "***"
-	delimiter  string = "/"
-	hashKey    string = "aqueduct"
-	// This is the Grafana Loki server address
-	logURL string = "http://34.25.83.71:9090/loki/api/v1/push"
-)
-
-type UsageStats struct {
-	ID      string `json:"id"`
-	Route   string `json:"route"`
-	Latency int64  `json:"latency"`
-}
-
-func reportUsage(startTime time.Time, r *http.Request) {
-	pathToken := strings.Split(r.URL.Path, delimiter)
+func reportUsage(startTime time.Time, environment string, statusCode int, urlPath string) {
+	pathToken := strings.Split(urlPath, delimiter)
 	for i, token := range pathToken {
 		if _, err := uuid.Parse(token); err == nil {
 			pathToken[i] = obfuscated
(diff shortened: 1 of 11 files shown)


src/golang/cmd/server/middleware/usage/usage.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/google/uuid"
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	obfuscated string = "***"
+	delimiter  string = "/"
+	hashKey    string = "aqueduct"
+	// This is the Grafana Loki server address
+	logURL string = "http://34.25.83.71:9090/loki/api/v1/push"
```
kenxu95, line 20: I don't know how this works in other open-source projects, but is it dangerous to expose the address of our metrics server publicly? It doesn't seem like pushing to this needs any credentials, so would anyone be able to flood Loki with requests?

src/golang/cmd/server/middleware/usage/usage.go
```diff
@@ -0,0 +1,86 @@
...
+		return
+	}
+
+	usage := UsageStats{
+		ID:      machineID,
+		Route:   strings.Join(pathToken, delimiter),
+		Latency: time.Since(startTime).Milliseconds(),
+	}
+
+	log.Errorf("This request took %d ms, request URL path: %s, machine id: %s.", usage.Latency, usage.Route, usage.ID)
```
kenxu95, line 51: Should this be log.Info()?

kenxu95 (approved):

src/golang/cmd/server/middleware/usage/usage.go
cw75: This was actually meant to be a debug log. I'll remove it when merging.

src/golang/cmd/server/middleware/usage/usage.go
cw75: Yeah this is a good question! On the Loki side we can implement policies to block random requests that don't fit our current schema, but currently there is no way to prevent an attacker from faking a bunch of valid log entries and sending them to our server. I'm looking into how other projects are addressing this.

src/python/bin/aqueduct
```diff
@@ -43,6 +43,9 @@ welcome_message = """
 Your API Key: %s
 
 The Web UI and the backend server (v%s) are accessible at: http://%s:%d/login?apiKey=%s
+
+Aqueduct collects usage data to improve its services.
```
likawind, line 47: I personally feel a bit more clarification might help having user more willing to use this feature. Telling user what's been collected and what they are used for is in fact enforced by regulation for larger apps:
"Aqueduct collects usage data to improve our service performance and scalability. We only collect the frequency of each REST endpoint and their performance stats. We do NOT collect any personal information or anything related to data processed in your workflows.
To disable data collection, run aqueduct service by using `...`"
Alternatively, the long sentence can go to a documentation and we put a link here (maybe as a follow up)

likawind (approved):
Dropped some thoughts on improving the message about data collection. Lmk if it makes sense!

src/python/bin/aqueduct
cw75: Yes, I'm going to make a GitBook page that describes in more detail which stats we are collecting, and add a link to here. Will ship them together.

PR #844, opened 2023-01-04
ENG-1880 Hide preview when artifact is cancelled.

agiron123:
This PR hides artifact previews when the artifact is cancelled.

ENG-1880

## Loom demo (if any)
https://www.loom.com/share/a78d9b2a5a71457090a417439cbb619a

### DIFF
--- a/src/ui/common/src/components/pages/artifact/id/index.tsx
+++ b/src/ui/common/src/components/pages/artifact/id/index.tsx
@@ -84,11 +84,11 @@ const ArtifactDetailsPage: React.FC<ArtifactDetailsPageProps> = ({
 
   const { metrics, checks } =
     !!workflowDagResultWithLoadingStatus &&
-    isSucceeded(workflowDagResultWithLoadingStatus.status)
+      isSucceeded(workflowDagResultWithLoadingStatus.status)
       ? getMetricsAndChecksOnArtifact(
-          workflowDagResultWithLoadingStatus?.result,
-          artifactId
-        )
+        workflowDagResultWithLoadingStatus?.result,
+        artifactId
+      )
       : { metrics: [], checks: [] };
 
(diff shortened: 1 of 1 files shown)


vsreekanti:
@agiron123, thanks for this — two things:
1. Can we show a little info alert just saying that the artifact was never created because an upstream operator failed? That way the user knows what they're looking at and why there's no data.
2. The fact that the status is wrong seems like a really big bug. How hard is that fix going to be?

agiron123:
> @agiron123, thanks for this — two things:
> 
> 1. Can we show a little info alert just saying that the artifact was never created because an upstream operator failed? That way the user knows what they're looking at and why there's no data.
> 2. The fact that the status is wrong seems like a really big bug. How hard is that fix going to be?

@vsreekanti 
re: 1: sure, will get an alert added. 

re: 2: In this particular case, the status of the operator doesn't exist. It comes back as undefined, which could be "Unknown" for all we know. I'm not sure why, just going off of what we get in the frontend here. Something we should look into though.

Something else weird I noticed was when I initially set out to do this, I tried starting a workflow and closing my aqueduct server terminal, and the workflow that was running when this happened is stuck in the pending state. I then ran another workflow and let it complete, and I still see a workflow in the pending state in the version selector.

agiron123:
@vsreekanti here's the update alert:

image

vsreekanti:
Looks good! Can you file a task for the other work? I want to make sure we look into it.

agiron123:
> Looks good! Can you file a task for the other work? I want to make sure we look into it.

@vsreekanti Created an issue here: 
https://linear.app/aqueducthq/issue/ENG-2183/artifact-status-set-to-succeeded-when-artifact-failed

vsreekanti (approved):

PR #847, opened 2023-01-05
Send usage stats to Grafana Loki

cw75:
This PR implements Go structs that can be serialized to Loki's expected log format and send it to Liki for further analysis.
The PR also adds a few new useful fields such as timestamp, environment (dev, test, or prod), status code, and log schema version number.

By default, the server runs in the "prod" mode. When we are doing dev, we should change the field via `aqueduct start --env=dev`. This will help Loki distinguish which logs are from our internal dev and which are from real users. We should modify our GitHub actions to use `aqueduct start --env=test` in the future.


src/golang/cmd/server/server/aqueduct_server.go
```diff
@@ -63,10 +63,11 @@ type AqServer struct {
 	// are no more active requests.
 	RequestMutex sync.RWMutex
 
+	Environment       string
```
kenxu95, line 66: Can you add a short description here too?

src/golang/cmd/server/middleware/usage/utils.go
```diff
@@ -0,0 +1,24 @@
+package usage
```
kenxu95, line 1: nit: Is there a more specific file name we can use than `utils.go`. I'm trying to get us off of dumping everything in utils.go/py if possible - have been using the term `models` for structs like these but not sure.

kenxu95 (approved):
Cool! Left a few non-blocking requests :)

src/golang/cmd/server/middleware/usage/utils.go
cw75: `models.go` sounds good

PR #848, opened 2023-01-05
[1/3] Refactor SDK Integration Tests into subdirectory with shared sibling directory

kenxu95:
Note for @cw75 I know we called this end-to-end tests before, but I think calling them `SDK Aqueduct Integration Tests` is more accurate.

In order to unlock our data integration test suite, which is in a follow up PR, we need to push our SDK Aqueduct Tests down to subdirectory `integration_tests/sdk/aqueduct_tests`, and pull out any shared helpers into `integration_tests/sdk/shared`.

Some notes:
- `integration_tests/sdk/conftest.py` and `integration_tests/sdk/aqueduct_tests/conftest.py` both apply to the `aqueduct_tests` suite.
- The commands have been updated here in the README.md, but not yet in the github actions.

ENG-2113


integration_tests/sdk/aqueduct_tests/operator_test.py
```diff
@@ -0,0 +1,44 @@
...
+        return df
+
+    output_artifact_from_decorator = dummy_sentiment_model(table_artifact)
+    df_normal = output_artifact_from_decorator.get()
+    output_artifact_from_to_operator = to_operator(dummy_sentiment_model_func)(table_artifact)
+    df_func = output_artifact_from_to_operator.get()
+
+    assert df_normal["positivity"].equals(df_func["positivity"])
+
+
```
kenxu95, line 29: For some reason, moving the tests down one and adding __init__.py to both the parent and child directories causes this test to fail (but everything still works outside of pytest). Any idea what's happening here? I put some more information in the Linear task itself.

likawind (commented):
Thanks for putting more organizations! I'm a bit confused on the hierarchy on a few things (maybe because I don't understand properly and we should clarify in README):
* I feel 'aqueduct tests' sounds too vague - I've been wondering what's 'aqueduct behavior' and what's not. I've thought about a few alternatives: 'workflow test', 'orchestration test', 'aqueduct engine test'
* We mentioned 'data integration tests' but there's no directory for it? Do we need one?

integration_tests/sdk/aqueduct_tests/operator_test.py
likawind: could it be the line `from sdk.aqueduct_tests. ...`? It seems other files use `from .test_functions ...` instead

integration_tests/sdk/README.md
```diff
@@ -1,18 +1,33 @@
...
 
-These tests run the SDK against an Aqueduct backend. Each test is built to clean up after itself. If it creates a workflow, it will attempt to delete it in the end. Tests can be run in parallel.
+These tests run the SDK against an Aqueduct backend. Each test is built to clean up after itself. If it creates a workflow, it will attempt to delete it in the end. All tests can be run in parallel.
+
+There are two different suites of SDK Integration Tests, each with their own purpose:
+1) Aqueduct Tests: These tests cover all Aqueduct behavior, from a user's perspective. They set up DAGs and usually publish flows. 
+Test cases are generic enough to be reusable across multiple types of data integrations and engines. They are found in the `aqueduct_tests/` folder.
+2) Data Integration Tests: While Aqueduct tests is the defacto test suite for testing Aqueduct behavior, one disadvantage of such
+powerful but generic test cases is that every data integration is different. Unlike compute, each data integration has its own set of
+APIs, abilities, and limitations, and Aqueduct Tests are philosophically less suitable for providing such coverage. These tests are meant
```
cw75, line 10: To make it clearer, instead of "These tests" let's say "Data integration tests"

cw75 (approved):
Left a nit in README but otherwise LGTM.

cw75:
@likawind maybe "Generic Tests"?

kenxu95:
@cw75 @likawind The issue with names like 'workflow test', 'orchestration test', 'aqueduct engine test' is that they don't encompass all the different test cases we have. Eg. we don't always test workflows, or orchestration, or just aqueduct engine.

The point of `aqueduct_tests` is to separate our aqueduct-specific tests from tests that rely on third party integrations (data integration tests).  Agree it's not a clean split since the data integration tests also rely on some Aqueduct functionality. Generic tests is sort of correct, but feels too broad to me. Open to discussion though - since it's just a directory name I don't think it needs to block anything right now.

integration_tests/sdk/aqueduct_tests/operator_test.py
kenxu95: yep I tried that do to no avail

kenxu95:
Skipping because there's a github actions aspect to this that I tested with a manual run.

PR #850, opened 2023-01-05
[3/3] Update the github actions to run the two SDK integration test suites

kenxu95:
Updates our Github Actions to run both SDK Integration Test suites against the Demo DB.

I'll check that this works before merging this stack of PRs.

ENG-2113

### DIFF
--- a/.github/workflows/periodic-integration-tests.yml
+++ b/.github/workflows/periodic-integration-tests.yml
@@ -79,8 +79,7 @@ jobs:
         working-directory: integration_tests/sdk
         env:
           SERVER_ADDRESS: localhost:8080
-          INTEGRATION: aqueduct_demo
-        run: pytest . -rP -n ${{ matrix.concurrency }}
+        run: pytest aqueduct_tests/ -rP -n ${{ matrix.concurrency }}
 
       - uses: actions/upload-artifact@v3
         if: always()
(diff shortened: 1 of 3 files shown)


.github/workflows/periodic-integration-tests.yml
```diff
@@ -79,8 +79,7 @@ jobs:
         working-directory: integration_tests/sdk
         env:
           SERVER_ADDRESS: localhost:8080
-          INTEGRATION: aqueduct_demo
-        run: pytest . -rP -n ${{ matrix.concurrency }}
+        run: pytest aqueduct_tests/ -rP -n ${{ matrix.concurrency }}
```
likawind, line 82: why don't we run data integration tests here?

likawind (approved):
I'm curious if we need to setup snowflake before running the tests here? For sqlite, I would also consider explicitly connect to demo DB as an external integration, since there's slight difference in sqlite code path and demo code path.

.github/workflows/periodic-integration-tests.yml
kenxu95: The periodic integration tests will be improved as part of the automation milestone. I just didn't bother adding them here since the these tests currently contribute very little and are destined to be replaced soon.

kenxu95:
@likawind The external SQLite suggestion is a good idea. Yes, you will need to have the third-party integrations set up with the demo data populated already in order to run this. I'll be writing a script for this to happen automatically.

PR #856, opened 2023-01-07
Adds SDK support for scheduling a workflow to run after a source

saurav-c:
This PR adds support for scheduling a workflow to run based on a cascading update. The `publish_flow` API is changed to now allow users to set a `source_flow_id`. If this field is set then the cron schedule is ignored.

ENG 2179

### DIFF
--- a/sdk/aqueduct/client.py
+++ b/sdk/aqueduct/client.py
@@ -322,7 +322,7 @@ class Client:
         checks: Optional[List[BoolArtifact]] = None,
         k_latest_runs: Optional[int] = None,
         config: Optional[FlowConfig] = None,
-        source_flow_id: Optional[Union[str, uuid.UUID]] = None,
+        source_flow: Optional[Union[Flow, str, uuid.UUID]] = None,
     ) -> Flow:
         """Uploads and kicks off the given flow in the system.
 
@@ -371,10 +371,11 @@ class Client:
                 - engine: Specify where this flow should run with one of your connected integrations.
                 - k_latest_runs: Number of most-recent runs of this flow that Aqueduct should store.
                     Runs outside of this bound are deleted. Defaults to persisting all runs.
-            source_flow_id:
-                Used to identify the source flow for this flow. If not empty, `schedule`
-                will be ignored and this flow will only run after each successful run of the
-                source flow.
+            source_flow:
+                Used to identify the source flow for this flow. This can be identified
+                via an object (Flow), name (str), or id (str or uuid). 
+                If not empty, `schedule` will be ignored and this flow will only 
+                run after each successful run of the source flow.
 
         Raises:
             InvalidUserArgumentException:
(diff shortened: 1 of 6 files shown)


saurav-c:
@vsreekanti for API changes

sdk/aqueduct/client.py
```diff
@@ -369,6 +371,10 @@ def publish_flow(
                 - engine: Specify where this flow should run with one of your connected integrations.
                 - k_latest_runs: Number of most-recent runs of this flow that Aqueduct should store.
                     Runs outside of this bound are deleted. Defaults to persisting all runs.
+            source_flow_id:
```
vsreekanti, line 374: Do we have to require to to be a UUID? Can we accept a name and convert?

vsreekanti (commented):
Mostly looks good — just would prefer to not only accept UUIDs but also names.

sdk/aqueduct/client.py
saurav-c: @vsreekanti Do you want 2 separate parameters:
- `source_flow_id`
- `source_flow_name`
And we use whichever one is set (using the UUID if both are present).
?

vsreekanti: Why don't we have the parameter be called `source_flow`:
1. We look to see if a workflow of that name exists. If it does, we use it.
2. If it doesn't, we try to convert it to a UUID. If it fails we error.
3. If it converts, we look for a workflow with that ID. If it doesn't exist we error; if it does, we use it.

saurav-c: Sounds good

kenxu95: Oh I didn't see Vikram's follow up comment lol, glad we came to the same name conclusion. Agreed it's just some conversion logic.

sdk/aqueduct/models/dag.py
```diff
@@ -28,6 +28,7 @@ class Schedule(BaseModel):
     trigger: Optional[TriggerType] = None
     cron_schedule: str = ""
     disable_manual_trigger: bool = False
+    source_id: Optional[uuid.UUID] = None
```
kenxu95, line 31: Can you add a comment here describing what this represents, and that it is only relevant for a specific trigger type?

sdk/aqueduct/utils/utils.py
```diff
@@ -104,6 +110,16 @@ def parse_user_supplied_id(id: Union[str, uuid.UUID]) -> str:
     return id
 
 
+def parse_user_supplied_id_as_uuid(id: Union[str, uuid.UUID]) -> uuid.UUID:
```
kenxu95, line 113: nit: Can we name the other helper symmetrically with `parse_..._as_str()`  then?

kenxu95 (commented):
Where are the integration tests?

sdk/aqueduct/client.py
saurav-c: Addressed

kenxu95 (commented):
Looks good just one comment. 

Just want to clarify will you add the integration tests with this PR? And will you be able to get that out before release? cc: @vsreekanti @cw75 I'd argue that we should properly add test cases before releasing features, but curious what ya'll think.

saurav-c:
@kenxu95 I'm currently working on the integration tests. I should be able to have that complete within the next hour if that's fine for the release @vsreekanti.

cw75:
Yeah sounds good to me

saurav-c:
@kenxu95 @cw75 Integration test case for this is here https://github.com/aqueducthq/aqueduct/pull/863. (I need to rebase all of these PRs to pick up Kenny's refactoring changes.)

sdk/aqueduct/client.py
```diff
@@ -416,6 +420,38 @@ def publish_flow(
                 "`artifacts` argument must either be an artifact or a list of artifacts."
             )
 
+        if source_flow and schedule != "":
+            raise InvalidUserArgumentException(
+                "Cannot create a flow with a schedule and a source flow, pick one."
```
kenxu95, line 425: super nit: "... both a schedule and.."

kenxu95 (approved):
cool, thanks!

PR #859, opened 2023-01-09
Eng 2125 improve design to better differentiate keys vs values

eunice-chan:
Smaller, light grey text for the keys to better differentiate key vs value.

Eng 2125

## Loom demo (if any)
Screen Shot 2023-01-07 at 8 31 55 AM


agiron123:
Can you provide a screenshot of the Workflows and Data list pages please? I'm not really sure if what I'm looking at in this screenshot properly addresses what the Linear task set out to do, which was to update the differentiation between metrics/check keys and values in the list view:
image

src/ui/common/src/components/tables/OperatorExecStateTable.tsx
```diff
@@ -81,8 +84,17 @@ export const OperatorExecStateTable: React.FC<OperatorExecStateTableProps> = ({
...
                     >
                       {tableType === OperatorExecStateTableType.Metric ||
-                      column.name === 'title' ? (
-                        value.toString()
+                      columnName === 'title' ? (
+                        <Box
+                          sx={
+                            columnName === 'title' && {
+                              color: theme.palette.gray['700'],
+                              fontSize: '12px',
```
agiron123, line 92: Can we try 16px here? Labels seem a bit small compared to their values.

agiron123:
Hmm... 🤔 I'm not too keen on this design. I feel like this design should make the metric's name more pronounced than the value, so kind of the opposite of what's going on here.

What if we simply increased the font weight of the title?

@vsreekanti open to feedback here.

agiron123:
@eunice-chan sorry for the runaround here, but how about we do this instead:
Keep the metric title's fontWeight at 400 and set the fontWeight of the value to be 300. 

Code comes out much simpler here too.

image

agiron123:
Something like this is what I had in mind:
https://github.com/aqueducthq/aqueduct/pull/874

agiron123:
This PR has been resolved here:
https://github.com/aqueducthq/aqueduct/pull/874

PR #861, opened 2023-01-09
[1/3] Adds SourceID field to Workflow.Schedule column

saurav-c:
This PR adds the `SourceID` field to the `Workflow.Schedule` column. This is necessary to support cascading updates. It also adds a method in the `Workflow` repo to get all target workflows for a given workflow. This is used to perform cascading updates at the end of workflow execution.

### DIFF
--- a/integration_tests/sdk/aqueduct_tests/flow_test.py
+++ b/integration_tests/sdk/aqueduct_tests/flow_test.py
@@ -160,6 +160,61 @@ def test_publish_with_schedule(client, flow_name, data_integration, engine):
     )
 
 
+def test_publish_flow_with_cascading_trigger(client, flow_name, data_integration, engine):
+    """Tests publishing a flow that is set to run on a cascading trigger."""
+    table_artifact = extract(data_integration, DataObject.SENTIMENT)
+    output_artifact = dummy_sentiment_model(table_artifact)
+
+    # Create a source flow
+    source_name = flow_name()
+    source_flow = publish_flow_test(
+        client,
+        name=source_name,
+        artifacts=output_artifact,
+        engine=engine,
+        schedule=aqueduct.daily(),
+    )
+
+    # Create a flow that is set to run after the above source_flow
+    name = flow_name()
+    flow = publish_flow_test(
+        client,
+        name=name,
+        artifacts=output_artifact,
+        engine=engine,
+        source_flow=source_flow,
+    )
+
+    # Trigger a run of the source flow
+    trigger_flow_test(
+        client,
+        source_flow,
+    )
+
+    # Verify that there are now 2 runs of flow
+    wait_for_flow_runs(
+        client,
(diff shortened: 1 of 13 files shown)


cw75 (commented):
@saurav-c Is it true that a workflow can only have one trigger type? Say a user configured a workflow to be triggered by another one. But the user may also want to trigger that workflow manually from the UI. Do we support this?

saurav-c:
@cw75 All workflows can be triggered manually regardless of trigger type. This is how the UI works rn.

cw75:
@saurav-c Got it. Another question I had is say I wrote a workflow that runs on Airflow, and want this workflow to be triggered by another workflow. Can I do that? If so what's the trigger type of this workflow?

saurav-c:
@cw75 We currently cannot support cascading updates for Airflow workflows.
1. An Airflow workflow cannot be the source because we cannot trigger new workflow runs at the end of Airflow execution. We will likely never be able to support this.
2.  Currently we cannot support setting a source trigger for an Airflow workflow, because that requires some additional changes to the end of the executor code when all target flows are triggered. I have error checking right now that prevents setting a source trigger for Airflow workflows, but we can remove this once the following task is implemented. I feel like we probably don't need to prioritize this for todays release. https://linear.app/aqueducthq/issue/ENG-2202/enable-providing-a-non-airflow-source-for-an-airflow-workflow

cw75:
@saurav-c got it, is the error checking in another PR? Agreed that we can work on enabling this later after the release.

saurav-c:
@cw75 I just added a commit to #866 to deal with case #2 from above. I had previously forgotten to do so.

cw75 (approved):

likawind:
LGTM, I think @cw75 's point makes sense that the trigger type can be mixed in future (e.g. both following a schedule or triggered manually). I'm also curious if 'airflow' is a trigger or this information can be captured by engine config. It seems the same argument apply to other self-orchestrated engine?

likawind (approved):

saurav-c:
@likawind Let me try to clarify a few things and then we can discuss offline if needed.
1. The following trigger types are currently supported (Periodic, Manual, Airflow, Cascading). 
2. Regardless of the the trigger type, you can always trigger a manual run from the SDK of the UI. This applies to Airflow triggers as well.
3. Moving the Airflow trigger type to be part of engine config is more of an internal refactoring task, but agreed we can do it.

likawind:
Thanks for clarifications! Btw this PR is ready for me, I was asking just to learn future plans -

For 1, I somehow mistyped the original question. I'm more curious if 'cascading' and 'scheduled' should be mutually exclusive moving forward? Or a workflow can be triggered by both?

For 3, I'm wondering what 'airflow' trigger means. Does that mean that airflow owns the workflow triggering and we just care about retrieving metadata of already triggered workflows?

PR #883, opened 2023-01-13
[1/2] Update SDK Integration Tests to run automagicky against various data integrations

kenxu95:
There are some significant test behavior changes that arise from this PR:

We finally introduce the concept of a test configuration file, located at `integration_tests/sdk/test-config.yml`. I've supplied an example format for people to copy over in `test-config-example.yml`. I imagine we'll have the full one stored in an S3 bucket somewhere.

This test configuration file will contain the server address and apikey - we'll no longer need to use environmental variables. The configuration file also contains entries for the different data integrations we will run the tests against. Every test case will run against every data integration defined in this file, unless the `--data` command line flag is supplied! 

Follow-up: Update our github actions and documentation to reflect these new changes.

ENG-1837

### DIFF
--- a/integration_tests/sdk/conftest.py
+++ b/integration_tests/sdk/conftest.py
@@ -5,13 +5,14 @@ from aqueduct.constants.enums import ServiceType
 from aqueduct.models.dag import DAG, Metadata
 
 from aqueduct import Client, globals
+from sdk.setup_integration import list_data_integrations, setup_data_integrations, get_aqueduct_config
 from sdk.shared import globals as test_globals
 from sdk.shared.utils import delete_flow, generate_new_flow_name
 from sdk.shared.validator import Validator
 
 
 def pytest_addoption(parser):
-    parser.addoption(f"--data", action="store", default="aqueduct_demo")
+    parser.addoption(f"--data", action="store", default=None)
     parser.addoption(f"--engine", action="store", default=None)
     parser.addoption(f"--keep-flows", action="store_true", default=False)
 
@@ -30,8 +31,11 @@ def pytest_configure(config):
     )
 
 
-API_KEY_ENV_NAME = "API_KEY"
-SERVER_ADDR_ENV_NAME = "SERVER_ADDRESS"
+# This is guaranteed to run before everything else, as it has the broadest scope and is set to autouse=True:
+# https://docs.pytest.org/en/stable/reference/fixtures.html#fixture-instantiation-order
+@pytest.fixture(scope="session", autouse=True)
+def setup_integrations():
+    setup_data_integrations()
 
 
 @pytest.fixture(scope="function")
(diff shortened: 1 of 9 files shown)


integration_tests/sdk/conftest.py
```diff
@@ -1,17 +1,20 @@
...
+    setup_data_integrations,
+)
 from sdk.shared import globals as test_globals
 from sdk.shared.utils import delete_flow, generate_new_flow_name
 from sdk.shared.validator import Validator
 
 
 def pytest_addoption(parser):
-    parser.addoption(f"--data", action="store", default="aqueduct_demo")
+    parser.addoption(f"--data", action="store", default=None)
```
likawind, line 17: why we removed the default here?

likawind: I think I understood now, that this flag now have controls specific data integration to use so we should follow test config file if not specified.

integration_tests/sdk/setup_integration.py
```diff
@@ -0,0 +1,192 @@
...
+    return data_integrations
+
+
+def get_aqueduct_config() -> Tuple[str, str]:
+    # Returns the apikey and server address.
+    test_config = _parse_config_file()
+    assert (
+        "apikey" in test_config and "address" in test_config
+    ), "apikey and address must be set in test-config.yml."
+    return test_config["apikey"], test_config["address"]
```
likawind, line 192: do you plan to use pydantic to parse this in the future?

integration_tests/sdk/conftest.py
kenxu95: yep, exactly :)

likawind (commented):
I assume you have another task for ppl to add new integrations to this suite :)

I have some questions on setup process - is the idea to have all integration start with the same set of data, so that the same tests can run consistently on them?

In this way I feel we could just require external integrations to be empty and set up by copying all data over? Just adding missing ones sounds like we need to ensure 'existing' tables to be consistent with demo. I feel this is rather hard to manage since there are many integrations we need to track.

integration_tests/sdk/test-config-example.yml
```diff
@@ -0,0 +1,29 @@
+apikey: <APIKEY>
```
hsubbaraj-spiral, line 1: Is there a place in documentation where are the required fields for each integration are laid out?

hsubbaraj-spiral (approved):

integration_tests/sdk/setup_integration.py
kenxu95: That's a good idea, made a task https://linear.app/aqueducthq/issue/ENG-2237/use-pydantic-for-test-config-struct-instead-of-dictionary

integration_tests/sdk/test-config-example.yml
kenxu95: For our internal dev, we should just be referencing the pydantic structs in `connect_config.py`. For users, I think adding some documentation would be good, although I don't know how much they would use something like that: https://linear.app/aqueducthq/issue/ENG-2238/add-documentation-for-connecting-integrations-from-the-sdk

kenxu95:
@likawind 

> I assume you have another task for ppl to add new integrations to this suite :)
Yep we'll get the democratizing stuff next week.

> I have some questions on setup process - is the idea to have all integration start with the same set of data, so that the same tests can run consistently on them?
Yep, the idea is to have all integrations start with the same set of data. 


> In this way I feel we could just require external integrations to be empty and set up by copying all data over? Just adding missing ones sounds like we need to ensure 'existing' tables to be consistent with demo. I feel this is rather hard to manage since there are many integrations we need to track.
I'm not totally sure how requiring external integrations to be empty is easier to manage? If I'm running my test suite multiple times, there may be significant overhead having to populate the data each time. Also, if the test is unsure about their demo data fidelity, they can just clear their integration and re-run right? I imagine we'll be "resetting" our integrations periodically anyways to prevent this, along with data buildup.

likawind:
@kenxu95 What you mentioned also makes sense. I'm fine with keeping the current implementation and revisiting once we actually get to managing those integrations.

likawind (approved):

PR #896, opened 2023-01-19
[1/2] Support Large Data on Databricks

hsubbaraj-spiral:
This PR enables Aqueduct to run an entire workflow on DatabricksJobs. It adds a version of `ExecuteDatabricks` that ExecuteWorkflow in AqueductEngine would use. There are some additions in `WorkflowDag` to more easily get the parent operators of a given operator. Following PRs will add support for reading/writing tempDFs within Spark.
 
Eng-2222


src/golang/lib/workflow/dag/workflow_dag.go
```diff
@@ -30,6 +30,9 @@ type WorkflowDag interface {
 	// OperatorInputs returns all the artifacts that are fed as input to the given operator.
 	OperatorInputs(operator.Operator) ([]artifact.Artifact, error)
 
+	// Operators that generate the input artifacts of a given operator.
+	OperatorParents(operator.Operator) ([]operator.Operator, error)
```
hsubbaraj-spiral, line 34: @saurav-c can you review the changes to this file in particular? Wanted to make sure that the additions follow the patterns. I originally was going to do something like ArtifactParent and push the finding of the parent logic to the DatabricksJobManager, but thought any logic involving the operators on a `workflowdag` belong here. Let me know what you think.

src/golang/lib/databricks/jobs.go
```diff
@@ -45,71 +45,122 @@ func CreateJob(
...
-				Libraries: []libraries.Library{
-					{
-						Pypi: &libraries.PythonPyPiLibrary{
-							Package: fmt.Sprintf("aqueduct-ml==%s", lib.ServerVersionNumber),
-						},
-					},
+			},
+			{
+				Pypi: &libraries.PythonPyPiLibrary{
+					Package: "snowflake-sqlalchemy",
```
saurav-c, line 112: Why is the `snowflake` package hardcoded?

src/golang/lib/engine/databricks_executor.go
```diff
@@ -0,0 +1,132 @@
...
+
+	"github.com/aqueducthq/aqueduct/lib/collections/shared"
+	"github.com/aqueducthq/aqueduct/lib/job"
+	dag_utils "github.com/aqueducthq/aqueduct/lib/workflow/dag"
+	"github.com/aqueducthq/aqueduct/lib/workflow/operator"
+	"github.com/databricks/databricks-sdk-go/service/jobs"
+	"github.com/dropbox/godropbox/errors"
+)
+
+func ExecuteDatabricks(
```
saurav-c, line 15: nit: Can you add a comment describing why this logic needs to be implemented separately

src/golang/lib/engine/databricks_executor.go
```diff
@@ -0,0 +1,132 @@
...
+	ctx context.Context,
+	workflowDag dag_utils.WorkflowDag,
```
saurav-c, line 17: nit: rename to just `dag`

src/golang/lib/workflow/dag/workflow_dag.go
```diff
@@ -345,6 +354,21 @@ func (w *workflowDagImpl) OperatorInputs(op operator.Operator) ([]artifact.Artif
 	return artifacts, nil
 }
 
+func (w *workflowDagImpl) OperatorParents(op operator.Operator) ([]operator.Operator, error) {
+	inputArtifactIds, ok := w.opToInputArtifacts[op.ID()]
```
saurav-c, line 358: nit: rename to `inputArtifactIDs`

src/golang/lib/workflow/dag/workflow_dag.go
```diff
@@ -345,6 +354,21 @@ func (w *workflowDagImpl) OperatorInputs(op operator.Operator) ([]artifact.Artif
...
+	if !ok {
+		return nil, errors.Newf("Unable to find operator %s (%s) on dag.", op.ID(), op.Name())
+	}
+
+	// Since each artifact can only be generated by one operator, the size of the input operators
+	// and the size of the input artifacts are the same.
+	ops := make([]operator.Operator, 0, len(inputArtifactIds))
+	for _, artifactId := range inputArtifactIds {
+		ops = append(ops, w.operators[w.artifactToInputOp[artifactId]])
```
saurav-c, line 367: Can we assume that the map lookups will always succeed or should we add error handling?

src/golang/lib/workflow/dag/workflow_dag.go
saurav-c: I think this makes sense to implement here

src/golang/lib/engine/aq_engine.go
```diff
@@ -717,10 +724,65 @@ func (eng *aqEngine) cleanupWorkflow(ctx context.Context, workflowDag dag_utils.
 	}
 }
 
+func (eng *aqEngine) executeWithEngine(
```
saurav-c, line 727: can you link to a todo linear task about how we'll want to change this so that we have a separate databricks `orchestrator`

src/golang/lib/engine/databricks_executor.go
```diff
@@ -0,0 +1,132 @@
...
+	if len(inProgressOps) == 0 {
+		return errors.Newf("No initial operators to schedule.")
+	}
+
+	// Wait a little bit for all active operators to finish before exiting on failure.
+	defer waitForInProgressOperators(ctx, inProgressOps, timeConfig.OperatorPollInterval, timeConfig.CleanupTimeout)
+
+	start := time.Now()
+
+	for len(inProgressOps) > 0 {
```
saurav-c, line 53: So, is it true that we don't need to do the checking of whether or not an operator is ready to run, because we provided a task dependency list when creating the databricks task?

src/golang/lib/job/databricks.go
```diff
@@ -171,3 +110,117 @@ func (j *DatabricksJobManager) EditCronJob(ctx context.Context, name string, cro
...
+	// Get the entrypoint file for Databricks.
+	storageConfig, err := spec.GetStorageConfig()
+	if err != nil {
+		return nil, errors.Wrap(err, "Spec unexpectedly has no storage config.")
+	}
+	scriptFile, specStr, err := j.mapJobTypeToFile(spec)
+	if err != nil {
+		return nil, err
+	}
+	bucket := storageConfig.S3Config.Bucket
```
saurav-c, line 129: Do we ensure that databricks can only be used when the metadata store is S3, so we don't need to explicitly check here?

src/golang/lib/engine/databricks_executor.go
```diff
@@ -0,0 +1,132 @@
...
+	ctx context.Context,
+	workflowDag dag_utils.WorkflowDag,
+	databricksJobManager *job.DatabricksJobManager,
+) ([]jobs.JobTaskSettings, error) {
+	dag := workflowDag
+	var taskList []jobs.JobTaskSettings
+
+	for _, op := range dag.Operators() {
+
+		// Get the upstream dependent operators
```
saurav-c, line 111: Could you clarify how this prevents operators from running multiple times. For example, if i have a simple dag with operators: A --> B --> C, my understanding is that this function will create a task for each of those operators with the appropriate parent dependencies:
- A has none
- B has A
- C has B

But then isn't the `databricksJobManager.LaunchMutlipleTaskJob` call in the function above running all of these tasks as part of one job?

saurav-c (commented):
Left some comments mainly for me to better understand how the operator flattening works, otherwise this looks good once some other things are fixed!

src/golang/lib/engine/aq_engine.go
hsubbaraj-spiral: https://linear.app/aqueducthq/issue/ENG-2295/refactor-executeworkflow-to-abstract-out-the-orchestration-step

src/golang/lib/engine/databricks_executor.go
hsubbaraj-spiral: Yep, with aqueduct orchestration we would check that the operators dependencies had completed before calling op.Launch(). Once we submit the workflow job to Databricks it handles the dependency graph and launching jobs in the correct order.

src/golang/lib/engine/databricks_executor.go
hsubbaraj-spiral: `CreateTaskList` doesn't actually launch any jobs. Here we are just assembling that task dependency graph that databricks needs. The translation works like this:
Workflow -> DatabricksJob
Operator -> Task 

A task is considered a "step" within a job. When the dependent tasks are complete Databricks' own scheduler knows to start the next task.

Within each task specification we have to provide the parent tasks (Databricks has no notion of input/output artifacts so this is how they create a dependency graph.) `CreateTaskList` creates a list of all these task specifications.

src/golang/lib/job/databricks.go
hsubbaraj-spiral: https://github.com/aqueducthq/aqueduct/blob/c56eb2e89b76d8064436fcefec4f7f0c6e9134ef/src/golang/lib/workflow/operator/config.go#L131 we do that here.

src/golang/lib/workflow/dag/workflow_dag.go
hsubbaraj-spiral: Yep we should add some error handling here, will do.

src/golang/lib/databricks/jobs.go
hsubbaraj-spiral: This needs to be updated. Thanks for the catch!

hsubbaraj-spiral:
Everything has been addressed other than the Snowflake library dependency.

src/golang/lib/databricks/jobs.go
```diff
@@ -95,21 +86,95 @@ func CreateJob(
...
+			{
+				Pypi: &libraries.PythonPyPiLibrary{
+					Package: fmt.Sprintf("aqueduct-ml==%v", lib.ServerVersionNumber),
+				},
+			},
+			{
+				Pypi: &libraries.PythonPyPiLibrary{
+					Package: "snowflake-sqlalchemy",
+				},
+			},
```
likawind, line 124: I have similar question if we should capture any requirement.txt dependencies here, or this can be done in follow-ups?

src/golang/lib/job/databricks.go
```diff
@@ -171,3 +112,118 @@ func (j *DatabricksJobManager) EditCronJob(ctx context.Context, name string, cro
...
+	if err != nil {
+		return -1, systemError(errors.Wrap(err, "Error creating the task id map."))
+	}
+	for taskName, taskID := range taskMap {
+		j.runMap[taskName] = taskID
+	}
+	return runID, nil
+}
+
+func (j *DatabricksJobManager) mapJobTypeToFile(spec Spec) (string, string, error) {
```
likawind, line 179: It's not very obvious what this function do, some comments would be helpful!

src/golang/lib/job/databricks.go
```diff
@@ -171,3 +112,118 @@ func (j *DatabricksJobManager) EditCronJob(ctx context.Context, name string, cro
...
+	// Add S3 Access Keys to all specs
+	storageConfig, err := spec.GetStorageConfig()
+	if err != nil {
+		return "", "", errors.Wrap(err, "Spec unexpectedly has no storage config.")
+	}
+	storageConfig.S3Config.AWSAccessKeyID = j.conf.AwsAccessKeyID
```
likawind, line 185: I'm a bit confused why there has to be an assignment here. Should the spec already have the right values specified?

It appears the databricks integration has some AWS credentials associated with it (stored in job manager), and this credential could override the one in storage config?

src/golang/lib/job/databricks.go
```diff
@@ -171,3 +112,118 @@ func (j *DatabricksJobManager) EditCronJob(ctx context.Context, name string, cro
...
+			return "", "", errors.Wrap(err, "Unable to encode spec.")
+		}
+		return databricks_lib.DatabricksFunctionScript, specStr, nil
+
+	} else if spec.Type() == ParamJobType {
+		specStr, err := EncodeSpec(spec, JsonSerializationType)
+		if err != nil {
+			return "", "", errors.Wrap(err, "Unable to encode spec.")
+		}
+		return databricks_lib.DatabricksParamScript, specStr, nil
```
likawind, line 205: are these 'Script' file names?

src/golang/lib/engine/aq_engine.go
```diff
@@ -717,10 +724,65 @@ func (eng *aqEngine) cleanupWorkflow(ctx context.Context, workflowDag dag_utils.
...
+	ctx context.Context,
+	workflowDag dag_utils.WorkflowDag,
```
likawind, line 729: similarly, rename to 'dag'

src/golang/lib/engine/aq_engine.go
```diff
@@ -717,10 +724,65 @@ func (eng *aqEngine) cleanupWorkflow(ctx context.Context, workflowDag dag_utils.
...
+		}
+
+		return ExecuteDatabricks(
+			ctx,
+			workflowDag,
+			workflowName,
+			workflowRunMetadata,
+			timeConfig,
+			opExecMode,
+			databricksJobManager,
```
likawind, line 767: As a follow up, why can't databrick be a separate engine implementing `execute()` with the same interface?

src/golang/lib/engine/databricks_executor.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	// Wait a little bit for all active operators to finish before exiting on failure.
+	defer waitForInProgressOperators(ctx, inProgressOps, timeConfig.OperatorPollInterval, timeConfig.CleanupTimeout)
+
+	start := time.Now()
+
+	for len(inProgressOps) > 0 {
+		if time.Since(start) > timeConfig.ExecTimeout {
+			return errors.New("Reached timeout waiting for workflow to complete.")
```
likawind, line 63: do we and should we kill all active jobs when trying to return error from this method?

src/golang/lib/engine/databricks_executor.go
```diff
@@ -0,0 +1,144 @@
...
+				return err
+			}
+
+			if execState.Status == shared.PendingExecutionStatus {
+				continue
+			} else if execState.Status == shared.RunningExecutionStatus {
+				continue
+			}
+
+			if !execState.Terminated() {
```
likawind, line 81: why the above two blocks couldn't be `if !execState.Terminated() { continue; }` ... and execute the rest without this check?

src/golang/lib/engine/databricks_executor.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/aqueducthq/aqueduct/lib/workflow/operator"
+	"github.com/databricks/databricks-sdk-go/service/jobs"
+	"github.com/dropbox/godropbox/errors"
+	"github.com/sirupsen/logrus"
+)
+
+// We seperate out the execution step for Databricks Jobs since
+// Databricks takes care of launching Tasks.
+// Steps:
+// 1. Convert each operator into a Task (includes parent dependency).
```
likawind, line 19: so here the main difference is that we don't schedule downstreams when one op succeeded? And the dependencies are handled by databricks?

src/golang/lib/engine/databricks_executor.go
```diff
@@ -0,0 +1,144 @@
...
+			}
+
+			// Add the operator to the completed stack, and remove it from the in-progress one.
+			if _, ok := completedOps[op.ID()]; ok {
+				return errors.Newf("Internal error: operator %s was completed twice.", op.Name())
+			}
+			completedOps[op.ID()] = op
+			delete(inProgressOps, op.ID())
+
+			time.Sleep(timeConfig.OperatorPollInterval)
```
likawind, line 100: Is this intentional to wait this interval for every single operator? I'm imagining we can do a batch poll (and even export if databrick has API to do this), wait for some interval, and do another batch poll

likawind: I think it's more critical if we need a good balance of execution time and rate limit. If neither is a problem we can solve it later

likawind (changes requested):
Great work! I have a few questions but feel free to address them later if you feel it's more proper.

src/golang/lib/databricks/jobs.go
saurav-c: Is this going to be updated as part of this PR?

src/golang/lib/engine/databricks_executor.go
```diff
@@ -0,0 +1,144 @@
...
+
+	for len(inProgressOps) > 0 {
+		if time.Since(start) > timeConfig.ExecTimeout {
+			return errors.New("Reached timeout waiting for workflow to complete.")
+		}
+
+		for _, op := range inProgressOps {
+			// Poll on the individual operator
+			execState, err := op.Poll(ctx)
+			logrus.Infof("execState of operator:%v", op.JobSpec().JobName())
```
saurav-c, line 69: nit: reminder to remove log statements

src/golang/lib/engine/databricks_executor.go
```diff
@@ -0,0 +1,144 @@
...
+			logrus.Info(execState.Status)
+			if err != nil {
+				return err
+			}
+
+			if execState.Status == shared.PendingExecutionStatus {
+				continue
+			} else if execState.Status == shared.RunningExecutionStatus {
```
saurav-c, line 77: nit: this can be an `if` instead of an `else if`, since you are doing a `continue` in the previous block.

src/golang/lib/job/databricks.go
saurav-c: @likawind not necessarily, the storage config could be using a filepath for credentials instead of the actual keys. So this logic is necessary here to ensure that we pass the actual values of the keys. A similar pattern exists for other job managers, such as lambda [here](https://github.com/aqueducthq/aqueduct/blob/5430c04d2e73e52b52b81349cf646657c37a6de9/src/golang/lib/job/lambda.go#L153)

src/golang/lib/workflow/dag/workflow_dag.go
```diff
@@ -53,11 +56,16 @@ type workflowDagImpl struct {
...
+	artifacts map[uuid.UUID]artifact.Artifact
+	// Operator -> Artifacts generated as outputs.
 	opToOutputArtifacts map[uuid.UUID][]uuid.UUID
-	opToInputArtifacts  map[uuid.UUID][]uuid.UUID
-	artifactToOps       map[uuid.UUID][]uuid.UUID
+	// Operator -> Artifacts that are used as inputs.
+	opToInputArtifacts map[uuid.UUID][]uuid.UUID
+	// Artifact -> Operators that use artifact as input. Can be many operators that use same artifact as input.
+	artifactToOutputOps map[uuid.UUID][]uuid.UUID
+	// Artifact -> Operator that generated artifact as ouput. An artifact can only be generated by one operator.
```
saurav-c, line 67: nit: reminder to check linting issues

saurav-c (commented):
left some additional comments, i think once those and @likawind's comments are addressed this is good to go!

src/golang/lib/job/databricks.go
hsubbaraj-spiral: Yes, these are the entrypoint file names used by Databricks Jobs. These files are uploaded to S3 when a user connects to a Databricks integration!

src/golang/lib/engine/databricks_executor.go
hsubbaraj-spiral: Databricks takes care of killing downstream tasks if an upstream task fails. There should be a more graceful exit in this method however, I'll add that in.

src/golang/lib/engine/databricks_executor.go
hsubbaraj-spiral: Yep this is true, will change. I originally had some different (and unneeded) logic for a Pending state operator, but didn't edit out the different blocks.

src/golang/lib/engine/databricks_executor.go
hsubbaraj-spiral: Yes, while executing with Databricks Jobs, we don't schedule any individual operators. We package all of them in one single job and upload it to Databricks. Their task orchestrator takes care of launching tasks once the appropriate upstream operators are finished.

src/golang/lib/engine/databricks_executor.go
hsubbaraj-spiral: Will get rid of these two blocks in favor of 
```
if !execState.Terminated() { continue }
```

src/golang/lib/engine/databricks_executor.go
likawind: cc @kenxu95 I think so! https://github.com/aqueducthq/aqueduct/blob/main/src/golang/lib/engine/aq_engine.go

src/golang/lib/engine/databricks_executor.go
```diff
@@ -0,0 +1,144 @@
...
+	dag := workflowDag
+
+	// Convert the operators into tasks
+	taskList, err := CreateTaskList(ctx, dag, workflowName, databricksJobManager)
+	if err != nil {
+		return errors.Wrap(err, "Unable to convert operators to Databricks tasks.")
+	}
+
+	// Launch the workflow job with all tasks
+	_, err = databricksJobManager.LaunchMutlipleTaskJob(ctx, workflowName, taskList)
```
kenxu95, line 43: typo: "Multiple"

src/golang/lib/job/databricks.go
```diff
@@ -126,21 +68,20 @@ func (j *DatabricksJobManager) Poll(ctx context.Context, name string) (shared.Ex
...
-	getRunReq := &jobs.GetRun{
-		RunId: runId,
-	}
-	getRunResp, err := j.databricksClient.Jobs.GetRun(ctx, *getRunReq)
+	runResp, err := databricks_lib.GetRun(ctx, j.databricksClient, runId)
 	if err != nil {
 		return shared.UnknownExecutionStatus, systemError(errors.Wrap(err, "Unable to get run from databricks."))
 	}
+	log.Info("Poll response")
+	log.Info(runResp.State)
```
kenxu95, line 76: nit: remove logs before merging.

src/golang/lib/databricks/jobs.go
hsubbaraj-spiral: Going to do this in a follow up PR.

src/golang/lib/databricks/jobs.go
hsubbaraj-spiral: requirements.txt dependencies are captured in the function execution, this is only for the python packages required for the data connector.

saurav-c (approved):
LGTM, reminder to check linter error. Are you planning to run integration tests on this before merging?

likawind (approved):

likawind:
Great job, thanks!

PR #897, opened 2023-01-19
[2/2][UI] Eng 2193 Allow connecting to email as integration

likawind:
## Describe your changes and why you are making these changes
This PR adds UI changes for email connection. Verified (as shown in demo) that connect, edit, and deletion works properly.

The only thing for feedback is the notification level selector that automatically covers 'lower' level of notifications when certain level is selected.

## Related issue number (if any)
ENG-2193

## Loom demo (if any)
https://www.loom.com/share/a41ad69dece24b6c9236b3fd30d0a4dd
Add new integration: 0 - 1:05
Level selector: 37s
Edit: 1:05 - 1:30
Delete: 1:30 - end

## Checklist before requesting a review
- [ ] I have created a descriptive PR title. The PR title should complete the sentence "This PR...".
- [ ] I have performed a self-review of my code.
- [ ] I have included a small demo of the changes. For the UI, this would be a screenshot or a Loom video.
- [ ] If this is a new feature, I have added unit tests and integration tests.
- [ ] I have run the integration tests locally and they are passing.
- [ ] I have run the linter script locally (See `python3 scripts/run_linters.py -h` for usage).
- [ ] All features on the UI continue to work correctly.
- [ ] Added one of the following CI labels:
    - `run_integration_test`: Runs integration tests
    - `skip_integration_test`: Skips integration tests (Should be used when changes are ONLY documentation/UI)

PR #898, opened 2023-01-19
[1/2][Backend] Eng 2193 add email integration to backend

likawind:
This PR adds email integration to backend. It mainly follows current patterns.

The only thing to note is the `target_serialized` field in config which stores the list of receivers.

It appears after I reinstalled all linters for go1.19, some files get changed. These changes do make sense to me.

ENG-2193

## Loom demo (if any)
See #897 


src/golang/lib/collections/shared/types.go
```diff
@@ -30,6 +30,14 @@ const (
 	UnknownExecutionStatus    ExecutionStatus = "unknown"
 )
 
+type NotificationLevel string
```
saurav-c, line 33: Can you use the constants defined [here] (https://github.com/aqueducthq/aqueduct/blob/main/src/golang/lib/models/shared/notification.go#L10) instead of redefining them. I'd like us to deprecate `lib/collections`.

src/golang/lib/lib_utils/utils.go
```diff
@@ -113,3 +114,35 @@ func ParseDatabricksConfig(conf auth.Config) (*integration.DatabricksIntegration
...
+		Port              string                   `json:"port" yaml:"port"`
+		TargetsSerialized string                   `json:"targets_serialized" yaml:"targets_serialized"`
+		Level             shared.NotificationLevel `json:"level"`
+	}
+	if err := json.Unmarshal(data, &c); err != nil {
+		return nil, err
+	}
+
+	var targets []string
+	if err := json.Unmarshal([]byte(c.TargetsSerialized), &targets); err != nil {
```
saurav-c, line 137: Is it possible to do all the unmarshalling in one step if `TargetSerialized` in the struct above is of type `[]string`?

src/golang/lib/notification/email.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"crypto/tls"
+	"net/smtp"
+
+	"github.com/aqueducthq/aqueduct/lib/collections/integration"
+)
+
+func AuthenticateEmail(conf *integration.EmailConfig) error {
+	// Reference: https://gist.github.com/jim3ma/b5c9edeac77ac92157f8f8affa290f45
+	auth := smtp.PlainAuth("", conf.User, conf.Password, conf.Host)
```
saurav-c, line 12: can you add an inline comment for what the "" argument is for

saurav-c (approved):
Left a few minor requests, otherwise LGTM!

src/golang/lib/lib_utils/utils.go
likawind: No, since the config stores a `map[string]string`, the raw json value will be `{..., "target_serialized": "[\"addr\"]"}` which requires another layer of deserialization. The issue comes from the fact that a config value must be a string, and we need the conversion if we want to use it as something else

saurav-c: makes sense!

src/golang/config/config_test.go
```diff
@@ -91,7 +91,7 @@ func setup(t *testing.T) {
 	data, err := yaml.Marshal(testConfig)
 	require.Nil(t, err)
 
-	err = ioutil.WriteFile(testConfigPath, data, 0644)
+	err = ioutil.WriteFile(testConfigPath, data, 0o644)
```
kenxu95, line 94: Is this an intended change?

src/golang/lib/models/shared/integration_config.go
```diff
@@ -43,6 +44,19 @@ type K8sIntegrationConfig struct {
 	UseSameCluster ConfigBool `json:"use_same_cluster"  yaml:"useSameCluster"`
 }
 
+type EmailConfig struct {
+	User     string            `json:"user" yaml:"user"`
+	Password string            `json:"password" yaml:"password"`
+	Host     string            `json:"host" yaml:"host"`
+	Port     string            `json:"port" yaml:"port"`
+	Targets  []string          `json:"targets" yaml:"targets"`
```
kenxu95, line 52: Can you add a comment about what a "Target" refers to?

src/golang/config/config_test.go
likawind: This is generated by my linter

PR #905, opened 2023-01-22
[2/2] Support Large Data on Databricks

hsubbaraj-spiral:
Enable spark temporary views as storage layer for Databricks Jobs Integration
Eng-2043


likawind (commented):
Thanks for the hard work! A lot of these changes look similar to the regular aq executors. I couldn't find much issue after a few glances, but I'm also not sure if there's any place I should take a closer look.

Could you add a few sentences in PR description or in-line explaining changes more specific to spark? That might help myself to understand what to pay attention to

saurav-c:
I have the same question as @likawind. Is there a plan to cut down on redundant code and how we approach changes to the aq executor package going forward?

src/python/aqueduct_executor/operators/connectors/data/execute_spark.py
```diff
@@ -0,0 +1,354 @@
...
+        failure_tip=TIP_DEMO_CONNECTION if is_demo else TIP_INTEGRATION_CONNECTION
+    )
+    def _authenticate() -> None:
+        op = setup_connector(spec.connector_name, spec.connector_config)
+        op.authenticate()
+
+    _authenticate()
+
+
+def run_extract_spark(
```
kenxu95, line 129: Where is this called? `_execute_spark()` above uses `run_extract()`

kenxu95 (commented):
Just took a cursory glance out of curiousity

likawind (commented):
The only thing I'm confused is the relationship between `connectors/data/execute_spark.py` and `spark/execute_data.py`. I get how the latter add spark session for all APIs and pass that session around for storage updates. But what does the first file do? I don't see obvious difference between the current aq executors and I'm also unsure where they should be called.

src/python/aqueduct_executor/operators/utils/utils.py
```diff
@@ -89,6 +90,53 @@ def read_artifacts(
     return inputs, artifact_types, serialization_types
 
 
+def read_artifacts_spark(
```
hsubbaraj-spiral, line 93: `read_artifacts_spark` and `write_artifact_spark` are the main functions to review. Currently the sample method is df.limit(100). Sampling with spark [dataframes requires a fraction](https://spark.apache.org/docs/3.1.2/api/python/reference/api/pyspark.sql.DataFrame.sample.html), and can still result in an overlarge size if the original DF is large enough. I'd like to hear thoughts/suggestions on this approach

hsubbaraj-spiral:
> The only thing I'm confused is the relationship between `connectors/data/execute_spark.py` and `spark/execute_data.py`. I get how the latter add spark session for all APIs and pass that session around for storage updates. But what does the first file do? I don't see obvious difference between the current aq executors and I'm also unsure where they should be called.

Sorry for the confusion. `connectors/data/execute_spark.py ` should not exist. I have removed the file and cut down on redundant code. It should be much cleaner to review now.

likawind (approved):
As long as your tests pass, I don't have additional comments on these changes. 🎉

likawind:
🎉

kenxu95 (approved):

PR #907, opened 2023-01-22
Fix a bug where operator execution fails when running on k8s

cw75:
This was because there's a small time gap between launching a job and spinning up a pod. We capture this case and return a pending status, so that the next time we poll, the pod will be created. I tested manually and it worked, so hopefully it will fix the issue @kenxu95 you ran into as well.

### DIFF
--- a/src/golang/lib/k8s/jobs.go
+++ b/src/golang/lib/k8s/jobs.go
@@ -12,8 +12,6 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
-var ErrNoPodExists = errors.New("No pod exists")
-
 // A helper function that takes in the name of a job, a container image, and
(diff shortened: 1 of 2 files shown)


src/golang/lib/k8s/jobs.go
```diff
@@ -109,6 +112,11 @@ func GetPod(ctx context.Context, name string, k8sClient *kubernetes.Clientset) (
 		return nil, err
 	}
 
+	if len(podList.Items) == 0 {
+		log.Infof("No pod has been created from job %s yet...", name)
+		return nil, ErrNoPodExists
```
kenxu95, line 117: Can we return the error directly? It would be good to get the stack trace.

kenxu95 (approved):

PR #910, opened 2023-01-24
[2/2][UI] Eng 2191 Allow user to connect to slack integration

likawind:
## Describe your changes and why you are making these changes
This PR adds all required UI work to allow user connecting to slack integration. it's very similar to  #897 and reuses the notification level selector.

Tested the following features are working properly:
* connect
* test-connect
* edit
* delete

## Related issue number (if any)
ENG-2191

## Loom demo (if any)
Integration list page:
Screenshot 2023-01-23 at 9 49 04 PM

Connect dialog:
Screenshot 2023-01-23 at 9 48 56 PM

Integration detail page:
Screenshot 2023-01-23 at 9 50 22 PM

## Checklist before requesting a review
- [ ] I have created a descriptive PR title. The PR title should complete the sentence "This PR...".
- [ ] I have performed a self-review of my code.
- [ ] I have included a small demo of the changes. For the UI, this would be a screenshot or a Loom video.
- [ ] If this is a new feature, I have added unit tests and integration tests.
- [ ] I have run the integration tests locally and they are passing.
- [ ] I have run the linter script locally (See `python3 scripts/run_linters.py -h` for usage).
- [ ] All features on the UI continue to work correctly.
- [ ] Added one of the following CI labels:
    - `run_integration_test`: Runs integration tests
    - `skip_integration_test`: Skips integration tests (Should be used when changes are ONLY documentation/UI)

PR #911, opened 2023-01-24
[1/2][Backend] Eng 2191 add slack integration to backend

likawind:
This PR adds backend changes for slack integration. It's very similar to #898 , and covers the necessary auth check when connecting to slack.

A follow up task is https://linear.app/aqueducthq/issue/ENG-2315/[m3]-detect-if-slack-token-has-channel-access-on-auth-time , that we will add checks to ensure the token has permission to all specified channels. We decide to do this later as we will explore slack go APIs more heavily in the next milestone.
 
ENG-2191

## Loom demo (if any)
See #910 

### DIFF
--- a/src/golang/cmd/server/handler/connect_integration.go
+++ b/src/golang/cmd/server/handler/connect_integration.go
@@ -329,6 +329,10 @@ func ValidateConfig(
 		return validateEmailConfig(config)
 	}
 
+	if service == integration.Slack {
+		return validateSlackConfig(config)
+	}
+
 	jobName := fmt.Sprintf("authenticate-operator-%s", uuid.New().String())
 	if service == integration.Conda {
 		return validateConda()
@@ -628,6 +632,19 @@ func validateEmailConfig(config auth.Config) (int, error) {
 	return http.StatusOK, nil
 }
 
+func validateSlackConfig(config auth.Config) (int, error) {
+	slackConfig, err := lib_utils.ParseSlackConfig(config)
+	if err != nil {
+		return http.StatusInternalServerError, err
+	}
+
+	if err := notification.AuthenticateSlack(slackConfig); err != nil {
+		return http.StatusBadRequest, err
+	}
+
+	return http.StatusOK, nil
+}
+
 // ValidatePrerequisites validates if the integration for the given service can be connected at all.
 // For now, it checks if an integration already exists for unique integrations including
 // conda, email, and slack.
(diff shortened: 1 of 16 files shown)


src/golang/lib/models/shared/integration_config.go
```diff
@@ -53,6 +53,12 @@ type EmailConfig struct {
 	Level    NotificationLevel `json:"level" yaml:"level"`
 }
 
+type SlackConfig struct {
+	Token    string            `json:"token" yaml:"token"`
+	Channels []string          `json:"channels" yaml:"channels"`
+	Level    NotificationLevel `json:"level" yaml:"level"`
```
kenxu95, line 59: This config corresponds to a single integration right? If I want to set up different channels with different notification levels, will that require me to set up multiple slack integrations on the UI?

kenxu95: If so, would it be better to have a notification level per channel, so you'd only have to set it up once.

likawind: Yeah this makes sense! But :) - we actually discussed this in a few deliverable meetings, and we explicitly decided to just have one level for all slack channels to start with.

It would require some data migration if we want to switch gear, but there shouldn't be a lot of burden comparing to the additional UI work. So I don't see a strong motivation to just enable this in backend as well.

src/golang/cmd/server/handler/connect_integration.go
```diff
@@ -663,16 +680,17 @@ func ValidatePrerequisites(
...
 		if err != nil {
 			return http.StatusInternalServerError, errors.Wrap(err, "Unable to verify if email is connected.")
 		}
 
-		if len(emailIntegrations) > 0 {
+		if len(integrations) > 0 {
 			return http.StatusBadRequest, errors.Newf(
 				"You already have an email integration %s connected.",
-				emailIntegrations[0].Name,
+				integrations[0].Name,
```
saurav-c, line 693: Change the error message to reflect that this can be a slack integration as well. Same thing for the error message right above this.

src/golang/lib/lib_utils/utils.go
```diff
@@ -146,3 +146,30 @@ func ParseEmailConfig(conf auth.Config) (*shared.EmailConfig, error) {
...
 	}, nil
 }
+
+func ParseSlackConfig(conf auth.Config) (*shared.SlackConfig, error) {
+	data, err := conf.Marshal()
+	if err != nil {
+		return nil, err
+	}
+
+	var c struct {
```
saurav-c, line 156: nit: there's not really a need to use yaml tags unless we are parsing from a yaml file, which we aren't doing in most of these use cases. we can clean up the rest of the codebase later, but just something worth leaving out going forward

saurav-c (approved):
Left a couple nits otherwise LGTM

PR #908, opened 2023-01-23
Eng 2122 unify use of color and size in status indicators

eunice-chan:
## Describe your changes and why you are making these changes
- Workflows Page: Fixed size of indicator next to workflow name
- Data Page: Fixed size of indicator next to workflow name and artifact name
- Check Detail Page: did not update the output icon colored since that is just the icon for the BoolArtifactType.
- Replace colors with more human-understandable names. E.g. theme.palette.Success vs theme.palette.green[’400’]

## Related issue number (if any)
Eng 2122

## Loom demo (if any)
Screen Shot 2023-01-22 at 5 34 39 PM
Screen Shot 2023-01-22 at 5 34 48 PM

## Checklist before requesting a review
- [x] I have created a descriptive PR title. The PR title should complete the sentence "This PR...".
- [x] I have performed a self-review of my code.
- [x] I have included a small demo of the changes. For the UI, this would be a screenshot or a Loom video.
- [N/A] If this is a new feature, I have added unit tests and integration tests.
- [N/A] I have run the integration tests locally and they are passing.
- [x] I have run the linter script locally (See `python3 scripts/run_linters.py -h` for usage).
- [x] All features on the UI continue to work correctly.
- [x] Added one of the following CI labels:
    - `run_integration_test`: Runs integration tests
    - `skip_integration_test`: Skips integration tests (Should be used when changes are ONLY documentation/UI)

PR #914, opened 2023-01-24
eng-2310 Add status in check + metric detail page header

eunice-chan:
Did not have status passed in as an arg before, so the statuses were not displayed properly.

ENG-2310

## Loom demo (if any)
Before:
check_status_header_bug
metric_status_header_bug

After:
check_status_header_fixed
metric_status_header_fixed


vsreekanti:
@eunice-chan, thanks for fixing this! How come the error message "unable to retrieve artifact content..." is shown in a success Alert? 

Also, isn't this expected — if there's an error during the execution of a metric, why are we trying to retrieve the artifact contents? It shouldn't exist.

agiron123:
Will these be the same color as the other icons once your other PR (#908) is merged in? The green in the header looks a little off here. Should be the lighter green like the one we are using in history tab:

image

agiron123:
These look really small, and a little too far from the header text.

Can you bump up the size on these to, say 32px? We probably should have done a little mocking before getting started here.

eunice-chan:
@vsreekanti 
I have been getting these errors due to an issue with my backend--I think it is related to my Python environment which I have been trying to resolve recently (https://aqueducthq.slack.com/archives/C01KF131001/p1674674160894239)

However, you are right it shouldn't be in a success alert. I've created a Linear task for this [here](https://linear.app/aqueducthq/issue/ENG-2357/fix-checkmetric-page-error-alert-to-not-have-success-theme).

After investigating this, I found the bug is also a straightforward change so I've made the change in this PR at:
- src/ui/common/src/components/workflows/artifact/check/history.tsx
- src/ui/common/src/components/workflows/artifact/metric/history.tsx

Here is the updated look:
update_details_alert

@agiron123 
The history list's icons are actually the ones that are off. I can update them to use theme.palette.Success, .Error, etc. I wasn't sure if that was something we wanted to do since the background is colored and the green used for the icon in history is 600 vs the green used in the other success icons (400). I'll update the history list's icon to use the Success, Error colors in the theme. **Let me know if you have any thoughts otherwise.**

You can see the change at:
- src/ui/common/src/components/workflows/artifact/check/history.tsx
- src/ui/common/src/components/workflows/artifact/metric/history.tsx

The icons look small because the size unification PR (https://github.com/aqueducthq/aqueduct/pull/908) hasn't been merged yet. Here is how it looks with the size+color unification PR changes and updating the history icons to use the StatusIndicator component:
updated_checks_page

vsreekanti:
Awesome, thanks @eunice-chan! The updated version you shared here looks exactly right. 🙂

agiron123 (approved):

PR #919, opened 2023-01-25
Adds data integration tests for Redshift

saurav-c:
This PR adds data integration tests for Redshift.

- [x] I ran the tests for Redshift and they all pass.
- [x] I uploaded the Redshift credentials to the shared S3 bucket.
- [x] I have updated instructions in our Notion testing page(with some follow on work to figure out some AWS IAM stuff)

ENG 2287

### DIFF
--- a/integration_tests/sdk/setup_integration.py
+++ b/integration_tests/sdk/setup_integration.py
@@ -111,6 +111,7 @@ def _setup_redshift_data(client: Client, redshift: RelationalDBIntegration) -> N
 
     _add_missing_artifacts(client, redshift, existing_table_names)
 
+
 def _setup_snowflake_data(client: Client, snowflake: RelationalDBIntegration) -> None:
     # Find all the tables that already exist.
     existing_table_names = set(snowflake.list_tables()["tablename"])
--- a/integration_tests/sdk/setup_integration.py
+++ b/integration_tests/sdk/setup_integration.py
@@ -105,6 +105,12 @@ def _add_missing_artifacts(
     )
 
 
+def _setup_redshift_data(client: Client, redshift: RelationalDBIntegration) -> None:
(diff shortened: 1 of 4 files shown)


integration_tests/sdk/setup_integration.py
```diff
@@ -170,7 +176,9 @@ def setup_data_integrations(filter_to: Optional[str] = None) -> None:
 
         # Setup the data in each of these integrations.
         integration = client.integration(integration_name)
-        if integration.type() == ServiceType.SNOWFLAKE:
+        if integration.type() == ServiceType.REDSHIFT:
```
eunice-chan, line 179: Maybe we can make a dictionary with key = integration.type() and value = function (_setup_{integration}_data)

sdk/aqueduct/integrations/connect_config.py
```diff
@@ -43,6 +43,9 @@ class PostgresConfig(BaseConnectionConfig):
     host: str
     port: Optional[str] = "5432"
 
+class RedshiftConfig(PostgresConfig):
+    port: Optional[str] = "5439"
```
kenxu95, line 47: Where does this default port number come from? Can we document it here and in `test-config-example.yml` too? Eg. it would be good to know, if we had a script like `start_redshift.py`, whether the port argument was required or not.

kenxu95 (approved):

sdk/aqueduct/integrations/connect_config.py
saurav-c: Added

integration_tests/sdk/setup_integration.py
saurav-c: It might be the same readability and we'll have to check if the integration exists in the map, so I might leave this as is for now.

PR #926, opened 2023-01-28
Eng 2195 Implement application logic to send notifications

likawind:
This PR covers main application logic to send notifications. We adjust a few interfaces to make things working with the current frameworks:
* added a few metadata interfaces for workflow dag object to get ID, name, and notification settings
* adjusted notification interface to let caller decide when to send notifications. This has clear and straight-forward interface definitions, but is also slightly less extensible. It requires more change if we want to move to channel-specific notification level settings.

ENG-2195

## Loom demo (if any)
For now it compiles.

As a part of this PR, will add some unit tests to cover small helpers.

Full demo is here: https://www.loom.com/share/b7929818b83743d881db8d712675fdc3


src/golang/lib/engine/aq_engine.go
```diff
@@ -760,22 +781,25 @@ func (eng *aqEngine) execute(
...
 				}
 				continue
 			} else if execState.Status == shared.RunningExecutionStatus {
 				continue
 			}
 
 			if !execState.Terminated() {
-				return errors.Newf("Internal error: the operator is expected to have terminated, but instead has status %s", execState.Status)
+				err = errors.Newf("Internal error: the operator is expected to have terminated, but instead has status %s", execState.Status)
+				return err
```
kenxu95, line 794: This pattern is fairly unconventional and I worry that we'll forget to set the error before returning in the future. I'm pretty sure what you can do is name the output argument ala:
```
func (eng *aqEngine) execute(
	ctx context.Context,
	workflowDag dag_utils.WorkflowDag,
	workflowRunMetadata *workflowRunMetadata,
	opExecMode operator.ExecutionMode,
) (err error) {
```

Then you can just `return errors.Newf(...)` directly and that error variable will be set in the defer(). Might be good too double check this but I'm 99% sure I've done this before.

src/golang/lib/engine/aq_engine.go
```diff
@@ -742,13 +747,29 @@ func (eng *aqEngine) execute(
 	}
 
 	// Wait a little bit for all active operators to finish before exiting on failure.
-	defer waitForInProgressOperators(ctx, inProgressOps, timeConfig.OperatorPollInterval, timeConfig.CleanupTimeout)
+	defer func() {
+		waitForInProgressOperators(ctx, inProgressOps, timeConfig.OperatorPollInterval, timeConfig.CleanupTimeout)
+		if err != nil {
+			notificationLevel = mdl_shared.ErrorNotificationLevel
+			notificationCtxMsg = err.Error()
```
kenxu95, line 754: Instead of an explicit `hasNotification` argument, I'd prefer if we defined a struct containing both the level and msg, and set some variable as a pointer to that struct. If that pointer is NULL, then there is no notification. That way it's all packaged behind one variable.

src/golang/lib/engine/aq_engine.go
```diff
@@ -846,14 +884,22 @@ func (eng *aqEngine) execute(
...
-			return errors.Newf("Internal error: operator %s has a non-zero dep count %d.", opID, depCount)
+			err = errors.Newf("Internal error: operator %s has a non-zero dep count %d.", opID, depCount)
+			return err
 		}
 	}
+
+	// avoid overriding a warning notification.
+	if notificationLevel != mdl_shared.WarningNotificationLevel {
+		hasNotification = true
+		notificationLevel = mdl_shared.SuccessNotificationLevel
```
kenxu95, line 901: Another advantage of packaging the level and msg between a struct is that we can write a set() method that enforces notification level logic like this outside of the core orchestration code.

kenxu95: Also, this if-statement is true for ErrorNotificationLevels too right? We wouldn't want to override that either right.

src/golang/lib/engine/notification.go
```diff
@@ -0,0 +1,82 @@
...
+		statusMsg = fmt.Sprintf("succeeded with warning%s", contextSuffix)
+	} else if level == shared.ErrorNotificationLevel {
+		statusMsg = fmt.Sprintf("failed%s", contextSuffix)
+	} else {
+		// For now, no caller will send message other than success, warning, or error.
+		// This line is in case of future use cases.
+		statusMsg = fmt.Sprintf("has a message: %s", contextMsg)
+	}
+
+	// Full message will look like "Workflow my_churn succeeded with warning: some context ."
```
kenxu95, line 40: Can we move this "Full message" comment up to the function signature?

src/golang/lib/notification/notification.go
```diff
@@ -4,25 +4,64 @@ import (
...
-	// The caller always call `Send()` when a notification is generated.
-	// There could be a level preference associated with the notification integration.
-	// For example, slack and email has `level` field in config,
-	// and only notifications beyond this level will be sent.
-	// In such cases, the implementation of `Send()` should reflect the level preference.
-	Send(msg string, level shared.NotificationLevel) error
+	// `ID()` is the unique identifier, typically mapped to the integration ID.
+	ID() uuid.UUID
+
+	// `Level()` is the default level beyond which a notification should send.
```
kenxu95, line 24: Can you define what "beyond which" means here? An example would be nice. Eg. what is sent for a "warning" level.

kenxu95 (changes requested):

src/golang/lib/engine/aq_engine.go
likawind: will this variable always override the current `err` variable set in the function body?

src/golang/lib/engine/aq_engine.go
likawind: Since the next line is `return nil` we should safely assume everything is `error free` and no need to worry about the error level case. Do you feel anything is not self-explanatory here?

src/golang/lib/engine/aq_engine.go
kenxu95: It should be equivalent to a `var err error` defined at the very beginning of the body. Golang should complain if you attempt to do any inappropriate local context overwrites I think?

src/golang/lib/engine/aq_engine.go
kenxu95: I think what I was confused about was that in the error case, we are expected to return an `opFailureError()` before getting here. In that case, can we assert here that the notificationLevel cannot be ERROR?

src/golang/lib/engine/aq_engine.go
```diff
@@ -797,12 +821,25 @@ func (eng *aqEngine) execute(
...
-							return errors.Wrapf(err, "Error when finishing execution of operator %s", op.Name())
+							err = errors.Wrapf(err, "Error when finishing execution of operator %s", op.Name())
+							return err
 						}
 					}
 				}
 
+				hasNotification = true
+				notificationLevel = mdl_shared.ErrorNotificationLevel
+				if execState.Error != nil {
```
kenxu95, line 832: In what cases do we not write to execState.Error? I thought it was expected that this was always written to on error.

src/golang/lib/engine/aq_engine.go
likawind: actually I see what you mean, yeah having the check covering 'ErrorNotfLevel' makes sense. It won't be the case in practice but this code block doesn't need to make this assumption

kenxu95:
BTW is there a demo for this?

src/golang/lib/engine/aq_engine.go
likawind: This is just a defensive check as this field is a pointer and we are accessing subfields in the block. This way if anything is wrong, the nil pointer access won't kill the whole process

likawind:
@kenxu95 updated, thanks for all the advices!

There's no demo at this point of the project since we haven't implement the actual code to send notifications. The behavior will be hidden until https://linear.app/aqueducthq/issue/ENG-2268/[m3]-implement-actual-code-to-send-email-notifications . The change should be safe as long as all integration tests and manual QA passes (both are done for this PR)

src/golang/lib/notification/notification.go
```diff
@@ -55,8 +100,32 @@ func NewNotificationFromIntegration(
...
+
+// `ShouldSend` determines if a notification at 'level' passes configuration
+// specified by `thresholdLevel`.
+// 'info' and 'neutral' will get through regardless of threshold.
+// And 'info' or 'neutral' threshold lets everything through.
+// Other states will follow the severity ordering.
+func ShouldSend(
+	thresholdLevel shared.NotificationLevel,
+	level shared.NotificationLevel,
+) bool {
```
kenxu95, line 117: I have a slight preference of making this lowercase and only use it within send() to simplify the callers. The Send() method noops anyways if content == nil, so it could follow the similar pattern if the severity isn't high enough.

kenxu95 (approved):
Unblocking with some thoughts!

src/golang/lib/notification/notification.go
likawind: The issue is whether to send is decided by caller due to the 'default - override' behavior (see the `sendNotification` implementation in `engine` package). An [alternative design](https://github.com/aqueducthq/aqueduct/pull/923) is to allow caller to pass in the overrides so that we can do this `shouldSend` within the package. But I found the API interface in this alternative is also confusing as we need to explain what 'override' means:

```
wfSettings := dag.Settings() // overrides
notification.Send(
  msg,
  level, // notification level
  wfSettings.Level, // override value
)
```

src/golang/lib/notification/notification.go
```diff
@@ -4,25 +4,70 @@ import (
...
 	//
-	// The caller always call `Send()` when a notification is generated.
-	// There could be a level preference associated with the notification integration.
-	// For example, slack and email has `level` field in config,
-	// and only notifications beyond this level will be sent.
-	// In such cases, the implementation of `Send()` should reflect the level preference.
-	Send(msg string, level shared.NotificationLevel) error
+	// This behavior is controlled by caller calling `ShouldSend()` function.
+	// This field is a 'global default' as we allow overriding this behavior in,
+	// for example, workflow specific settings.
```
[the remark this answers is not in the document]
likawind: Yeah, the `global default` means it applies to all workflows unless separately specified in `workflow settings`. I think what I meant is this `Level()` returned is a default behavior, and caller may not pick this behavior in favor of other specific settings (and those settings are also not visible to this `notification` object)

PR #929, opened 2023-01-30
Fixes bug for listing big query tables

saurav-c:
This PR fixes a bug in the SDK when listing tables in big query. We were previously using an explicit SQL query to look up tables in Big Query, but this query does not work without providing additional information about the Google Cloud project ID. Instead of surfacing this information to the SDK client and having to construct a query, it is better to use our `integration/[]/objects` endpoint, since this achieves the same functionality without having to maintain hardcoded SQL queries. 

We should do the same for the other integrations as well. This task is documented [here](https://linear.app/aqueducthq/issue/ENG-2358/refactor-sdk-list-tables-implementation-to-use-the).

ENG 2340

## Testing
I verified that the `list_tables` API works for a Big Query integration via a test notebook. Full testing of big query will be added as part of the ongoing testing work.

Screen Shot 2023-01-30 at 12 04 36 AM

### DIFF
--- a/sdk/aqueduct/backend/api_client.py
+++ b/sdk/aqueduct/backend/api_client.py
@@ -54,7 +54,7 @@ class APIClient:
     REGISTER_WORKFLOW_ROUTE = "/api/workflow/register"
     REGISTER_AIRFLOW_WORKFLOW_ROUTE = "/api/workflow/register/airflow"
     LIST_INTEGRATIONS_ROUTE = "/api/integrations"
-    LIST_TABLES_ROUTE = "/api/%s/tables"
+    LIST_INTEGRATION_OBJECTS_ROUTE = "/api/integration/%s/objects"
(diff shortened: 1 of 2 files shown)


sdk/aqueduct/backend/api_client.py
```diff
@@ -54,7 +54,7 @@ class APIClient:
     REGISTER_WORKFLOW_ROUTE = "/api/workflow/register"
     REGISTER_AIRFLOW_WORKFLOW_ROUTE = "/api/workflow/register/airflow"
     LIST_INTEGRATIONS_ROUTE = "/api/integrations"
-    LIST_TABLES_ROUTE = "/api/tables"
+    LIST_INTEGRATION_OBJECTS_ROUTE = "/api/integration/%s/objects"
```
cw75, line 57: nit: `LIST_INTEGRATION_OBJECTS_ROUTE_TEMPLATE`

cw75 (approved):
LGTM, left a naming nit inline.

PR #933, opened 2023-01-30
Fix bug where integration credentials aren't properly migrated to the cloud

cw75:
This also fixes ENG-2262: https://linear.app/aqueducthq/issue/ENG-2262/fix-a-bug-where-preview-fails-after-switching-metadata-store-to-cloud#comment-ce9e194a

The root cause was we were not initializing the vault dynamically, so the new metadata vault info wasn't picked up.


src/golang/config/config.go
```diff
@@ -45,8 +45,8 @@ func APIKey() string {
 }
 
 // Storage returns the storage layer config.
-func Storage() shared.StorageConfig {
-	return *globalConfig.StorageConfig
+func Storage() *shared.StorageConfig {
```
cw75, line 48: @saurav-c I found making it returning a pointer is more convenient for other parts of the codebase. Just making sure you are aware of this and there's no side effect.

saurav-c: It's more convenient, but do you think it's more dangerous to return a pointer, because then the caller needs to make sure not to change any of the struct values and since this is a globally shared object this could result in weird side effects.

src/golang/cmd/server/server/handlers.go
```diff
@@ -15,7 +15,6 @@ func (s *AqServer) Handlers() map[string]handler.Handler {
 		routes.ConnectIntegrationRoute: &handler.ConnectIntegrationHandler{
 			Database:   s.Database,
 			JobManager: s.JobManager,
-			Vault:      s.Vault,
```
saurav-c, line 18: Can we remove the vault object from the server, that way we prevent any misuse in the future.

saurav-c (approved):
Looks good, left 2 comments below. We can merge once they are addressed.

src/golang/config/config.go
cw75: I see, yeah I think it make sense to be safe. I'll revert the change.

src/golang/cmd/server/server/handlers.go
cw75: Seems that we have a lot of other handlers currently making use of this `s.Vault`. Are these all bugs?

PR #944, opened 2023-02-01
Implement actual code to send slack notifications

likawind:
This PR implements codes to actually send slack notifications. The main steps are:
* creating a client using token provided
* do a linear search based on channel names
* send message to each channel

ENG-2267

## Loom demo (if any)
First part of https://www.loom.com/share/b7929818b83743d881db8d712675fdc3


src/golang/lib/notification/slack.go
```diff
@@ -26,13 +29,86 @@ func (s *SlackNotification) Level() shared.NotificationLevel {
...
+	for _, name := range names {
+		namesSet[name] = true
+	}
+
+	// Slack channel names should be unique, based on
+	// https://twitter.com/slackhq/status/965358920146608128?lang=en .
+	// However, we make this list dynamic size on purpose to prevent potential
+	// compatibility issue in future. We will still send notifications to all
+	// channels matching the given name.
+	results := []slack.Channel{}
```
kenxu95, line 54: I think we should be using `make(slack.Channel, 0, len(channels))` here instead. It's the same thing, but more canonical and makes the append more efficient.

kenxu95: I guess I'm not entirely sure what you mean by "making the list of dynamic size on purpose".

kenxu95 (changes requested):
Nice! Some small requests and then will stamp.

src/golang/lib/notification/slack.go
likawind: I thought `make()` will put a hard-limit on the capacity but it appears the size still extend if `append(` go beyonds the original capacity. So it should actually work!

kenxu95: yeah it's the same thing but generally better since you can specify the capacity for it to pre-allocate.

kenxu95 (approved):

PR #945, opened 2023-02-02
[2/2] [GH Actions] Update Pre-merge Integration Tests to use credential and config files

kenxu95:
This PR does the following:
- Updates our pre-merge integration tests to pick a python version at random, instead of iterating over them all. This mechanism will be useful in the future for our other tests.
- Refactors common server setup and upload artifact functionality into their appropriate composite actions, which makes the main yml file much cleaner.
- For the upload artifact composite action, you must specify a prefix, which allows for deduplication so that the same artifact (eg. server logs) can be uploaded at the end of a github workflow that has multiple parallel jobs.
- Also requests a beefier machine - with 8 cores we can finish these tests in 10 minutes.

ENG-2350

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -5,81 +5,32 @@ on:
   workflow_dispatch:
 
 jobs:
-  run-tests:
-    runs-on: ubuntu-latest
-    timeout-minutes: 45
-    strategy:
-      matrix:
-        # These are all the Python versions that we support.
-        python-version: ['3.7', '3.8', '3.9', '3.10']
-    
-    name: Run Fast Integration Tests with Python Version ${{ matrix.python-version }}
+  run-tests-aqueduct-engine-and-db:
+    runs-on: ubuntu-latest-8-cores
+    timeout-minutes: 20
+
+    name: Run all Integration Tests with Aqueduct Engine and DB
     steps:
       - uses: actions/checkout@v2
 
-      - name: Create the logs directory
-        run: mkdir -p logs
-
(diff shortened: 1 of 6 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -5,81 +5,32 @@ on:
...
-
-      - name: Fetch the API key
-        run: echo "API_KEY=$(aqueduct apikey)" >> $GITHUB_ENV
 
       - name: Set up the SDK Integration Tests
         working-directory: integration_tests/sdk
         run: |
-          echo "apikey: $(aqueduct apikey)" >> test-config.yml
-          echo "address: localhost:8080" >> test-config.yml
+          echo "apikey: $(aqueduct apikey)" >> test-credentials.yml
```
likawind, line 22: so for non aq data / compute testings they will use another action?

likawind (approved):

likawind:
:shipit:

.github/workflows/integration-tests.yml
kenxu95: not another action, another job in the same action.

likawind: is the job added? I only find this job which runs the default settings, that's why I'm a bit confused

kenxu95: it's not added because the plumbing isn't quite ready yet (eg. we still need to setup the data integrations on the fly)

likawind: sounds cool! Just making sure I'm understanding the context

PR #947, opened 2023-02-03
Allows user to provide filepath to BigQuery credentials when connecting via SDK

saurav-c:
Previously we only allowed users to connect to BigQuery from the SDK if they provided a string of the svc account credentials. This is not user-friendly so we should allow them the option of specifying a file, since typically these credentials are downloaded in a JSON file.

This was also a pre-req for our integration tests, since it is easier to specify a credentials file instead of copying a long JSON string in the test-credentials file.

ENG 2386

## Test
See the following notebook commands:
Screen Shot 2023-02-02 at 5 19 36 PM

### DIFF
--- a/integration_tests/no_concurrency/sdk/requirements_test.py
+++ b/integration_tests/no_concurrency/sdk/requirements_test.py
@@ -147,7 +147,6 @@ def test_default_requirements_installation(client):
 
 
 def test_requirements_invalid_arguments(client):
-
     with pytest.raises(InvalidUserArgumentException):
 
         @op(requirements=123)
--- a/sdk/aqueduct/backend/api_client.py
+++ b/sdk/aqueduct/backend/api_client.py
@@ -255,7 +255,7 @@ class APIClient:
         headers.update(
             {
                 "integration-name": name,
-                "integration-service": service,
+                "integration-service": service.value,
                 "integration-config": config.json(),
             }
         )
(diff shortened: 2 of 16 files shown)


sdk/aqueduct/backend/api_client.py
```diff
@@ -255,7 +255,7 @@ def connect_integration(
         headers.update(
             {
                 "integration-name": name,
-                "integration-service": service,
+                "integration-service": service.value,
```
saurav-c, line 258: @kenxu95 I found this was giving me an error saying that `service` was of type Enum so that couldn't be used in the headers. This ensures that the actual string value is passed along. I'm not sure why it wasn't causing errors earlier.

saurav-c:
cc @vsreekanti since there are some SDK changes involved in terms of how a user can connect to big query from the SDK

integration_tests/no_concurrency/sdk/requirements_test.py
```diff
@@ -147,7 +147,6 @@ def test_default_requirements_installation(client):
 
 
 def test_requirements_invalid_arguments(client):
-
```
saurav-c, line 150: All of these changes were from the linter

vsreekanti:
The API looks fine. Do we require an absolute path, or is a relative path okay?

saurav-c:
@vsreekanti Relative is fine

likawind (approved):
LGTM, my only question is if this path is always the local file system, rather than some S3 location if they are using different infrastructures?

saurav-c:
@likawind This is only meant to be local relative to the SDK. We are not storing this filepath in the backend, so this filepath should be a local filepath. It probably is unnecessary to support S3 paths for this feature.

PR #949, opened 2023-02-03
Adds Python pkg version upper bounds

saurav-c:
This PR adds version bounds for all Python packages when the Aqueduct engine is used. For the engines that use Docker images we already have some version bounds in place and need to do more testing before we modify them. This will at least give us some cover from any compatibility issues popping up on the Aqueduct server upgrade itself.

I generated these upper bounds by installing Aqueduct (and all connectors) on a clean instance. Then I took `pip freeze` and used those values for the upper bounds. 

ENG 2337


cw75 (commented):
@saurav-c I think we should also upper-bound the SDK requirements?

cw75 (approved):

PR #951, opened 2023-02-03
Fixes bug in handling user error with mismatch in output count

saurav-c:
This PR fixes a bug in how we handle the following types of errors:
1. There is an error in the user code, i.e. an exception is thrown.
2. The number of outputs do not match what was expected.

Previously, we were prioritizing Error #2 over #1 and escalating that to the user. The correct use case is to catch Error #1 first and if so provide the user with the appropriate stack trace.

This PR also adds a new integration test case for this specific behavior.

ENG 2333

### DIFF
--- a/integration_tests/sdk/aqueduct_tests/error_handling_test.py
+++ b/integration_tests/sdk/aqueduct_tests/error_handling_test.py
@@ -14,6 +14,11 @@ def bad_op(df):
     return df
 
 
+@aqueduct.op(num_outputs=2)
+def bad_op_multiple_outputs(df):
+    return bad_op(df)
+
+
 # These tips should match executor code so that we can verify the correct error is generated.
 TIP_OP_EXECUTION = "Error executing operator. Please refer to the stack trace for fix."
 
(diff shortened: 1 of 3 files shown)


src/python/aqueduct_executor/operators/function_executor/execute.py
```diff
@@ -290,14 +308,6 @@ def run(spec: FunctionSpec) -> None:
             result_types[0] = ArtifactType.BOOL
             results[0] = True
         else:
-            # Error if the number of expected outputs is not what was returned by the user's function.
-            if len(results) != len(spec.expected_output_artifact_types):
```
saurav-c, line 294: This error check is redundant as we already checked the counts above.

integration_tests/sdk/aqueduct_tests/error_handling_test.py
```diff
@@ -27,6 +32,15 @@ def test_handle_bad_op_error(client, data_integration):
         assert TIP_OP_EXECUTION in e.message
 
 
+def test_handle_bad_op_with_multiple_outputs(client, data_integration):
+    table_artifact = extract(data_integration, DataObject.SENTIMENT)
+
+    try:
```
kenxu95, line 38: The way to do it in tests is to use `with pytest.raises(AqueductError, match=TIP_OP_EXECUTION)`

integration_tests/sdk/aqueduct_tests/error_handling_test.py
```diff
@@ -27,6 +32,15 @@ def test_handle_bad_op_error(client, data_integration):
...
+        bad_op_multiple_outputs(table_artifact)
+    except AqueductError as e:
+        assert TIP_OP_EXECUTION in e.message
```
kenxu95, line 41: Do you mind pasting the Stack trace of the error? Just curious as to what it looks like

kenxu95 (approved):
Unblocking, but please address comments :)

PR #953, opened 2023-02-03
Allows users to customize artifact names

kenxu95:

Allows for both:
```
artf = fn()
artf.set_name('featurized data')
```

And:
```
@op(outputs=["featurized data"])
def foo():
    ...
```


vsreekanti:
API looks good! Just to double check, how'd we end up dealing with multiple outputs in the decorator?

kenxu95:
If there are multiple outputs, you either have to name all of them:
```
@op(outputs=["output1", "output2"])
def foo():
    ...
```

or none

```
@op(num_outputs=2)
def foo():
    ...
```

or you can do both, but it's unnecessary:

```
@op(num_outputs=2, outputs=["output1", "output2"])
def foo():
    ...
```

vsreekanti:
Okay, that's great!

sdk/aqueduct/decorator.py
```diff
@@ -373,15 +407,25 @@ def op(
...
+    if num_outputs is None and outputs is None:
+        num_outputs = 1
+    elif num_outputs is not None and outputs is not None:
+        if len(outputs) != num_outputs:
+            raise InvalidUserArgumentException(
+                "`len(outputs) must be equivalent to `num_outputs`. Getting %d and %d"
+                % (len(outputs), num_outputs)
+            )
+    elif num_outputs is None and outputs is not None:
+        num_outputs = len(outputs)
```
eunice-chan, line 420: Will there be any issue if len(outputs) != actual num_outputs?

kenxu95: yep, it'll fail at execution time.

sdk/aqueduct/decorator.py
```diff
@@ -373,15 +407,25 @@ def op(
...
+
+    # If not set, default number of outputs is one.
```
likawind, line 422: I'm a bit confused as after the above 3 if-elif blocks the only remaining condition is `num_output is not None` and outputs is None` ? It seems that we should do some assertion here instead?

likawind (approved):
The implementation and testings looks good! Just one minor comments.

I think the change does reflect the discussions in `product` channel, but will let Vikram confirm on API pattern.

sdk/aqueduct/decorator.py
kenxu95: Yeah the casework here is a little tricky - if a `num_output` is provided, then the `outputs` is an optional argument, hence why that case isn't specified here. Open to suggestions, but the relationship isn't symmetric

PR #957, opened 2023-02-05
Compete Athena integration test

cw75:
Note that currently we only support reading from Athena.
I setup an Athena DB with two tables from our demo db: "customers" and "hotel_reviews", and gave everyone within our AWS org permission to access. So filling in the missing entries in `test-credentials-example.yml` should give you access to that DB. These missing fields are:`output_location` (any s3 file path that you have access to, in order to store temp results generated by Athena) and your AWS credentials. Other fields like `region` and `database` should be kept as is.


integration_tests/sdk/data_integration_tests/athena_test.py
```diff
@@ -0,0 +1,48 @@
...
+def test_athena_integration_table_retrieval(client, data_integration):
+    df = data_integration.table(name="hotel_reviews")
+    assert len(df) == 100
+    assert list(df) == [
+        "hotel_name",
+        "review_date",
+        "reviewer_nationality",
+        "review",
+    ]
+
```
kenxu95, line 26: Based on a quick skim through relational_test.py, I think we're just missing the case where we parameterize sql queries here? Can you also add that?

kenxu95 (approved):
Unblocking but please address comments.

integration_tests/sdk/test-credentials-example.yml
```diff
@@ -32,3 +32,17 @@ data:
     # (Alternative) Config File Credentials
     # config_file_path:
     # config_file_profile:
+  test_athena:
+    type: Athena
+    database: test
+    region: us-west-1
+    # We should all have access to this bucket, so no need to change the value here.
+    output_location: s3://aqueduct-athena-temp/tmp/
```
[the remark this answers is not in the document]
cw75: In terms of security I think all we're exposing is the bucket name and path, but nobody outside of our org have access to it, so I'm not super concerned here.

integration_tests/sdk/data_integration_tests/athena_test.py
cw75: added

PR #972, opened 2023-02-09
Allows user to specify optional Snowflake schema in UI

saurav-c:
This PR fixes a bug where the UI did not allow users to specify a non-default schema. We already have backend support for custom schemas. This change is also backwards compatible, as previous integrations added from the UI will continue to use the default (`public`) schema.

ENG 2419

## Loom demo (if any)
https://www.loom.com/share/29d975604d9248b2bde216c6b068abe9

### DIFF
--- a/src/ui/common/src/components/integrations/dialogs/snowflakeDialog.tsx
+++ b/src/ui/common/src/components/integrations/dialogs/snowflakeDialog.tsx
@@ -1,5 +1,5 @@
 import Box from '@mui/material/Box';
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 
 import { SnowflakeConfig } from '../../../utils/integrations';
 import { readOnlyFieldDisableReason, readOnlyFieldWarning } from './constants';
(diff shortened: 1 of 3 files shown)


saurav-c:
cc @vsreekanti for UI changes

src/ui/common/src/components/integrations/dialogs/snowflakeDialog.tsx
```diff
@@ -24,6 +25,18 @@ export const SnowflakeDialog: React.FC<Props> = ({
...
   editMode,
 }) => {
+  const [schema, setSchema] = useState<string>(
+    value?.schema ?? Placeholders.schema
+  );
+
+  useEffect(() => {
+    if (schema) {
+      onUpdateField('schema', schema);
+    } else {
```
agiron123, line 35: Is it ever possible for the length of schema to be 0?

From glancing at this code, it looks like the answer is no, but just making sure.

agiron123 (approved):

src/ui/common/src/components/integrations/dialogs/snowflakeDialog.tsx
saurav-c: Nope there should always be a schema, so if its blank, we want to use the default `public` one.

PR #978, opened 2023-02-10
Fix bug where we are retrieving artifact payloads on workflow page

cw75:
In most cases, we don't need the artifact payload when we're on the workflow page. This PR adds an optional `metadata-only` header to the get artifact result route, and modifies the UI accordingly to only set this header to true if we need the artifact payload.

We only need the artifact payload when 1) we're clicking into an artifact node, and 2) we are on the workflow page and the DAG contains metric and check artifacts, in which case we need their values because we're showing them on the workflow page.

This change regresses the "Export Csv" functionality when clicking into a Table artifact node. But since we can export it by clicking "View Artifact Details" anyway, the "Export Csv" button is actually redundant and I think we can deprecate it.

See demo below:
https://www.loom.com/share/8d311ee791134208a05542fdf850f5b7

### DIFF
--- a/src/golang/cmd/server/handler/get_artifact_result.go
+++ b/src/golang/cmd/server/handler/get_artifact_result.go
@@ -166,6 +166,10 @@ func (h *GetArtifactResultHandler) Prepare(r *http.Request) (interface{}, int, e
 	}
 
 	metadataOnlyString := r.Header.Get(routes.MetadataOnlyHeader)
+	if metadataOnlyString == "" {
+		metadataOnlyString = "false"
+	}
+
 	metadataOnly, err := strconv.ParseBool(metadataOnlyString)
 	if err != nil {
 		return nil, http.StatusInternalServerError, errors.Wrap(err, "Unexpected error when converting metadata-only header to bool.")
--- a/src/golang/cmd/server/handler/get_artifact_result.go
+++ b/src/golang/cmd/server/handler/get_artifact_result.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"mime/multipart"
 	"net/http"
+	"strconv"
 
 	"github.com/aqueducthq/aqueduct/cmd/server/routes"
 	"github.com/aqueducthq/aqueduct/lib/collections/artifact"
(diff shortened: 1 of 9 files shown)


src/ui/common/src/components/pages/workflow/id/index.tsx
```diff
@@ -221,8 +216,16 @@ const WorkflowPage: React.FC<WorkflowPageProps> = ({
         );
       }
 
-      for (const artfId of [...op.inputs, ...op.outputs]) {
```
cw75, line 224: @vsreekanti @agiron123 This `...op.inputs` part confused me quite a bit. If we're looping through all operators, won't including both op's inputs _and_ outputs lead to a bunch of redundant artifact fetches??

src/ui/common/src/components/pages/workflow/id/index.tsx
```diff
@@ -319,16 +322,6 @@ const WorkflowPage: React.FC<WorkflowPageProps> = ({
     };
 
     if (currentNode.type === NodeType.TableArtifact) {
-      // Since workflow is pending, it doesn't have a result set yet.
```
vsreekanti, line 322: I think I understand, but to double check — why is this code no longer necessary?

vsreekanti (approved):
LGTM! Couple questions/comments inline.

src/golang/cmd/server/handler/get_artifact_result.go
```diff
@@ -159,10 +165,21 @@ func (h *GetArtifactResultHandler) Prepare(r *http.Request) (interface{}, int, e
 		return nil, http.StatusBadRequest, errors.Wrap(err, "The organization does not own this artifact.")
 	}
 
+	metadataOnlyString := r.Header.Get(routes.MetadataOnlyHeader)
+	if metadataOnlyString == "" {
```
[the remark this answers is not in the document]
cw75: I think it should be default false because this route is also used by the SDK, and we don't want to alter the existing behavior there.

src/ui/common/src/components/pages/workflow/id/index.tsx
cw75: This is where I'm deprecating the export csv button.

src/ui/common/src/components/pages/workflow/id/index.tsx
cw75: Yes, and I verified that after this fix, redundant fetches seem gone.

src/golang/cmd/server/handler/get_artifact_result.go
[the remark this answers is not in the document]
vsreekanti: Okay, got it!

PR #981, opened 2023-02-10
Log performance breakdown when running Python operators

cw75:
This gives us useful information about where the bottlenecks are.

### DIFF
--- a/src/dockerfiles/gpu/start-function-executor-gpu.sh
+++ b/src/dockerfiles/gpu/start-function-executor-gpu.sh
@@ -2,23 +2,23 @@
 # NOTE: Keep this in sync with the `start-function-executor.sh` in `/src/dockerfiles/function`.
 CONDA_ENV=$1
 echo $CONDA_ENV
-FUNCTION_EXTRACT_PATH=$(conda run -n $CONDA_ENV python3 -m aqueduct_executor.operators.function_executor.get_extract_path --spec "$JOB_SPEC")
+FUNCTION_EXTRACT_PATH=$(conda run --no-capture-output -n $CONDA_ENV python3 -m aqueduct_executor.operators.function_executor.get_extract_path --spec "$JOB_SPEC")
 EXIT_CODE=$?
 if [ $EXIT_CODE != "0" ]; then exit $(($EXIT_CODE)); fi
 
-conda run -n $CONDA_ENV python3 -m aqueduct_executor.operators.function_executor.extract_function --spec "$JOB_SPEC"
+conda run --no-capture-output -n $CONDA_ENV python3 -m aqueduct_executor.operators.function_executor.extract_function --spec "$JOB_SPEC"
 EXIT_CODE=$?
 if [ $EXIT_CODE != "0" ]; then exit $(($EXIT_CODE)); fi
 
 if test -f "${FUNCTION_EXTRACT_PATH}op/requirements.txt"
 then
-      conda run -n $CONDA_ENV python3 -m pip freeze >> "${FUNCTION_EXTRACT_PATH}op/local_deps.txt"
-      conda run -n $CONDA_ENV python3 -m aqueduct_executor.operators.function_executor.install_requirements --local_path="${FUNCTION_EXTRACT_PATH}op/local_deps.txt" --requirements_path="${FUNCTION_EXTRACT_PATH}op/requirements.txt" --missing_path="${FUNCTION_EXTRACT_PATH}op/missing.txt" --spec "$JOB_SPEC" --conda_env "$CONDA_ENV"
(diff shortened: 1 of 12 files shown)


saurav-c (approved):
Unblocking but can we address some of these issues, either in this PR or follow on work:
1. Add some context to these logs and make them structured. For example, add a workflow ID or operator ID as part of the logs (if this context is shared with the Python operators).
2. Make it easier for us to find these logs and group them based on a particular workflow. Right now we would have to find the log file for a specific operator and then we can read thru all these log statements.

cw75:
Screen Shot 2023-02-11 at 11 51 57 AM
@saurav-c I formatted it into a dictionary that contains its job id, stage, and latency, and colored the log so it's easier to find.

src/dockerfiles/gpu/start-function-executor-gpu.sh
```diff
@@ -2,23 +2,23 @@
 # NOTE: Keep this in sync with the `start-function-executor.sh` in `/src/dockerfiles/function`.
 CONDA_ENV=$1
 echo $CONDA_ENV
-FUNCTION_EXTRACT_PATH=$(conda run -n $CONDA_ENV python3 -m aqueduct_executor.operators.function_executor.get_extract_path --spec "$JOB_SPEC")
+FUNCTION_EXTRACT_PATH=$(conda run --no-capture-output -n $CONDA_ENV python3 -m aqueduct_executor.operators.function_executor.get_extract_path --spec "$JOB_SPEC")
```
kenxu95, line 5: What does this new flag do?

src/python/aqueduct_executor/operators/connectors/data/main.py
```diff
@@ -12,4 +16,18 @@
...
+    print_with_color(
+        "Starting %s job: %s" % (spec.type.value, spec.name), color=PrintColorType.GREEN
+    )
+    begin = time.time()
+
     execute.run(spec)
+
+    end = time.time()
+    performance = {
+        "job": spec.name,
```
kenxu95, line 28: We can file this as a follow-up task, but we should really print the operator name here (we'll have to pass it in through the job spec). The spec name is generally pretty useless :/

src/python/aqueduct_executor/operators/function_executor/execute.py
```diff
@@ -216,21 +218,49 @@ def run(spec: FunctionSpec) -> None:
...
         derived_from_bson = SerializationType.BSON_TABLE in serialization_types
-        print("Invoking the function...")
+
+        print_with_color(
+            "Running function for %s job: %s" % (spec.type.value, spec.name),
+            color=PrintColorType.GREEN,
+        )
+        begin = time.time()
         results, system_metadata = _execute_function(spec, inputs, exec_state)
+        end = time.time()
```
kenxu95, line 254: The repetitiveness of this pattern makes me think this latency logging should be a decorator. Do you think that would be a possible change we could make here? It would really clean up the code and prevent bugs.

src/python/aqueduct_executor/operators/utils/enums.py
```diff
@@ -81,3 +81,8 @@ class SerializationType(str, Enum, metaclass=MetaEnum):
     STRING = "string"
     BYTES = "bytes"
     TF_KERAS = "tensorflow-keras-model"
+
+
+class PrintColorType(str, Enum, metaclass=MetaEnum):
+    GREEN = "\33[32m"
```
kenxu95, line 87: wth is this format lol. Can you add a quick description about what this format is, so someone can recreate the colors themselves?

kenxu95 (changes requested):
This will definitely be super helpful. Left a refactoring comment to be addressed.

Also, totally agree with Saurav's points, can we make tasks for his suggestions? I think they're out of scope of this PR though.

src/python/aqueduct_executor/operators/function_executor/execute.py
kenxu95: I would imagine something simple like this would suffice: https://dev.to/kcdchennai/python-decorator-to-measure-execution-time-54hk

Of course, you would pass in the variables needed to print everything correctly.

cw75:
@kenxu95 refactored and addressed comments.

cw75:
Also created issues.

src/python/aqueduct_executor/operators/utils/utils.py
```diff
@@ -215,3 +217,35 @@ def write_compile_airflow_output(storage: Storage, path: str, dag_file: bytes) -
...
     storage.put(path, dag_file)
+
+
+def print_with_color(log: str, color: PrintColorType = PrintColorType.YELLOW) -> None:
+    assert color in PrintColorType
+    CEND = "\33[0m"
+    print(color + log + CEND)
+
+
+def timeit(
```
kenxu95, line 228: nit: can we call it `time_it` to be consistent with our other methods?

kenxu95 (approved):

PR #975, opened 2023-02-10
Eng 2341 Allow user to configure if a notification applies to all workflows

likawind:
## Describe your changes and why you are making these changes
This PR improves notification selector to allow user to disable the notification. If disabled, the notification is not applied to all workflows by default while users can still override in workflow specific settings.

As shown in demo, the new selector shows level selection only if the notification is enabled. Due to the way we handle state, it 'memorizes' the previously checked option (e.g. check succeed -> check disable -> uncheck disable, the 'succeed' will be automatically checked).

The PR is unfortunately a mix of go and react code, but go change is relatively simple, as we just add one field to notification config.

Scenarios tested:
* new integration enabled / disabled
* override a disable notification in workflow
* edit integration and update enable / disable status

## Related issue number (if any)
ENG-2341

## Loom demo (if any)
https://www.loom.com/share/6ec5783e0da1417497bff9c277c42904

Screenshot 2023-02-09 at 6 22 35 PM

## Checklist before requesting a review
- [ ] I have created a descriptive PR title. The PR title should complete the sentence "This PR...".
- [ ] I have performed a self-review of my code.
- [ ] I have included a small demo of the changes. For the UI, this would be a screenshot or a Loom video.
- [ ] If this is a new feature, I have added unit tests and integration tests.
- [ ] I have run the integration tests locally and they are passing.
- [ ] I have run the linter script locally (See `python3 scripts/run_linters.py -h` for usage).
- [ ] All features on the UI continue to work correctly.
- [ ] Added one of the following CI labels:
    - `run_integration_test`: Runs integration tests
    - `skip_integration_test`: Skips integration tests (Should be used when changes are ONLY documentation/UI)

PR #984, opened 2023-02-11
ENG-2396 Improve UI for workflow to configure default notifications

likawind:
This PR adds an additional checkbox in workflow settings page to allow configuring whether or not to use default notifications. This box has a link that opens up account page (in a separate tab) to show current notifications.

As a part of the PR, we moved notification's checkbox entry to its own component to avoid code duplication
ENG-2396

## Loom demo (if any)
https://www.loom.com/share/af9658bc77f3461f9ca3783e8ec5cd3a


vsreekanti:
Few nits:
* There's a space between "settings" and the following period. 
* Is default notification settings a link? The text look like it, but it doesn't seem to be based on where you're clicking.
* When you uncheck the default box, is what's populated there the default settings? I think it is, but I wanted to double check.

likawind:
@vsreekanti 
- "Is default notification settings a link" Yes, it opens a new tab to the account page so that user don't lost the current session. I'm recording the browser tab and that's why it didn't show up.
- "When you uncheck the default box, is what's populated there the default settings" that's a good point, for now we just show empty settings and user need to click '+' to add. It's indeed nice to just show the default settings when the user uncheck.

vsreekanti:
* I think it's maybe a little confusing that most of the text is covered by the link. Instead, let's use and link the second instance of the word settings:
    > Use default notification settings. See settings for more details.
* Let's create a backlog nice to have task for making it the default, but no need to prioritize right now.

likawind:
@vsreekanti just to confirm on your suggesting putting the link:
Use default notification settings. See (settings)[link] for more details.

vsreekanti:
Yep!

agiron123:
I'm a little confused by this UI, but this is something we can address in the future.

It seems like there are three things that are addressed by this UI:
 1. Notification Channel - i.e. slack or email, where do these go
 2. Severity level - warning, error, success should send notification.
 3. Reset selection to defaults.

The main thing that confuses me is severity level being a checkbox that doesn't let you select more than one item. Maybe we can try making it a dropdown and indent it 32px underneath the ```my_gmail``` integration? The trash can to remove is nice, can definitely keep that.

Next, use default settings is placed at the top of this section, but I'd imagine that this is something that users probably want to do last. Maybe this use default notification settings item can be a button that we show above the save button?

agiron123 (approved):

likawind:
@agiron123 thanks for feedbacks
- the dropdown still allows you to select slack / email based on the integrations you already have, is that what you are asking for?
- would you suggest to move 'use default' to the botton? In this way when the user uncheck, the '+' button and the whole section will appear before this botton.
- Which part of the level selector do you find confusing? The settings page is already content heavy so I slightly prefer not indenting, and not sure how the indentation helps clarify any information

likawind:
Btw, while we keep these product discussion open, given codewise things are looking fine, I will merge everything to #975 so that I don't have to switch between branches to address any comments

PR #988, opened 2023-02-13
[1/2] Allow for previewing an operator with custom parameters

kenxu95:
@vsreekanti to watch the demo and comment on any API changes

The point of this PR is to allow for a user more easily try out different parameters for an operator that has multiple outputs. Whereas before we could only do `artifact.get(parameters={...})`, now you can simply pass in the value of the new parameter directly, eg. `foo(123)`.

This change in itself isn't that hard, but the naming ramifications are significant. 
- An implicit parameter is named with "<op_name>:<param_name>"
- Implicit parameters can overwrite each other, as long as they are consumed by the same operator. Otherwise, we will error.

```
def foo(param):
    ....

foo(222) # This creates an input parameter named `foo:param`.
foo(333) # This overwrites the previously created `foo_param` with the new value.
```
Moreover, as I was building this, I discovered some bugs in our naming policy that are now fixed:
- We decided to use (1) to deduplicate instead of just "1". Moreover, this index suffix is now only appended
   if there is a collision.
- Default artifact names derived from operator names can now collide with other artifacts that were renamed! We used to rely on operator name uniqueness to enforce this, but since https://github.com/aqueducthq/aqueduct/pull/953, that is no longer the case.

ENG-2343
Also solves ENG-2399 as part of the naming collision fix

## Loom demo (if any)
https://www.loom.com/share/badd37d94227417db2df5aa11f9d5254 (This is out of date now. @cw75 has the context)


sdk/aqueduct/artifacts/base_artifact.py
```diff
@@ -39,12 +40,20 @@ def set_operator_type(self, operator_type: OperatorType) -> None:
...
+        # If this a parameter artifact, we will also need to change the name of the parameter,
+        # to preserve our invariant that a param op and its artifact always have the same name.
+        op = self._dag.must_get_operator(with_output_artifact_id=self._artifact_id)
+        if get_operator_type(op) == OperatorType.PARAM:
+            if self._dag.get_operator(with_name=name) is not None:
+                raise InvalidUserActionException(
+                    "Parameter names must be globally unique. There already exists an operator named `%s`."
+                    % name,
+                )
+            self._dag.update_operator_name(op.id, name)
```
kenxu95, line 54: What happens if the new parameter operator name collides with another operator?

kenxu95:
@cw75 I've updated the PR to reflect our latest discussion.

sdk/aqueduct/utils/naming.py
```diff
@@ -0,0 +1,101 @@
...
+
+    Operator collisions are resolved. Artifact collisions are not tolerated.
+
+    Args:
+        dag:
+            The current dag.
+        candidate_op_name:
+            The operator name that we want to check and potentially deduplicate.
+        overwrite_existing_op_name:
+            If set, any operator collisions will be ignored. Otherwise, we will bump
```
cw75, line 44: instead of "ignored", do you mean we overwrite the existing one?

sdk/aqueduct/utils/naming.py
```diff
@@ -0,0 +1,101 @@
...
+            the operator name until an available one is found.
+        candidate_artifact_names:
+            The custom output artifact names the caller wants to assign to the operator
+            outputs. If not set, we will use our default artifact naming scheme. Regardless,
+            if there already exist artifact's with the same name, we will error.
+        num_outputs:
+            Must be consistent with `candidate_artifact_names`, if set. The number of
+            output artifacts to expected.
+        only_resolve_op_name:
```
cw75, line 53: Just reading the comment here, since we do have uniqueness constraint on artifact names, if we skip the validation won't it cause bad behavior?

sdk/aqueduct/utils/naming.py
```diff
@@ -0,0 +1,101 @@
...
+    if isinstance(candidate_artifact_names, str):
+        candidate_artifact_names = [candidate_artifact_names]
+    assert isinstance(candidate_artifact_names, list)
+    assert num_outputs == len(candidate_artifact_names)
+
+    for candidate_artifact_name in candidate_artifact_names:
+        dag.validate_artifact_name(candidate_artifact_name)
+    artifact_names = candidate_artifact_names
+
+    return op_name, artifact_names
```
cw75, line 101: Can we just return `candidate_artifact_names`?

sdk/aqueduct/utils/naming.py
```diff
@@ -0,0 +1,101 @@
...
+    # First, resolve the candidate op name. If we can overwrite an existing op name, then
+    # don't even bother finding an unallocated op name.
+    if not overwrite_existing_op_name:
+        prefix = candidate_op_name
+        curr_suffix = 1
+        while dag.get_operator(with_name=candidate_op_name) is not None:
+            candidate_op_name = prefix + " (%d)" % curr_suffix
+            curr_suffix += 1
+
+    # If we're overwriting an existing operator, temporarily delete the operator from the dag
```
cw75, line 69: Not sure I totally understand this part. Are we expecting the caller to add the operator with the same name to the DAG later on? Also just to confirm, when we delete this operator, all of its downstream ops are also deleted right?

integration_tests/sdk/aqueduct_tests/param_test.py
```diff
@@ -81,27 +82,150 @@ def test_get_with_custom_parameter(client):
...
+        client.create_param("another_fn:another_param", default="this should fail")
+
+
+def test_change_param_artifact_name(client, flow_name, engine):
+    """Test that changing a parameter artifact name is possible."""
+    param = client.create_param("param", default=123)
+    param.set_name("new param name")
+    new_param = param  # Move the parameter to a different variable
+
+    # The output name collides with the old param name, but we already moved it out.
```
cw75, line 216: do you mean the "operator name" collides with the old param name?

cw75 (approved):
@kenxu95 LGTM! Left some minor questions and comments inline.

vsreekanti:
@kenxu95, I see you said the demo is out of date — is there a new video I should watch?

sdk/aqueduct/artifacts/create.py
```diff
@@ -1,30 +1,140 @@
...
+    candidate_name: str,
     default: Any,
     description: str = "",
+    op_name_for_implicit_param: Optional[str] = None,
 ) -> BaseArtifact:
-    """Creates a parameter operator and return an artifact that can be fed into other operators."""
+    """Creates a parameter operator and return an artifact that can be fed into other operators.
+
+    For implicitly created parameters, the naming collision policy is as follows:
+    - We will bump the parameter name until it is unique, unless the colliding operator satisfies
```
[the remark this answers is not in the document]
kenxu95: oops I forgot to update this docstring.

sdk/aqueduct/artifacts/create.py
```diff
@@ -1,30 +1,140 @@
...
+            op_name_for_implicit_param,
+        )
+        if not is_overwrite:
+            warnings.warn(
+                """Input to function argument `%s` is not an artifact type. We have implicitly created a parameter named `%s` and your input will be used as its default value. This parameter will be used when running the function."""
+                % (param_name, param_name)
+            )
+    else:
+        colliding_op = dag.get_operator(with_name=param_name)
+        if colliding_op is not None and _operator_is_implicitly_created_param(colliding_op):
```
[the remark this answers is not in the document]
kenxu95: Yeah, resolving that collision is beyond the scope of this PR.

sdk/aqueduct/utils/naming.py
kenxu95: That's what I meant by ignored, but can change the wording around a bit.

sdk/aqueduct/utils/naming.py
cw75: I think my understanding is correct after reading through the whole PR. Maybe worth adding a comment clarifying the contract here?

sdk/aqueduct/utils/naming.py
kenxu95: Yes. We only ever skip the artifact name validation in the load operator case, since there is no artifact. I can add that to the docstring.

sdk/aqueduct/utils/naming.py
kenxu95: I wanted to make it clear at what point a candidate name became the resolved name though - it makes a little more sense for the op_name assignment above.

sdk/aqueduct/utils/naming.py
cw75: That'd be great.

sdk/aqueduct/utils/naming.py
kenxu95: No, this is so that the artifact naming collision detection is accurate. If I don't delete the operator I'm colliding with, then it's artifact may collide with the candidate artifact in the second half of this method. Extending this logic, any artifact downstream of this existing operator needs to be removed before we can compute artifact name uniqueness.

cw75: I think what you're saying make sense, but still within this function, we aren't really "temporarily" deleting the operator because we're not adding it back later within this function right? So it's still up to the caller to add the candidate operator to this DAG?

sdk/aqueduct/artifacts/create.py
```diff
@@ -1,30 +1,140 @@
...
+        param_name = op_name_for_implicit_param + ":" + param_name
+        is_overwrite = _resolve_implicit_param_name(
+            dag,
+            param_name,
+            op_name_for_implicit_param,
+        )
+        if not is_overwrite:
+            warnings.warn(
+                """Input to function argument `%s` is not an artifact type. We have implicitly created a parameter named `%s` and your input will be used as its default value. This parameter will be used when running the function."""
+                % (param_name, param_name)
```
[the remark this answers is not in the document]
cw75: @kenxu95 reminder to address this as well.

PR #989, opened 2023-02-13
[2/2] Improve parameter management from Aqueduct client

kenxu95:
Adds the ability to list and delete parameters in the current working context. See the `test_param_management` test case to see how this would be used.

ENG-2343

### DIFF
--- a/integration_tests/sdk/aqueduct_tests/param_test.py
+++ b/integration_tests/sdk/aqueduct_tests/param_test.py
@@ -512,3 +512,42 @@ def test_parameter_type_changes(client, flow_name, engine):
         expected_status=ExecutionStatus.FAILED,
         parameters={"number": "This is a string"},
     )
+
+
+def test_param_management(client):
+    # Create some implicit parameters that are consumed by downstream operators.
+    @op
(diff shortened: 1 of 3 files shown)


integration_tests/sdk/aqueduct_tests/param_test.py
```diff
@@ -512,3 +512,42 @@ def noop(input):
...
+    def foo(param1, param2):
+        return 123
+
+    foo_output = foo(1000, "string val")
+
+    @op
+    def bar(input):
+        return input
+
+    bar_output = bar(foo_output)
```
cw75, line 529: Since parameters 1 and 2 are already consumed by `foo`, we don't really need `bar` here for the tests to work right?

cw75 (approved):
@kenxu95 LGTM! Left a question inline about a test case.

integration_tests/sdk/aqueduct_tests/param_test.py
kenxu95: yeah it's unnecessary. I just added it to make sure downstream operators would be removed too. I'll add that.

PR #990, opened 2023-02-13
Fix bug where we don't allow user switch engine back to Aqueduct

kenxu95:
This change adds the "Aqueduct" option to switch the engine back. The string comparison is case-insensitive. I filed https://linear.app/aqueducthq/issue/ENG-2461/error-if-the-user-tries-to-name-a-new-compute-integration-aqueduct to make this integration name unique.

ENG-2452


integration_tests/sdk/aqueduct_tests/external_compute_test.py
```diff
@@ -0,0 +1,57 @@
...
+    def noop(input):
+        return input
+
+    @op(engine=engine, requirements=[])
+    def noop_on_third_party(input):
+        return input
+
+    # Only `noop_on_third_party` is run on outside compute.
+    output = noop_on_third_party(noop(table_artifact))
+    publish_flow_test(client, output, name=flow_name(), engine=None)
```
cw75, line 28: @kenxu95 what are we actually testing here? Are we just testing to make sure nothing errored? Also why do we set `engine=None`?

cw75 (approved):
@kenxu95 Left a question inline but otherwise LGTM

integration_tests/sdk/aqueduct_tests/external_compute_test.py
kenxu95: Yeah we're just testing that a flow with multiple external computes does not error. I can add a line for checking the artifact values too.

We set engine to None so that `noop()` is executed by Aqueduct, but `noop_on_third_party()` is executed on the external compute engine.

PR #993, opened 2023-02-14
Use Spark integration on SDK

hsubbaraj-spiral:
Add Spark integration on SDK.

### DIFF
--- a/sdk/aqueduct/constants/enums.py
+++ b/sdk/aqueduct/constants/enums.py
@@ -197,18 +197,16 @@ class RuntimeType(str, Enum, metaclass=MetaEnum):
     K8S = "k8s"
     LAMBDA = "lambda"
     DATABRICKS = "databricks"
-<<<<<<< HEAD
+    SPARK = "spark"
 
 
 class NotificationLevel(Enum, metaclass=MetaEnum):
     SUCCESS = "success"
     WARNING = "warning"
     ERROR = "error"
-=======
     SPARK = "spark"
 
 
 class SparkRuntimeType(str, Enum, metaclass=MetaEnum):
     DATABRICKS = "databricks"
     SPARK = "spark"
->>>>>>> Use Spark integration on SDK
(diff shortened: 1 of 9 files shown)


sdk/aqueduct/constants/enums.py
```diff
@@ -180,9 +181,15 @@ class ExecutionMode(str, Enum, metaclass=MetaEnum):
...
+class RuntimeType(str, Enum, metaclass=MetaEnum):
     AQUEDUCT = "aqueduct"
     AIRFLOW = "airflow"
     K8S = "k8s"
     LAMBDA = "lambda"
     DATABRICKS = "databricks"
+    SPARK = "spark"
+
+
+class SparkRuntimeType(str, Enum, metaclass=MetaEnum):
```
likawind, line 193: is this the canonical way to decide if an enum is within a subset? On top of my head I would do `SPARK_RUNTIM_TYPES = [RuntimeType.SPARK, RuntimeType.DATABRICKS]` or `def requireSpark(t RuntimeType)` instead

likawind (approved):
LGTM besides some nits

sdk/aqueduct/constants/enums.py
kenxu95: We should add a unit test making sure that SparkRuntimeType is always a strict subset of RuntimeType. And another one making sure that RuntimeType is a subset of ServiceType.

sdk/aqueduct/models/config.py
```diff
@@ -52,7 +58,13 @@ class Config:
 # TODO(...): this is deprecated.
 class FlowConfig(BaseModel):
     engine: Optional[
-        Union[AirflowIntegration, K8sIntegration, LambdaIntegration, DatabricksIntegration]
+        Union[
```
kenxu95, line 61: Can you add ENG-2511 into the deprecated TODO right above?

sdk/aqueduct/constants/enums.py
kenxu95: @likawind The MetaEnum class allows us to use the `in` syntax - from what I remember reading it seemed like ok practice. I don't know if its canonical but it seems pythonic :P We should definitely add a comment here noting that it's a subset of RuntimeType though.

kenxu95 (commented):
Can you add an entry for Spark in `connect_config` too? Please check if it's possible to connect via `client.connect_integration()` :)

PR #995, opened 2023-02-14
[1/2] Add golang apache livy client

hsubbaraj-spiral:
Add Livy client that will be used by SparkJobManager to ship operators to Spark EMR.

### DIFF
--- a/src/golang/lib/livy/client_test.go
+++ b/src/golang/lib/livy/client_test.go
@@ -1,10 +1,10 @@
 package livy
 
 import (
-	"encoding/json"
 	"net/http"
 	"net/http/httptest"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 )
@@ -31,14 +31,10 @@ func TestCreateSession(t *testing.T) {
 	defer cleanup()
 
 	mux.HandleFunc("/sessions", func(w http.ResponseWriter, r *http.Request) {
-		respSession := &Session{
-			ID:    1,
-			State: Idle,
-		}
-		resp, _ := json.Marshal(respSession)
+		jsonResp := `{"id": 1, "state": "idle"}`
 		w.Header().Set("Content-Type", "application/json")
 		w.WriteHeader(http.StatusCreated)
-		w.Write(resp)
+		w.Write([]byte(jsonResp))
 	})
 
 	// Call CreateSession with expected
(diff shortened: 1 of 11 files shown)


src/golang/lib/livy/client_test.go
```diff
@@ -0,0 +1,173 @@
...
+	mux.HandleFunc("/sessions", func(w http.ResponseWriter, r *http.Request) {
+		jsonResp := `{"id": 1, "state": "idle"}`
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusCreated)
+		w.Write([]byte(jsonResp))
+	})
+
+	// Call CreateSession with expected
+	sessionReq := &CreateSessionRequest{
+		Kind: "pyspark",
```
likawind, line 42: does these test require an active spark cluster?

src/golang/lib/livy/entrypoints.go
```diff
@@ -0,0 +1,75 @@
...
+
+const (
+	InstallAqueductEntrypoint = `import sys
+import subprocess
+
+print(1)
+subprocess.check_call([sys.executable, '-m', 'pip', 'install', 'aqueduct-ml==0.2.1'])
+`
+
+	FunctionEntrypoint = `import base64
```
likawind, line 11: How are these entry points being used?

src/golang/lib/livy/types.go
```diff
@@ -0,0 +1,124 @@
...
+	Available      StatementState = "available"
+	StatementError StatementState = "error"
+	Cancelling     StatementState = "cancelling"
+
+	OK    StatementOutputStatus = "ok"
+	Error StatementOutputStatus = "error"
+)
+
+// Livy Session.
+type Session struct {
```
likawind, line 31: Where are these statements come from? Looks like there's some public reference on how to define those?

likawind (approved):
Let's get the functional part working as long as it didn't break main. I think we can merge this change, but I'm curious what's the difference between our own library vs the go API: https://pkg.go.dev/github.com/bilcus/livy-api-go-client .

src/golang/lib/livy/entrypoints.go
hsubbaraj-spiral: These entrypoints are the base code that are being shipped to Spark. We inject the spec into the entrypoint before shipping.

src/golang/lib/livy/types.go
hsubbaraj-spiral: https://livy.incubator.apache.org/docs/latest/rest-api.html

likawind: I see, let's ship this version and get the feature done first. I do feel we can separately explore the livy go package mentioned above, as it appear to have everything implemented

src/golang/lib/livy/client_test.go
hsubbaraj-spiral: No, these use the `httptest` package to mock the server.

src/golang/lib/livy/entrypoints.go
```diff
@@ -0,0 +1,65 @@
+package livy
```
saurav-c, line 1: Do you see a future refactor where we might put this python code in our python codebase so it's easier to maintain?

hsubbaraj-spiral: Yep, I have to investigate the best way to do this. We want to ship this code as a string and need it available from the server binary. Not sure if the server binary has access to aqueduct-ml package.

Issue #1003, opened 2023-02-15
[v0.2.2] Redefining multiple parameters fails


kenxu95:
**Description**
Running an operator with multiple input parameters twice appears to fail.

**Steps to reproduce the behavior**
In my jupyter notebook, this happens if I run the following block of code twice:

```
@aq.op
def test(p1, p2):
    return p1 + p2

test(1, 2).get()
```

I hit `AssertionError: Implicit parameters can only be consumed by a single operator.`

**Expected behavior**
No error.

**Screenshots/Logs**

**Configuration Details**
 - OS: MacOS
 - Browser: N/A
 - Aqueduct Version 0.2.2

PR #1004, opened 2023-02-15
Fix bug where rerunning a function with multiple implicit parameters fails

kenxu95:
This bug came about because we processed each implicit parameter one at a time, performing validation and then creating the parameter. However, the act of creating the parameter would alter the validation step for the remaining implicit parameters, causing the assertion error described in the task.

The solution is to perform the validation for all implicit parameters at once, and only afterwards to alter the state of the dag.

ENG-2481
https://github.com/aqueducthq/aqueduct/issues/1003

### DIFF
--- a/integration_tests/sdk/aqueduct_tests/param_test.py
+++ b/integration_tests/sdk/aqueduct_tests/param_test.py
@@ -150,7 +150,7 @@ def test_implicitly_created_param_overwrites(client, flow_name, engine):
     assert flow_run.artifact("different_fn:param").get() == "different value"
 
 
-def test_multiple_implicitly_created_param_multiple(client):
+def test_multiple_implicitly_created_param(client):
     @op
     def foo(param1, param2):
         return param1 + param2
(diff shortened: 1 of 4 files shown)


vsreekanti (approved):
@kenxu95, this mostly looks good to me. Just to make sure I understand, this behavior with `op:param` is new, right? Are there any implications of that?

Otherwise, seems fairly straightforward.

integration_tests/sdk/aqueduct_tests/param_test.py
```diff
@@ -150,7 +150,16 @@ def different_fn(param):
     assert flow_run.artifact("different_fn:param").get() == "different value"
 
 
-def test_implicit_created_param_failures(client):
+def test_multiple_implicitly_created_param_multiple(client):
```
cw75, line 153: duplicate "multiple"?

cw75 (commented):
@kenxu95 I'm not sure I understand why "the act of creating the parameter would alter the validation step for the remaining implicit parameters". The other implicit parameters will have different names right? So why does it matter?

kenxu95:
@cw75 because we use the operator name that consumes the parameter as the mechanism for deciding whether to overwrite (because we're replacing the same operator) or error. If you call `create_param_artifact()` on the first parameter in the overwriting case, then the operator will be removed from the DAG. Later, when you process the second parameter, it will not see that operator in the dag, which means some assumptions were broken (hence why we ran into the assertions).

cw75 (approved):

PR #1008, opened 2023-02-16
ENG-2084 Add details to settings page about what storage is being used for metadata

agiron123:
This PR adds backend route to fetch global config from server and UI code to diplay on settings page.

ENG-2084

## Loom demo (if any)

File Metadata Storage Information:
image

GCS Metadata Storage Information:
image

S3 Metadata Storage Information:
image

### DIFF
--- a/src/golang/cmd/server/handler/get_config.go
+++ b/src/golang/cmd/server/handler/get_config.go
@@ -44,7 +44,6 @@ func (h *GetConfigHandler) Perform(ctx context.Context, interfaceArgs interface{
 	storageConfig := config.Storage()
 	storageConfigPtr := &storageConfig
 	storageConfigPublic, err := storageConfigPtr.ToPublic()
-
 	if err != nil {
 		return nil, http.StatusInternalServerError, errors.Wrap(err, "Unable to retrieve storage config.")
 	}
--- a/src/golang/cmd/server/handler/get_config.go
+++ b/src/golang/cmd/server/handler/get_config.go
@@ -1,0 +1,26 @@
+package handler
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/aqueducthq/aqueduct/config"
+)
+
+type GetConfigHandler struct {
+	GetHandler
+}
+
+func (*GetConfigHandler) Name() string {
+	return "GetConfig"
+}
+
+func (h *GetConfigHandler) Prepare(r *http.Request) (interface{}, int, error) {
+	globalConfig := config.GetGlobalConfig()
+	return globalConfig, http.StatusOK, nil
+}
+
+func (h *GetConfigHandler) Perform(ctx context.Context, interfaceArgs interface{}) (interface{}, int, error) {
+	globalConfig := config.GetGlobalConfig()
+	return globalConfig, http.StatusOK, nil
+}
--- a/src/golang/cmd/server/handler/get_config.go
+++ b/src/golang/cmd/server/handler/get_config.go
@@ -5,7 +5,21 @@ import (
 	"net/http"
 
 	"github.com/aqueducthq/aqueduct/config"
+	aq_context "github.com/aqueducthq/aqueduct/lib/context"
+	"github.com/aqueducthq/aqueduct/lib/models/shared"
+	"github.com/dropbox/godropbox/errors"
 )
+
+type getConfigArgs struct {
+	*aq_context.AqContext
+}
+
+type getConfigResponse struct {
+	AqPath              string                     `json:"aqPath"`
+	RetentionJobPeriod  string                     `json:"retentionJobPeriod"`
+	ApiKey              string                     `json:"apiKey"`
+	StorageConfigPublic shared.StorageConfigPublic `json:"storageConfig"`
+}
 
 type GetConfigHandler struct {
 	GetHandler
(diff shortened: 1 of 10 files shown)


src/golang/cmd/server/handler/get_config.go
```diff
@@ -0,0 +1,26 @@
...
+
+type GetConfigHandler struct {
+	GetHandler
+}
+
+func (*GetConfigHandler) Name() string {
+	return "GetConfig"
+}
+
+func (h *GetConfigHandler) Prepare(r *http.Request) (interface{}, int, error) {
```
agiron123, line 18: This is basically a no-op. I'm not sure what's the best way to skip or just ignore what comes from here in golang. Or is the better way to do things to do the work of getting the config here and then pass as a parameter into Perform?

src/golang/config/config.go
```diff
@@ -17,11 +17,15 @@ var (
 )
 
 type serverConfiguration struct {
-	AqPath             string                `yaml:"aqPath"`
-	EncryptionKey      string                `yaml:"encryptionKey"`
-	RetentionJobPeriod string                `yaml:"retentionJobPeriod"`
-	ApiKey             string                `yaml:"apiKey"`
-	StorageConfig      *shared.StorageConfig `yaml:"storageConfig"`
+	AqPath             string                `yaml:"aqPath" json:"aqPath"`
+	EncryptionKey      string                `yaml:"encryptionKey" json:"encryptionKey"`
```
agiron123, line 21: I've added json keys here, but realize that we probably don't want to expose the encryption key. I'm probably going to make a struct in my handler to contain all fields except for the encryptionKey.

src/ui/common/src/components/pages/AccountPage.tsx
```diff
@@ -32,6 +32,55 @@ type AccountPageProps = {
   Layout?: React.FC<LayoutProps>;
 };
 
+/*
```
agiron123, line 35: Remove comment.

Need to check storageConfigType and adjust to gcs or s3 accordingly. They have different parameters like buckets and regions. Coming up in a future commit!

src/ui/common/src/components/pages/AccountPage.tsx
```diff
@@ -103,6 +155,17 @@ client = aqueduct.Client(
   const [showNotificationUpdateSnackbar, setShowNotificationUpdateSnackbar] =
     useState(false);
 
+  useEffect(() => {
```
agiron123, line 158: This could be better served as a redux reducer. Just putting it here for now as I don't know other places in our UI where it will be used at the moment.

src/ui/common/src/components/pages/AccountPage.tsx
```diff
@@ -185,6 +248,14 @@ client = aqueduct.Client(
       )}
       {notificationSection}
 
+      <Box>
+        <Typography variant="h5" sx={{ mt: 3 }}>
+          Metadata Storage
+        </Typography>
+        <Typography variant="body1">Storage Config Type: {serverConfig?.storageConfig?.type || 'loading ...'}</Typography>
```
agiron123, line 255: Add checks for GCS and s3 storage.

src/golang/cmd/server/handler/get_config.go
```diff
@@ -0,0 +1,26 @@
+package handler
```
agiron123, line 1: TODO: Make a new struct here and omit the encryptionKey

agiron123: done. Thanks for your help here @saurav-c ❤️

src/golang/config/config.go
agiron123: Going to remove these json tags, no longer needed here.

src/ui/common/src/components/pages/AccountPage.tsx
agiron123: Redux reducer code is coming in follow up PR, #1009

src/ui/common/src/components/workflows/WorkflowSettings.tsx
```diff
@@ -261,6 +261,12 @@ const WorkflowSettings: React.FC<WorkflowSettingsProps> = ({
 
   const dispatch: AppDispatch = useDispatch();
 
+  console.log('workflowDag: ', workflowDag);
```
agiron123, line 264: remove logs, cleanup stuff :)

vsreekanti:
Nice work, @agiron123! The information here looks great, but I think the presentation warrants a little bit of cleanup. In particular, it doesn't look great to show lowercase, unformatted names for things like "gcs" or "s3". 

Here's a few things that should be pretty quick but will help quite a bit: 
1. Let's use the logos we already have on the integrations page for each of the storage mechanisms. I know we don't have one for local file storage, so we can use the Aqueduct logo as the default for now. 
2. Let's update the layout to have a little bit more differentiation between the various fields. Here's a suggestion:
```
[logo] <h3>{Storage Type}</h3>
[logo] <small-grey-text>Bucket:</> {Bucket Name} 
[logo] {region && <small-grey-text>Region:</> {region}}
```
In case it's confusing, I put logo on multiple lines just because I imagined it would take the full height of the container.

saurav-c (approved):
Go changes LGTM

agiron123:
> Nice work, @agiron123! The information here looks great, but I think the presentation warrants a little bit of cleanup. In particular, it doesn't look great to show lowercase, unformatted names for things like "gcs" or "s3".
> 
> Here's a few things that should be pretty quick but will help quite a bit:
> 
> 1. Let's use the logos we already have on the integrations page for each of the storage mechanisms. I know we don't have one for local file storage, so we can use the Aqueduct logo as the default for now.
> 2. Let's update the layout to have a little bit more differentiation between the various fields. Here's a suggestion:
> 
> ```
> [logo] <h3>{Storage Type}</h3>
> [logo] <small-grey-text>Bucket:</> {Bucket Name} 
> [logo] {region && <small-grey-text>Region:</> {region}}
> ```
> 
> In case it's confusing, I put logo on multiple lines just because I imagined it would take the full height of the container.
@vsreekanti 
Updated Screenshot:
image

vsreekanti:
Yep, this is great — thanks! 

A few minor nits on top of this:
1. Let's make the storage type bigger than the rest of the text rather than smaller.
2. Maybe add a bit more spacing under the "Metadata Storage" header.
3. Let's use the proper title of the storage engine (S3 rather than s3, Google Cloud Storage rather than gcs).
4. Using the same techniques we used on the workflow table views, let's differentiate the Bucket/Region keys from their values.

Great work!

agiron123:
> Yep, this is great — thanks!
> 
> A few minor nits on top of this:
> 
> 1. Let's make the storage type bigger than the rest of the text rather than smaller.
> 2. Maybe add a bit more spacing under the "Metadata Storage" header.
> 3. Let's use the proper title of the storage engine (S3 rather than s3, Google Cloud Storage rather than gcs).
> 4. Using the same techniques we used on the workflow table views, let's differentiate the Bucket/Region keys from their values.
> 
> Great work!

@vsreekanti I've added a follow on task to address here:
https://linear.app/aqueducthq/issue/ENG-2506/storage-metadata-design-feedback

PR #1018, opened 2023-02-22
Add ability for integration test suite to setup external storage layer before running

kenxu95:
A user can cause the server to perform a migration before running the tests, you just have to enter the data integration under `storage:` in `test-config.yml`. Note that, for simplicity, if the integration already exists, we will ignore this migration. This task was only meant to unlock these tests in S3.

Additional things this PR does:
- Adds `delete_integration()` to the SDK. I wrote this when I was being a little more ambitious with the automated storage migration management, but that feature wasn't quite mature enough to easy use - so I scrapped it in favor of just doing the migration only if the integration doesn't exist before. This API still seems useful though? So I left it. @vsreekanti 
- Adds a naming check on the backend, so you can't create multiple integrations with the same name through the SDK. Both the SDK and UI have their own checks against this, but I think there is a race condition with the asynch storage migration code that allows this. At least, this provides additional defense.

Testing:
Ran the tests locally from a clean server and verified that the S3 migration was successful.

ENG-2347


saurav-c (approved):

saurav-c:
Reminder to check failing SDK unit tests

kenxu95:
REMINDER to self, set this new job as a blocker in the branch protection rules!

PR #1027, opened 2023-02-27
Allows users to reuse operator in a workflow

saurav-c:
This PR allows users to reuse an operator in a workflow, as long as the input artifacts provided to the operator are different from previous instances of the operator. If a user attempts to add a new operator that collides with an existing operator and it has the exact same input artifacts in the same order, we simply overwrite the original operator.

ENG 2462

### DIFF
--- a/integration_tests/sdk/aqueduct_tests/naming_test.py
+++ b/integration_tests/sdk/aqueduct_tests/naming_test.py
@@ -117,13 +117,3 @@ def test_naming_collision_with_different_types(client, data_integration):
     _ = dummy_sentiment_model(table_artifact)
     with pytest.raises(InvalidUserActionException):
         _ = extract(data_integration, DataObject.SENTIMENT, op_name="dummy_sentiment_model")
-
-
-def test_naming_collision_with_dependency(client, data_integration):
-    # Overwrite is invalid when the operator being replaced is an upstream dependency.
-    table_artifact = extract(data_integration, DataObject.SENTIMENT, op_name="sentiment_model")
-    dummy_model_output = dummy_model(table_artifact)
-    dummy_model_2_output = dummy_model_2(dummy_model_output)
-
-    with pytest.raises(InvalidUserActionException):
-        _ = dummy_model(dummy_model_2_output)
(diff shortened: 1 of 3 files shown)


integration_tests/sdk/aqueduct_tests/naming_test.py
```diff
@@ -113,22 +113,7 @@ def test_naming_collision_with_different_types(client, data_integration):
     # An overwrite is invalid because the operators are of different types.
     table_artifact = extract(data_integration, DataObject.SENTIMENT, op_name="sql query")
 
-    # Function collides with existing sql artifact
-    _ = extract(data_integration, DataObject.SENTIMENT, op_name="dummy_model")
```
saurav-c, line 117: Removing this because the following is now allowed:
```
wines = demodb.sql("select * from wine;", name="wine_foo")

@op
def wine_foo(df):
    return df

wines_foo_result = wine_foo(wines)
```

integration_tests/sdk/aqueduct_tests/naming_test.py
```diff
@@ -113,22 +113,7 @@ def test_naming_collision_with_different_types(client, data_integration):
...
-    with pytest.raises(InvalidUserActionException):
-        dummy_model(table_artifact)
-
     # SQL collides with existing function
     _ = dummy_sentiment_model(table_artifact)
     with pytest.raises(InvalidUserActionException):
         _ = extract(data_integration, DataObject.SENTIMENT, op_name="dummy_sentiment_model")
-
-
-def test_naming_collision_with_dependency(client, data_integration):
```
saurav-c, line 127: Removing this because the following is now allowed:

```
wines = demodb.sql("select * from wine;")

@op
def f1(df):
    return df

@op
def f2(df):
    return df

a = f1(wines)
b = f2(a)
c = f1(b)
```

kenxu95: Can you also add this as a test case and validate the artifact names?

kenxu95 (approved):
The change looks good to me, just a couple test requests.

Also, since this task doesn't really address eng-2462, we should probably change that task or change the branch name to Linear doesn't auto-close it.

PR #1032, opened 2023-02-28
Fix requirements issues on fresh aqueduct install

likawind:

This PR fixes a number of requirement issues when running a fresh aqueduct install:
* `module 'lib' has no attribute 'X509_V_FLAG_CB_ISSUER_CHECK'`: This happens when importing `requirements` module. As a result, we put tighter constraint on this module, esp. the lowerbound.
* `pkg_resources.UnknownExtra: jsonschema 3.2.0 has no such extra feature 'format-nongpl'`: This happens when client trying to fetch the current SDK version. I believe it's caused by the `require('aqueduct-sdk')` call is trying to fetch the entire dependency tree and hit a mismatch. This mismatch shouldn't affect our behavior since it's an warning during `pip install`. As a result, we simply change `require()` to `get_distribution()` which get us the package version without doing a dependency traverse.

## Loom demo (if any)
Set up a new machine, install local version, and verify things are working using manual QA (the same steps will fail on current master, which proves the fix works.)


sdk/requirements/python-3-7.txt
```diff
@@ -1,7 +1,7 @@
 setuptools<=67.2.0
 cloudpickle<=2.2.0
 numpy<=1.24.0
-requests<=2.28.2
+requests>=2.28.1,<=2.28.2
```
vsreekanti, line 4: Do we have any concerns about restricting the version to a very narrow range? @cw75 @kenxu95

kenxu95: I feel like it's ok as a very temporary solution but we should definitely look into resolving this. What I don't understand is why these bounds are here in the first place? @likawind can you add some documentation for what packages `requests` collides with when we stray outside these bounds (in both upper and lower directions)? I'm not sure I'm seeing it from your PR description.

cw75: @kenxu95 We added the upper-bounds in case the latest versions have some backward-incompatible changes. We should periodically test and bump these upper-bounds, so that they are not lagging behind a major release.

kenxu95: ok looks like the upper bound is the latest version. Can we still document the conflict for the lower bound?

kenxu95 (approved):

likawind:
@kenxu95 @cw75 after more research I noticed this `module 'lib' has no attribute 'X509_V_FLAG_CB_ISSUER_CHECK'` issue is more tricky than I expected and simply pinning the `requests` package may not resolve it. So given there are broader concerns on limiting the version, I'm going to remove the changes over `requirements` module for now until we have a clear understanding of the problem.

likawind:
Linear: https://linear.app/aqueducthq/issue/ENG-2542/investigate-a-dependency-issue-related-to-pyopenssl

PR #1035, opened 2023-03-01
Allows users to optionally specify snowflake role

saurav-c:
This PR allows users to specify a role when using a Snowflake role. Currently all operations for Snowflake would just assume the default role of the user.

ENG 2490

## Loom demo (if any)
https://www.loom.com/share/49cdc9aefc9f4800b35b21ffec36d073

- [ ] I have included a small demo of the changes. For the UI, this would be a screenshot or a Loom video. **TODO**

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -17,7 +17,6 @@ jobs:
     name: All Integration Tests with Basic Config
     steps:
       - uses: actions/checkout@v2
-        ref: eng-2490-allow-users-to-optionally-specify-role-snowflake
 
       - uses: ./.github/actions/setup-server
         timeout-minutes: 5
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -17,6 +17,7 @@ jobs:
     name: All Integration Tests with Basic Config
     steps:
       - uses: actions/checkout@v2
+        ref: eng-2490-allow-users-to-optionally-specify-role-snowflake
 
       - uses: ./.github/actions/setup-server
         timeout-minutes: 5
--- a/src/ui/common/src/components/integrations/dialogs/dialog.tsx
+++ b/src/ui/common/src/components/integrations/dialogs/dialog.tsx
@@ -78,7 +78,7 @@ import {
   SlackDefaultsOnCreate,
   SlackDialog,
 } from './slackDialog';
-import { SnowflakeDialog } from './snowflakeDialog';
+import { isSnowflakeConfigComplete, SnowflakeDialog } from './snowflakeDialog';
 import { isSparkConfigComplete, SparkDialog } from './sparkDialog';
 import { SQLiteDialog } from './sqliteDialog';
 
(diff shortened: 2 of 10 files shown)


src/ui/common/src/components/integrations/dialogs/dialog.tsx
```diff
@@ -482,23 +482,24 @@ export function isConfigComplete(
...
       return isAthenaConfigComplete(config as AthenaConfig);
-    case 'Kubernetes':
-      return isK8sConfigComplete(config as KubernetesConfig);
     case 'Conda':
       return true;
     case 'Databricks':
       return isDatabricksConfigComplete(config as DatabricksConfig);
     case 'Email':
       return isEmailConfigComplete(config as EmailConfig);
+    case 'Kubernetes':
```
saurav-c, line 493: Just a little alphabetical sorting here

agiron123: Thanks for doing this. I wonder if there is a linter rule that we can apply here, I'll see if I can find something like that.

Perhaps something like this could benefit us in the future:
https://palantir.github.io/tslint/rules/member-ordering/

src/ui/common/src/components/integrations/dialogs/dialog.tsx
```diff
@@ -482,23 +482,24 @@ export function isConfigComplete(
...
+      return isK8sConfigComplete(config as KubernetesConfig);
+    case 'S3':
+      return isS3ConfigComplete(config as S3Config);
     case 'Slack':
       return isSlackConfigComplete(config as SlackConfig);
     case 'Spark':
       return isSparkConfigComplete(config as SparkConfig);
-
+    case 'Snowflake':
```
agiron123, line 501: As we keep adding integrations, this switch statement is going to keep growing. This actually seems like a great use case for a TypeScript class. 

I know it's a bit different than the other code that we have in our frontend codebase, but it would be nice to have a base integration class that each of these Config objects implement and provide their own 'isConfigComplete' function implementations.

This doesn't have to happen now of course, but just some food for thought when we get to hundreds of integrations :)

saurav-c:
cc @vsreekanti for UI (and SDK) changes, please take a look at the demo

src/ui/common/src/components/integrations/dialogs/snowflakeDialog.tsx
```diff
@@ -113,6 +114,26 @@ export const SnowflakeDialog: React.FC<Props> = ({
...
+        description="The role to use when accessing the database above."
+        placeholder={Placeholders.role}
+        onChange={(event) => onUpdateField('role', event.target.value)}
+        value={value?.role ?? null}
+      />
     </Box>
   );
 };
+
+export function isSnowflakeConfigComplete(config: SnowflakeConfig): boolean {
```
agiron123, line 131: This doesn't validate the form correctly unfortunately. See this screenshot, I'm able to get a blue confirm button without entering a username field:

image

Instead, what I might recommend here to get this PR out the door is to check the length of each required input and make sure that it's > 0.

In the future, we should look into using a library like react-hooks-form or rolling our own Form component to do validation.

agiron123: I could even trigger it like this:
image

agiron123 (changes requested):
Left some comments regarding form validation. Once that's fixed up you can merge.

.github/workflows/integration-tests.yml
```diff
@@ -11,6 +11,7 @@ jobs:
     name: All Integration Tests with Basic Config
     steps:
       - uses: actions/checkout@v2
+        ref: eng-2490-allow-users-to-optionally-specify-role-snowflake
```
agiron123, line 14: Is this just for your testing of this branch?

src/ui/common/src/components/integrations/dialogs/snowflakeDialog.tsx
saurav-c: Fixed

src/ui/common/src/components/integrations/dialogs/dialog.tsx
likawind: @agiron123 that's a good idea, could you create a linear task for this?

likawind (approved):
The code looks good. How hard it is to add an integration test for this?

saurav-c:
@likawind This isn't something we need to modify the integration tests for. If you simply set a role field in the `test-credentials` file it will use a role.

src/ui/common/src/components/integrations/dialogs/dialog.tsx
agiron123: Sure thing @likawind, you can find it here:
https://linear.app/aqueducthq/issue/ENG-2556/integrations-page-refactor

src/ui/common/src/components/integrations/dialogs/snowflakeDialog.tsx
```diff
@@ -113,6 +114,27 @@ export const SnowflakeDialog: React.FC<Props> = ({
...
       />
+
+      <IntegrationTextInputField
+        spellCheck={false}
+        required={false}
+        label="Role"
+        description="The role to use when accessing the database above."
+        placeholder={Placeholders.role}
+        onChange={(event) => onUpdateField('role', event.target.value)}
+        value={value?.role ?? null}
```
agiron123, line 125: I think you can do just ```value?.role```, unless that returns undefined, I'm actually not 100% sure what value we're getting here.

agiron123 (approved):
:shipit:

PR #1040, opened 2023-03-02
Make dynamic k8s cluster resilient against server failures and race conditions

cw75:
This PR mainly implements `ResyncClusterState`, which ensures the "eventual consistency" between the database k8s cluster state and the terraform state. Due to server failures and race conditions, the database state and the terraform state may temporarily diverge. For example, if you kill the server in the middle of `terraform apply`, the database state will be `Creating` but the terraform action has been killed (and the EKS cluster may be in a broken state). `ResyncClusterState` detects this kind of divergence and resyncs the state by deleting the EKS cluster and updating the database state to `Terminated` (essentially resetting the state to ground zero).

This divergence will eventually be discovered by either a user request or the dynamic teardown job, and the system fixes this inconsistency by making a call to `ResyncClusterState`.

I tested various server failure and race condition scenarios and verified that the solution is robust, with one exception described [here](https://linear.app/aqueducthq/issue/ENG-2527/make-dynamic-integration-resilient-against-race-condition-and-server), which I will put in the documentation.

In addition, the PR makes the following minor changes:
1. move `terraform init` to cloud integration registration phase, as this only needs to happen once.
2. increase default timeout to 30min to account for cluster creation time.
3. refactor the `create` action implementation of `modify_dynamic_engine.go` to reuse lower-level helper function.


src/golang/lib/dynamic/engine.go
```diff
@@ -21,7 +22,9 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-var terraformDir = filepath.Join(os.Getenv("HOME"), ".aqueduct", "server", "dynamic", "terraform")
+const stateLockErr = "Error acquiring the state lock"
```
kenxu95, line 25: nit: `ErrMsg`

kenxu95 (commented):
I think it makes sense to me - left a question though.

cw75:
@kenxu95 addressed comments based on our Slack conversation.

src/golang/cmd/executor/executor/dynamic_teardown.go
```diff
@@ -41,6 +41,11 @@ func (ex *DynamicTeardownExecutor) Run(ctx context.Context) error {
 	}
 
 	dynamicIntegration := dynamicIntegrations[0]
+
+	if err := dynamic.ResyncClusterState(ctx, &dynamicIntegration, ex.IntegrationRepo, ex.Database); err != nil {
```
likawind, line 45: Should all DB access be in a txn to avoid concurrency issue, if this method can be called concurrently?

likawind (commented):
LGTM. Left a question and will let @kenxu95 do the final call.

cw75:
@likawind in fact I explicitly made them to be not within a transaction. The reason is that intermediate state like "Creating" and "Terminating" is useful both for users and for ourselves. So the approach here is to allow temporary inconsistencies and recover later on.

src/golang/lib/engine/engine.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 const (
-	DefaultExecutionTimeout     = 15 * time.Minute
+	DefaultExecutionTimeout     = 30 * time.Minute
```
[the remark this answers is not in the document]
kenxu95: Can we get rid of this change then?

kenxu95 (approved):
Unblocking but wondering about the default timeout.

cw75:
@kenxu95 I think the timeout change is still relevant, because the original 15min is too short if an operator execution involves spinning up an EKS cluster.

PR #1043, opened 2023-03-02
[1/2] Allow user to register workflow without triggering

likawind:
This PR adds a `run_now` flag to `publish_workflow` API to allow user to register the workflow without trigging it. Also include the proper backend code and integration tests.

ENG-2497

## Loom demo (if any)
Integration test should be sufficient


src/golang/cmd/server/handler/register_workflow.go
```diff
@@ -260,15 +269,17 @@ func (h *RegisterWorkflowHandler) Perform(ctx context.Context, interfaceArgs int
...
-	_, err = h.Engine.TriggerWorkflow(
-		ctx,
-		workflowId,
-		shared_utils.AppendPrefix(dbWorkflowDag.Metadata.ID.String()),
-		timeConfig,
-		nil, /* parameters */
-	)
-	if err != nil {
-		return emptyResp, http.StatusInternalServerError, errors.Wrap(err, "Unable to trigger workflow.")
+	if args.runNow {
```
likawind, line 272: @saurav-c confirming this change should automatically apply to airflow?

sdk/aqueduct/backend/api_client.py
```diff
@@ -339,8 +341,11 @@ def register_airflow_workflow(
     def _construct_register_workflow_request(
         self,
         dag: DAG,
+        run_now: bool,
     ) -> Tuple[Dict[str, str], Dict[str, str], Dict[str, IO[Any]]]:
         headers = self._generate_auth_headers()
+        # This header value will be string "True" or "False"
+        headers.update({"run-now": str(run_now)})
```
saurav-c, line 348: Is there anything preventing us from turning this `headers` map into `Dict[str, Any]` instead of the `Dict[str, str]` that is is now. This would allow us to avoid having to do manual bool--> string conversion when using bools in headers.

saurav-c: nit: can we add a constant that defines this header so it is easy to refer to in the future

src/golang/cmd/server/handler/register_workflow.go
saurav-c: Actually this route only registers non-Airflow workflows. Airflow workflows go to the handler defined in `register_airflow_workflow.go`. 

However, since Airflow requires the user to copy over the DAG file, there is no notion of `runNow` for it.

saurav-c (approved):
Left a few minor comments, otherwise LGTM

sdk/aqueduct/client.py
```diff
@@ -479,6 +479,7 @@ def publish_flow(
         k_latest_runs: Optional[int] = None,
         config: Optional[FlowConfig] = None,
         source_flow: Optional[Union[Flow, str, uuid.UUID]] = None,
+        run_now: bool = True,
```
saurav-c, line 482: nit: Can you add a comment below documenting how this should be used by users since this is the doc string that they will access.

sdk/aqueduct/backend/api_client.py
likawind: I think http headers has to be string since the protocol doesn't differentiate on their types. From API perspective, `r.Header.Get()` returns a string value, so it's always required to do an explicit type conversion here

saurav-c: I see I'd rather we use `flag, err := strconv.ParseBool(flagStr)` in that case for parsing string to bool instead of explicitly checking a string match that has to match the upper or lower case exactly.

kenxu95 (changes requested):
I agree with Saurav's comments but want to take another look after revision.

The one thing that sticks out to me is the `run_now` terminology. I think I would prefer a less term like `trigger_immediately` that is less temporal-related and uses our existing "trigger" vocab. What do ya'll think?

likawind:
cc @vsreekanti ^

kenxu95 (commented):
Wait actually what does a user see on the UI when the workflow has been published but hasn't run? Not convinced an integration test is sufficient in this case.

integration_tests/sdk/aqueduct_tests/flow_test.py
```diff
@@ -318,6 +318,22 @@ def test_refresh_flow(client, flow_name, data_integration, engine):
...
+
+    output = foo()
+    flow = client.publish_flow(
+        name=flow_name(),
+        artifacts=output,
+        engine=engine,
+        run_now=False,
+    )
+
+    assert len(flow.list_runs()) == 0
```
kenxu95, line 334: What does `flow.latest()` return? Can we check that too.

likawind:
@kenxu95 addressed all comments in https://github.com/aqueducthq/aqueduct/pull/1043/commits/73adf2e32a46513abbf1d1d2f98d4a391ec55a72

kenxu95 (approved):

PR #1011, opened 2023-02-21
[1/5] Decouple exec env data model creation and conda env creation

likawind:
## Describe your changes and why you are making these changes
This PR decouples data model creation of execEnv rows and conda env creation such that all operators has an 'execEnv' attached to reflect its dependencies and python versions. The terminology is a little bit confusing after this change since the `execEnv` actually captures dependency information.

This single PR is incomplete as we didn't change any execution logic, and thus the execution will break

## Related issue number (if any)
ENG-2470

## Checklist before requesting a review
- [ ] I have created a descriptive PR title. The PR title should complete the sentence "This PR...".
- [ ] I have performed a self-review of my code.
- [ ] I have included a small demo of the changes. For the UI, this would be a screenshot or a Loom video.
- [ ] If this is a new feature, I have added unit tests and integration tests.
- [ ] I have run the integration tests locally and they are passing.
- [ ] I have run the linter script locally (See `python3 scripts/run_linters.py -h` for usage).
- [ ] All features on the UI continue to work correctly.
- [ ] Added one of the following CI labels:
    - `run_integration_test`: Runs integration tests
    - `skip_integration_test`: Skips integration tests (Should be used when changes are ONLY documentation/UI)

PR #1044, opened 2023-03-03
Enable 'list operators on integration' for compute integrations

likawind:
This PR adds support for listing operators on compute integrations, where this feature was only previously available for data integrations. We also improved the code structure / readability by putting some application logic to operator package for better reusability.

The query looks for:
* all operators whose `spec.engine_config` field contains the given integration ID
* all operators whose `engine_config` is null and the dag's `engine_config` contains the given integration ID.

ENG-2534

## Tests
Added unit test for the new query.
Confirmed feature parity for existing features (data integrations and conda).


src/golang/lib/workflow/operator/service.go
```diff
@@ -0,0 +1,40 @@
...
+		return operatorRepo.GetExtractAndLoadOPsByIntegration(ctx, integrationID, DB)
+	}
+
+	// This feature is not supported for the given service.
+	if !shared.IsComputeIntegration(integrationObject.Service) {
+		return nil, nil
+	}
+
+	if integrationObject.Service == shared.Conda {
+		return operatorRepo.GetWithExecEnv(ctx, DB)
```
cw75, line 36: I remember we changed the contract so that every operator now has an exec env associated with it? If so is this logic still correct?

src/golang/lib/models/shared/service.go
```diff
@@ -71,8 +71,8 @@ func ParseService(s string) (Service, error) {
 	}
 }
 
-func GetRelationalDatabaseIntegrations() map[Service]bool {
-	return map[Service]bool{
+func IsRelationalDatabaseIntegration(service Service) bool {
+	relationalIntegrations := map[Service]bool{
```
cw75, line 75: same comment about making this a variable at the beginning of `service.go`?

src/golang/lib/models/shared/service.go
```diff
@@ -85,6 +85,31 @@ func GetRelationalDatabaseIntegrations() map[Service]bool {
...
+
+func IsComputeIntegration(service Service) bool {
+	computeIntegrations := map[Service]bool{
+		Airflow:    true,
+		Lambda:     true,
+		Conda:      true,
+		Databricks: true,
+		Kubernetes: true,
+		Spark:      true,
+	}
```
cw75, line 109: I think we can make this a variable outside of this function at the beginning of `service.go`?

src/golang/lib/repos/sqlite/operator.go
```diff
@@ -282,6 +282,46 @@ func (*operatorReader) GetRelationBatch(
...
+	integrationID uuid.UUID,
+	DB database.Database,
+) ([]models.Operator, error) {
+	query := fmt.Sprintf(`
+		SELECT DISTINCT %s FROM
+		operator, workflow_dag, workflow_dag_edge
+		WHERE
+		workflow_dag_edge.workflow_dag_id = workflow_dag.id
+		AND (
+			workflow_dag_edge.from_id = operator.id
```
cw75, line 296: I wonder if `from_id` itself is sufficient to capture all operators. Do we ever have an operator that doesn't generate any artifact? I think even an operator that returns nothing will generate an artifact.

src/golang/lib/repos/sqlite/operator.go
```diff
@@ -282,6 +282,46 @@ func (*operatorReader) GetRelationBatch(
...
+					OR json_extract(workflow_dag.engine_config, '$.k8s_config.integration_id') = $1
+					OR json_extract(workflow_dag.engine_config, '$.lambda_config.integration_id') = $1
+					OR json_extract(workflow_dag.engine_config, '$.databricks_config.integration_id') = $1
+				)
+			)
+			OR (
+				json_extract(operator.spec, '$.engine_config.airflow_config.integration_id') = $1
+					OR json_extract(operator.spec, '$.engine_config.k8s_config.integration_id') = $1
+					OR json_extract(operator.spec, '$.engine_config.lambda_config.integration_id') = $1
+					OR json_extract(operator.spec, '$.engine_config.databricks_config.integration_id') = $1
```
cw75, line 313: This seems brittle because every time we add a new compute engine, we will need to update this query to include the new engine. Maybe the right thing to do is to have an explicit `engine_config.integration_id` field (can be null) if this field is generalizable to all compute engine types. If so let's create a task for this. 

For now, instead of hardcoding, can we programmatically generate each condition across all compute engines? Assume we have access to (need to get rid of conda):
```
	computeIntegrations := map[Service]bool{
		Airflow:    true,
		Lambda:     true,
		Conda:      true,
		Databricks: true,
		Kubernetes: true,
		Spark:      true,
	}
```
We can then loop this dict and generate the sql query conditions.

cw75 (changes requested):
@likawind Looks good overall! Requested a few changes and some questions inline.

src/golang/lib/workflow/operator/service.go
likawind: This will be resolved in #1011 as a merge conflict. For now, this particular call should be consistent with main

src/golang/lib/repos/sqlite/operator.go
likawind: I didn't closely verify this, but I need to do this to pass the unit tests where we generate a operator with upstream but no downsteam

cw75: I see. Yeah I think this is fine for now to be safe.

src/golang/lib/repos/sqlite/operator.go
likawind: I think programmatically generating the conditions makes sense! We can keep this set in `models` package so that everything are specified in the same place. I'd like to be conservative on adding a new shared field as it requires some data migration and the assumption may not hold in future (for example, it's already not used in conda now)

cw75: Right. Then as you suggested let's just put this set in `models` and generate the query conditions.

src/golang/lib/models/shared/service.go
```diff
@@ -37,6 +37,43 @@ const (
...
+}
+
+// ServiceToEngineConfigIntegrationIDField contains
+// all services with `integration_id` in its 'engine_config' field.
+// This is used in SQL queries to retrieve engine configs (workflow or operator)
+// based on integration ID.
+//
+// The key should be the service type, and value should be the json tag
+// for the corresponding field that contains the integration ID.
+var ServiceToEngineConfigIntegrationIDField map[Service]string = map[Service]string{
```
cw75, line 70: I think calling it `ServiceToEngineConfigField` is more accurate, because the mapping value is the engine config json tag, not the integration id.

src/golang/lib/workflow/operator/service.go
```diff
@@ -27,14 +27,15 @@ func GetOperatorsOnIntegration(
...
 	}
 
+	// Other eligible cases
 	if integrationObject.Service == shared.Conda {
 		return operatorRepo.GetWithExecEnv(ctx, DB)
 	}
 
-	return operatorRepo.GetByEngineIntegrationID(ctx, integrationID, DB)
+	// This feature is not supported for the given service.
+	return nil, nil
```
cw75, line 40: Currently, which service do we not support? And if so should we error instead of returning nothing?

cw75 (approved):
Unblocking, but left a couple minor comments inline.

src/golang/lib/workflow/operator/service.go
likawind: Notifications for example. I think it's fine to return an empty list for now, mainly that we don't have much checking in frontend / SDK to ensure we don't call any API that involves with this query. So if we want to return an error we need to proceed really carefully.

PR #1045, opened 2023-03-03
[ENG-2273] Updates layout of workflow header

vsreekanti:

This PR makes a number of changes to the workflow header layout:
1. It replaces the old status chip (which is now deprecated) with the new `StatusIndicator` component. This now lives to the left of the workflow header.
2. It makes the workflow title an `h5` instead of an `h4`.
3. It replaces the old version selector layout with a more compact layout that lives right next to the workflow title. 
4. It puts the workflow engine, next run time, and description behind a see more button.
5. It deprecates the workflow status bar.
6. It moves the run workflow button an action button to the right of the workflow DAG layout.
7. It adds navigation buttons to move between workflow runs to the action list to the right of the DAG layout.
8. It adds a refresh button that updates the state of the current workflow run (useful on workflow triggers) without forcing a refresh of the whole page.

Changes 1-5 are covered in the v1 demo below. Changes 6-8 are in the v2 demo, which builds on feedback from v1.

The PR has a lot of LOC, but it's really not that much — a lot of this is from deprecated code or code that was moved around: 
* I moved all the run workflow dialog code into its own component, which was 250 LOC added and close to 300 removed from the `WorkflowHeader` component.
* I added a little bit of wrapping around some of the existing component in the workflow DAG page, which changed the indentation and caused git to think I modified all that code.
* Removing the status bar component was -800 LOC.
* Removing the old status chip was -100 LOC.
* The workflow DAG details page is a bit of a pile of code, so I added comments to it as I worked my way through it for posterity.

ENG-2273

## Loom demo (if any)

v1: https://www.loom.com/share/e3674b4d93ca420caf77d4e31a465995
v2: https://www.loom.com/share/ef93d7e021c34935b18578af4a3faed2


src/ui/common/src/components/workflows/version_selector.tsx
```diff
@@ -83,16 +108,42 @@ export const VersionSelector: React.FC = () => {
...
-          {getMenuItems()}
-        </Select>
-      </FormControl>
+      <Button
+        sx={{
+          backgroundColor: !!menuAnchor ? theme.palette.gray[50] : 'white',
+          ':hover': {
+            backgroundColor: theme.palette.gray[50],
+          },
+          p: 1,
```
likawind, line 117: why we have both `p` and `pr` here?

src/ui/common/src/components/pages/workflow/id/index.tsx
```diff
@@ -68,7 +73,10 @@ const WorkflowPage: React.FC<WorkflowPageProps> = ({
   const urlSearchParams = parse(window.location.search);
   const location = useLocation();
   const path = location.pathname;
-  const [currentTab, setCurrentTab] = React.useState<string>('Details');
+
+  const [currentTab, setCurrentTab] = useState<string>('Details');
+  const [showTriggerDialog, setShowTriggerDialog] = useState(false);
```
likawind, line 78: I don't know if we are using `trigger` elsewhere in the UI, but it may be easier to follow if we start naming everything by `run workflow` in frontend codes

src/ui/common/src/components/pages/workflows/components/RunWorkflowDialog.tsx
```diff
@@ -0,0 +1,276 @@
...
+          sx={{ width: '100%' }}
+        >
+          {errorMessage}
+        </Alert>
+      </Snackbar>
+    </>
+  );
+};
+
+export { RunWorkflowDialog };
```
likawind, line 276: I think you can directly do `export const RunWorkflowDialog: React.FC<>` or do `constant RunWorkflowDialog{...}; export default RunWorkflowDialog`. I personally prefer the latter if the file contains only one component. @agiron123 thoughts?

src/ui/common/src/components/pages/workflows/components/RunWorkflowDialog.tsx
```diff
@@ -0,0 +1,276 @@
+import {
```
likawind, line 1: Should there be some code get removed? Looks like this component is a current feature but I'm having a hard time figuring where they come from

likawind (approved):
I leave some notes on namings, etc. but functionality wise everything looks great!

src/ui/common/src/components/workflows/version_selector.tsx
vsreekanti: Good catch. This was a remnant of something I was trying out.

src/ui/common/src/components/pages/workflows/components/RunWorkflowDialog.tsx
vsreekanti: Yes, this code was removed from `workflowHeader.tsx`. It's reflected in the PR, but it's wrapped up in a bunch of other code changes in that file.

PR #1046, opened 2023-03-03
Add K8s integration tests to our periodic integration tests

kenxu95:
## Describe your changes and why you are making these changes
You can now add a K8s integration to the "compute" section of `test-config.yml`.

Testing: I was able to run flow_test.py against K8s in about 30 minutes (including spinnup and teardown). I need to verify that this works against the full suite. I think I also saw some transient errors that need to be sorted out. 

[UPDATE] Got a full run working: https://github.com/aqueducthq/aqueduct/actions/runs/4326011758/jobs/7552902876

## Related issue number (if any)
ENG-2346

## Checklist before requesting a review
- [ ] I have created a descriptive PR title. The PR title should complete the sentence "This PR...".
- [ ] I have performed a self-review of my code.
- [ ] I have included a small demo of the changes. For the UI, this would be a screenshot or a Loom video.
- [ ] If this is a new feature, I have added unit tests and integration tests.
- [ ] I have run the integration tests locally and they are passing.
- [ ] I have run the linter script locally (See `python3 scripts/run_linters.py -h` for usage).
- [ ] All features on the UI continue to work correctly.
- [ ] Added one of the following CI labels:
    - `run_integration_test`: Runs integration tests
    - `skip_integration_test`: Skips integration tests (Should be used when changes are ONLY documentation/UI)

PR #1051, opened 2023-03-04
Delete dynamic k8s cluster only when no pods are running

cw75:
This PR adds `safeToDeleteCluster` to check whether there're pods still in "Running" or "ContainerCreating" status before deleting the dynamic k8s cluster.

From the SDK, the user can skip this check by doing `.delete(force=True)`. 


src/golang/lib/dynamic/engine.go
```diff
@@ -318,5 +333,44 @@ func ResyncClusterState(
...
+func safeToDeleteCluster(
+	ctx context.Context,
+	engineIntegration *models.Integration,
+) (bool, error) {
+	useSameCluster, err := strconv.ParseBool(engineIntegration.Config[shared.K8sUseSameClusterKey])
+	if err != nil {
+		return false, errors.Wrap(err, "Error parsing use_same_cluster flag")
+	}
+
+	k8sClient, err := k8s.CreateK8sClient(engineIntegration.Config[shared.K8sKubeconfigPathKey], useSameCluster)
```
likawind, line 351: For better reusability, should the client be handled and passed around by caller?

src/golang/lib/dynamic/engine.go
```diff
@@ -189,10 +192,22 @@ func CreateDynamicEngine(
 // If any step fails, it returns an error.
 func DeleteDynamicEngine(
 	ctx context.Context,
+	force bool,
```
likawind, line 195: I feel we could be more specific on what 'force' means by using `skipSafetyCheck` or `skipActiveCheck`. We may just keep the `force` terminology in SDK and handlers that talks to SDK.

likawind (approved):
LGTM, left a few nitpicky comments

sdk/aqueduct/integrations/dynamic_k8s_integration.py
```diff
@@ -42,11 +42,11 @@ def create(self) -> None:
             "Cluster is currently in %s status. It could take 12-15 minutes for the cluster to be ready..."
             % status.value
         )
-        globals.__GLOBAL_API_CLIENT__.edit_engine(
+        globals.__GLOBAL_API_CLIENT__.edit_dynamic_engine(
             action="create", integration_id=str(self._metadata.id)
         )
 
-    def delete(self) -> None:
+    def delete(self, force: bool = False) -> None:
```
kenxu95, line 49: This is a user facing function, so let's please document it and especially note what `force` means.

sdk/aqueduct/integrations/dynamic_k8s_integration.py
```diff
@@ -64,8 +64,13 @@ def delete(self) -> None:
             "Cluster is currently in %s status. It could take 6-8 minutes for the cluster to be terminated..."
             % status.value
         )
-        globals.__GLOBAL_API_CLIENT__.edit_engine(
-            action="delete", integration_id=str(self._metadata.id)
+
+        action = "delete"
+        if force:
+            action = "force-delete"
```
kenxu95, line 70: We should make `create`, `delete` and `force-delete` an enum so it's clear what possible "actions" there are.

src/golang/cmd/server/handler/edit_dynamic_engine.go
```diff
@@ -56,8 +56,9 @@ func (*EditDynamicEngineHandler) Headers() []string {
 }
 
 const (
-	createAction string = "create"
-	deleteAction string = "delete"
+	createAction      string = "create"
+	deleteAction      string = "delete"
+	forceDeleteAction string = "force-delete"
```
kenxu95, line 61: Let's make these a type? Maybe call it `DynamicEngineAction` or something

src/golang/cmd/executor/executor/dynamic_teardown.go
```diff
@@ -64,6 +64,7 @@ func (ex *DynamicTeardownExecutor) Run(ctx context.Context) error {
 			log.Info("Reached keepalive threshold, tearing down the cluster...")
 			if err = dynamic.DeleteDynamicEngine(
 				ctx,
+				false, // don't force deletion if there're pods still running
```
kenxu95, line 67: Can you explain a bit here about why we don't allow force deletions in this codepath in particular?

src/golang/lib/dynamic/engine.go
kenxu95: I agree with this - I'd argue for something more specific though like "skipRunningPodsCheck". Otherwise, I'd argue for adding a function comment here.

src/golang/lib/dynamic/engine.go
```diff
@@ -318,5 +333,44 @@ func ResyncClusterState(
...
+	if err != nil {
+		return false, errors.Wrap(err, "Error parsing use_same_cluster flag")
+	}
+
+	k8sClient, err := k8s.CreateK8sClient(engineIntegration.Config[shared.K8sKubeconfigPathKey], useSameCluster)
+	if err != nil {
+		return false, errors.Wrap(err, "Error while creating K8sClient")
+	}
+
+	pods, err := k8sClient.CoreV1().Pods(k8s.AqueductNamespace).List(ctx, metav1.ListOptions{})
```
kenxu95, line 356: I think we should put any `k8sClient.CoreV1()` calls into helper functions in the `src/golang/lib/k8s`. Maybe we call it `dynamic.go`? I think it would promote better code structure.

kenxu95 (changes requested):

sdk/aqueduct/integrations/dynamic_k8s_integration.py
cw75: adding documentation to both delete and create.

src/golang/lib/dynamic/engine.go
cw75: I pushed this helper down to `lib/k8s/dynamic.go` and am creating the client within the helper. This way I think the scope of this function is clearer.

cw75:
@kenxu95 @likawind addressed all comments.

src/golang/cmd/server/handler/edit_dynamic_engine.go
```diff
@@ -55,9 +55,12 @@ func (*EditDynamicEngineHandler) Headers() []string {
...
 }
 
+type dynamicEngineAction string
+
 const (
-	createAction string = "create"
-	deleteAction string = "delete"
+	createAction      dynamicEngineAction = "create"
+	deleteAction      dynamicEngineAction = "delete"
+	forceDeleteAction dynamicEngineAction = "force-delete"
```
kenxu95, line 63: nit: Can you note that this reflects `K8sClusterActionType` in Python and that they should be kept in sync?

src/golang/cmd/server/handler/edit_dynamic_engine.go
```diff
@@ -79,7 +82,7 @@ func (*EditDynamicEngineHandler) Prepare(r *http.Request) (interface{}, int, err
 
 	return &editDynamicEngineArgs{
 		AqContext:     aqContext,
-		action:        action,
+		action:        dynamicEngineAction(action),
```
kenxu95, line 85: Can you check that the action string is a valid right before this? We can make a simple helper above called `bool isValidDynamicAction(action string)`.

src/golang/lib/dynamic/engine.go
```diff
@@ -187,12 +188,31 @@ func CreateDynamicEngine(
 // 3. Remove the kubeconfig file.
 // 4. Update the dynamic integration's DB record: set config["status"] to "Terminated".
 // If any step fails, it returns an error.
+// If skipPodsStatusCheck is set to false, it checks whether there are pods in Running or ContainerCreating
+// statue and if so, reject the deletion request.
```
kenxu95, line 192: spelling: status

src/golang/lib/dynamic/engine.go
```diff
@@ -318,5 +338,5 @@ func ResyncClusterState(
 	// inconsistent state between the database and terraform. In this case, we resync the state by
 	// deleting the cluster and updating the database state to be Terminated.
 	log.Error("Dynamic k8s cluster might be in an inconsistent state. Resolving state by deleting the cluster...")
-	return DeleteDynamicEngine(ctx, engineIntegration, integrationRepo, db)
+	return DeleteDynamicEngine(ctx, true, engineIntegration, integrationRepo, db)
```
kenxu95, line 341: nit: Can you add an inline comment `true, // skipPodsStatusCheck` for clarity? Each arg would go on it's own line then.

kenxu95 (approved):
Cool! Left a couple more nitty comments but unblocking.

@cw75 do you mind holding off on this until I get https://github.com/aqueducthq/aqueduct/pull/1046 merged? (probably tonight, depending on the results of the final test run, but will let you know). I don't thiiink this PR affects it, but I think that PR should have first-merge priority since it helps prevent these new bugs from being introduced at all.

PR #1052, opened 2023-03-05
Make EKS cluster parameters configurable

cw75:
This PR allows users to configure dynamic k8s cluster parameters. See the SDK code for updates to the APIs.


sdk/aqueduct/backend/api_client.py
```diff
@@ -337,6 +337,7 @@ def edit_dynamic_engine(
         self,
         action: str,
         integration_id: str,
+        config_delta: DynamicK8sConfig = DynamicK8sConfig(),
```
kenxu95, line 340: I don't think making a class the default value is good practice in Python because it leads to some potentially unexpected behavior. Better is to use None and initialize the object inside.

The reason is that the class is instantiated once per definition, which means multiple calls will reuse the same class. This can be really confusing if you mutate the config at any point.

sdk/aqueduct/backend/api_client.py
```diff
@@ -349,8 +350,12 @@ def edit_dynamic_engine(
 
         url = self.construct_full_url(self.EDIT_DYNAMIC_ENGINE_ROUTE_TEMPLATE % integration_id)
 
-        if action == "create" or action == "delete" or action == "force-delete":
-            resp = requests.post(url, headers=headers)
+        body = {
+            "config_delta": config_delta.json(exclude_none=True),
+        }
+
+        if action in ["create", "update", "delete", "force-delete"]:
```
kenxu95, line 357: Request to make this an enum.

sdk/aqueduct/constants/enums.py
```diff
@@ -214,6 +214,7 @@ class SparkRuntimeType(str, Enum, metaclass=MetaEnum):
 
 class K8sClusterStatusType(str, Enum, metaclass=MetaEnum):
     CREATING = "Creating"
+    UPDATING = "Updating"
```
kenxu95, line 217: Let's document exactly what this means. Also, can you write a bit at the top of the enum about about what the lifecycle of this cluster status is expected to be? `creating -> active -> terminating -> terminated`. And where does updating fit in?

sdk/aqueduct/integrations/connect_config.py
```diff
@@ -147,18 +147,37 @@ class SlackConfig(BaseConnectionConfig):
...
 
 
+class DynamicK8sConfig(BaseConnectionConfig):
+    keepalive: Optional[str]
+    cpu_node_type: Optional[str]
+    gpu_node_type: Optional[str]
+    min_cpu_node: Optional[str]
+    max_cpu_node: Optional[str]
+    min_gpu_node: Optional[str]
+    max_gpu_node: Optional[str]
```
kenxu95, line 164: Could you add some description for each of these? Specifically wondering about things like:
- Are these strings of numbers for the min/max stuff? 
- What does do the min/max mean? Are they hard bounds? Are they inclusive? What size is the cluster initially created to, and can that be outside of the bounds?
- What are the accepted `cpu/gpu_node_type`s?

sdk/aqueduct/integrations/connect_config.py
```diff
@@ -147,18 +147,37 @@ class SlackConfig(BaseConnectionConfig):
...
+    secret_access_key: str
+    region: str
+    k8s: Optional[DynamicK8sConfig]
+
+
+class _AWSConfigWithStringField(BaseConnectionConfig):
+    access_key_id: str
+    secret_access_key: str
+    region: str
+    k8s_serialized: Optional[str]
```
kenxu95, line 178: Can we document what type of serialization we are using? And what this is serializing.

sdk/aqueduct/integrations/connect_config.py
```diff
@@ -147,18 +147,37 @@ class SlackConfig(BaseConnectionConfig):
...
+
+
+class AWSConfig(BaseConnectionConfig):
+    access_key_id: str
+    secret_access_key: str
+    region: str
+    k8s: Optional[DynamicK8sConfig]
+
+
+class _AWSConfigWithStringField(BaseConnectionConfig):
```
kenxu95, line 174: I probably would prefer `_AWSConfigWithSerializedConfig`?

sdk/aqueduct/integrations/dynamic_k8s_integration.py
```diff
@@ -24,7 +42,9 @@ def status(self) -> str:
 
         return engine_statuses[self._metadata.name].status.value
 
-    def create(self) -> None:
+    def create(self, config_delta: Union[Dict[str, str], DynamicK8sConfig] = {}) -> None:
```
kenxu95, line 45: These are user facing functions so we definitely need to document what `config_delta` means and does. I think I prefer the term `config` too. It's not really a delta if we're updating it to the new end state right?

sdk/aqueduct/integrations/dynamic_k8s_integration.py
```diff
@@ -34,16 +54,49 @@ def create(self) -> None:
...
+            config_delta=config_delta,
+        )
+
+    def update(self, config_delta: Union[Dict[str, str], DynamicK8sConfig] = {}) -> None:
+        config_delta = parse_dynamic_k8s_config(config_delta)
+
+        engine_statuses = globals.__GLOBAL_API_CLIENT__.get_dynamic_engine_status(
+            engine_integration_ids=[str(self._metadata.id)]
+        )
+        if len(engine_statuses) != 1:
```
kenxu95, line 79: nit: I'd find it a bit more readable if we change this to `len(engine_statuses) == 0` and assert `len(engine_status) <= 1`. Adds extra protection too against unexpected duplicate integrations with the same name.

src/golang/cmd/server/handler/edit_dynamic_engine.go
```diff
@@ -57,8 +59,10 @@ func (*EditDynamicEngineHandler) Headers() []string {
 
 const (
 	createAction      string = "create"
+	updateAction      string = "update"
 	deleteAction      string = "delete"
 	forceDeleteAction string = "force-delete"
+	configDeltaKey    string = "config_delta"
```
kenxu95, line 65: Would appreciate some documentation for `configDeltaKey`. Where do we expect to use this key?

kenxu95 (changes requested):

sdk/aqueduct/backend/api_client.py
cw75: wow interesting, didn't know that...

sdk/aqueduct/integrations/connect_config.py
cw75: Added description to each field. In terms of the accepted cpu/gpu node types, it's whatever that's supported by AWS that the account has permission to provision.

sdk/aqueduct/integrations/dynamic_k8s_integration.py
cw75: I added explanation of what `config_delta` means. It is a "delta" so I'd partial to keeping this name.

cw75:

sdk/aqueduct/integrations/connect_config.py
```diff
@@ -147,18 +147,48 @@ class SlackConfig(BaseConnectionConfig):
...
+    secret_access_key: str
+    region: str
+    k8s: Optional[DynamicK8sConfig]
+
+
+class _AWSConfigWithSerializedConfig(BaseConnectionConfig):
+    access_key_id: str
+    secret_access_key: str
+    region: str
+    k8s_serialized: Optional[str]  # this is a json-serialized string of DynamicK8sConfig
```
kenxu95, line 189: specifically, this is intended to be the serialized version of `AWSConfig.k8s` right? Might be good to make that relationalship clear.

sdk/aqueduct/integrations/connect_config.py
```diff
@@ -275,6 +309,15 @@ def _prepare_slack_config(config: SlackConfig) -> _SlackConfigWithStringField:
     )
 
 
+def _prepare_aws_config(config: AWSConfig) -> _AWSConfigWithSerializedConfig:
+    return _AWSConfigWithSerializedConfig(
+        access_key_id=config.access_key_id,
+        secret_access_key=config.secret_access_key,
+        region=config.region,
+        k8s_serialized=(None if config.k8s is None else config.k8s.json(exclude_none=True)),
```
kenxu95, line 317: nit: I think `config.k8s or config.k8s.json(exclude_none=True)` does what you want in a more succinct fashion.

sdk/aqueduct/integrations/connect_config.py
```diff
@@ -147,18 +147,48 @@ class SlackConfig(BaseConnectionConfig):
...
+    # for the node types available.
+    cpu_node_type: Optional[str]
+    # The EC2 instance type of the GPU node group. See https://aws.amazon.com/ec2/instance-types/
+    # for the node types available.
+    gpu_node_type: Optional[str]
+    # Minimum number of nodes in the CPU node group. The cluster autoscaler cannot scale below this number.
+    # This is also the initial number of CPU nodes in the cluster.
+    min_cpu_node: Optional[str]
+    # Maximum number of nodes in the CPU node group. The cluster autoscaler cannot scale above this number.
+    max_cpu_node: Optional[str]
```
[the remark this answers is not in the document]
kenxu95: On the SDK that we could type it was Dict[str, Union[int, str]]. If that's too long we can always alias it.

src/golang/lib/models/shared/integration_config.go
```diff
@@ -100,9 +139,58 @@ type SlackConfig struct {
...
+
+		if fieldValue != "" {
+			configMap[fieldName] = fieldValue
+		}
+	}
+
+	return configMap
+}
+
+func (config *DynamicK8sConfig) Merge(newConfig *DynamicK8sConfig) {
```
kenxu95, line 171: Could we add docstrings for these two methods? Especially since their implementations are a little opaque.

kenxu95: I also wonder if `Update()` is a better term? Don't have a big preference though

kenxu95 (commented):
Mainly wondering why the configuration dictionary can be of type Dict[str, Union[str, int]] - all the other comments are mainly about internal polish.

sdk/aqueduct/integrations/connect_config.py
cw75: hmm won't this just return a bool?

cw75:

sdk/aqueduct/integrations/connect_config.py
kenxu95: No it'll return the value of the first expression, unless it's evaluated to False (aka None in this case), whereas it'll use the second expression. It should do the same thing as your inline if-else statement.

sdk/aqueduct/integrations/connect_config.py
```diff
@@ -150,18 +150,55 @@ class SlackConfig(BaseConnectionConfig):
...
+    # Minimum number of nodes in the GPU node group. The cluster autoscaler cannot scale below this number.
+    # This is also the initial number of GPU nodes in the cluster.
+    min_gpu_node: Optional[str]
+    # Maximum number of nodes in the GPU node group. The cluster autoscaler cannot scale above this number.
+    max_gpu_node: Optional[str]
```
kenxu95, line 178: Shouldn't the countable fields be `Optional[Union[str, int]]` now? Since we construct this model from user input.

sdk/aqueduct/integrations/dynamic_k8s_integration.py
```diff
@@ -1,10 +1,38 @@
...
 from aqueduct import globals
 
 
+def parse_dynamic_k8s_config(
+    config_delta: Union[Dict[str, Union[int, str]], DynamicK8sConfig]
+) -> DynamicK8sConfig:
+    if not isinstance(config_delta, dict) and not isinstance(config_delta, DynamicK8sConfig):
+        raise InvalidUserArgumentException(
+            "`config_delta` argument must be either a dict or DynamicK8sConfig."
+        )
```
kenxu95, line 18: Would it be possible to typecheck the possible keys here? The pydantic error messaging is really bad so it might be nice for us to give them some help here.

kenxu95 (approved):
Unblocking but please address comments :)

sdk/aqueduct/integrations/connect_config.py
cw75: @kenxu95 honestly I find my original version much easier to understand. :) I'll keep it unless you feel strongly.

cw75: and I don't think I understand your explanation still...I don't want it to return the value of the first expression (`config.k8s`) right? And if it is evaluated to None, how can I then return the second expression?

sdk/aqueduct/integrations/dynamic_k8s_integration.py
cw75: I'm trying to just see if `config_delta = DynamicK8sConfig(**config_delta, extra="ignore")` works. If so, even if they provide extra keys, we should be fine.

cw75: Actually, even without the `extra="ignore"`, I think the error message looks fine: if you have extra field it'll just say extra field not permitted. Also if the type of each field is wrong, the pydantic serialization will succeed but the Golang side will catch the deserialization error, which is also fine, so I'll keep the current implementation now.

sdk/aqueduct/integrations/connect_config.py
kenxu95: `config.k8s or config.k8s.json(exclude_none=True)`

config.k8s.json(exclude_none=True) is the second expression that is returned.

kenxu95: fine with me either way, I just think I saw this pattern in our javascript code too

PR #1055, opened 2023-03-06
Adds backend support for changing storage layer back to local

saurav-c:
This PR implements a new route `api/config/storage/{integrationID}` which can be used to change the storage layer to an existing integration. Right now there is only support for changing back to the local storage layer using this route.

There is future work to add support for specifying any existing integration (that is compatible for storage, such as S3 or GCS) and setting that as the new storage layer. This is tracked by: https://linear.app/aqueducthq/issue/ENG-2574/add-backend-support-for-changing-the-storage-layer-to-an-existing

ENG 2175

## Test
Tested via a Postman route

### DIFF
--- a/src/golang/cmd/server/handler/configure_storage.go
+++ b/src/golang/cmd/server/handler/configure_storage.go
@@ -1,0 +1,48 @@
+package handler
+
+import (
+	"context"
+	"net/http"
+
+	aq_context "github.com/aqueducthq/aqueduct/lib/context"
+	"github.com/aqueducthq/aqueduct/lib/database"
+	"github.com/aqueducthq/aqueduct/lib/engine"
+	"github.com/aqueducthq/aqueduct/lib/repos"
+)
+
+// Route: /config/storage
+// Method: POST
+// Request:
+//
+//	Headers:
+//		`api-key`: user's API Key
+//
+// Response: none
+type ConfigureStorageHandler struct {
+	PostHandler
+
+	Database database.Database
+	Engine   engine.Engine
+
+	ArtifactRepo repos.Artifact
+	DAGRepo      repos.DAG
+	DAGEdgeRepo  repos.DAGEdge
+	OperatorRepo repos.Operator
+	WorkflowRepo repos.Workflow
+}
+
+type configureStorageArgs struct {
+	*aq_context.AqContext
+}
+
+func (*ConfigureStorageHandler) Name() string {
+	return "ConfigureStorage"
+}
+
+func (h *ConfigureStorageHandler) Prepare(r *http.Request) (interface{}, int, error) {
+	return nil, -1, nil
+}
+
+func (h *ConfigureStorageHandler) Perform(ctx context.Context, interfaceArgs interface{}) (interface{}, int, error) {
+	return nil, -1, nil
+}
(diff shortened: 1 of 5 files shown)


saurav-c:
cc @agiron123 so you know which api route to use from the frontend for changing the storage layer. note that this route will perform the migration synchronously

src/golang/cmd/server/handler/configure_storage.go
```diff
@@ -0,0 +1,140 @@
...
+
+	Database database.Database
+
+	ArtifactRepo       repos.Artifact
+	ArtifactResultRepo repos.ArtifactResult
+	DAGRepo            repos.DAG
+	IntegrationRepo    repos.Integration
+	OperatorRepo       repos.Operator
+
+	PauseServer   func()
```
likawind, line 40: What's the go standard to call a function pointer? I personally feel `PauseServerFn` is easier to understand

src/golang/cmd/server/handler/configure_storage.go
```diff
@@ -0,0 +1,140 @@
...
+		ctx,
+		&currentStorageConfig,
+		&newStorageConfig,
+		args.OrgID,
+		h.DAGRepo,
+		h.ArtifactRepo,
+		h.ArtifactResultRepo,
+		h.OperatorRepo,
+		h.IntegrationRepo,
+		h.Database,
```
likawind, line 127: do we need a transaction here?

likawind (commented):
Most implementation makes sense to me by eyeballing it, I have a few clarification questions which we can chat in person or in eng sync:
* you mentioned that storage migration requires an integration to exists for target storage. Is this resolved or not an issue?
* Do we want a separate API route to trigger this or just want to migrate back when the integration is deleted? If the former case, there seems to be some required UI work as well since we don't explicitly provide a UI entry point to manage storage. If the latter case, I'm wondering if we should move some implementations to `lib` and call it as a part of `delete_integration` handler.

src/golang/cmd/server/handler/configure_storage.go
saurav-c: Nope, the function being called performs operations that need to be atomic inside of a transation.

src/golang/cmd/server/handler/configure_storage.go
saurav-c: Sure, addressed

kenxu95 (commented):
I think we should definitely add tests for this in our `sdk/no_concurrency` tests, given how many problems we've been seeing with this.

Maybe we can put the `no_concurrency` folder inside `integration_tests/sdk` so that we get to share the same setup framework (test-config and test-credentials). I suspect you can write a test case decorator that skips these tests if concurrency is set to > 1.

Eg.
```
def test_example(request):
    num_workers = request.config.getoption("-n")
```
Happy to help pair on this!

saurav-c:
Discussed with @kenxu95 offline that adding integration tests to this will require 1-2 days of work so we're tracking this in https://linear.app/aqueducthq/issue/ENG-1885/add-integration-tests-for-storage-migration-process.

likawind (approved):
Based on above discussions, things looks good to the scope of this PR!

PR #1049, opened 2023-03-04
ENG 2439 add engine and resource to metrics and checks

likawind:
## Describe your changes and why you are making these changes
This PR adds engine and resource configs for metrics and checks so that the behavior is consistent to `@op()` decorators. We simply moved the relevant parts in `@op()` decorator to a helper function and applies it to both `check()` and `metrics()` implementations.

## Related issue number (if any)
ENG 2439

## Loom demo (if any)
Testing WIP

## Checklist before requesting a review
- [ ] I have created a descriptive PR title. The PR title should complete the sentence "This PR...".
- [ ] I have performed a self-review of my code.
- [ ] I have included a small demo of the changes. For the UI, this would be a screenshot or a Loom video.
- [ ] If this is a new feature, I have added unit tests and integration tests.
- [ ] I have run the integration tests locally and they are passing.
- [ ] I have run the linter script locally (See `python3 scripts/run_linters.py -h` for usage).
- [ ] All features on the UI continue to work correctly.
- [ ] Added one of the following CI labels:
    - `run_integration_test`: Runs integration tests
    - `skip_integration_test`: Skips integration tests (Should be used when changes are ONLY documentation/UI)

PR #1056, opened 2023-03-07
Eng 2509 fix bug where the metrics plot fails to render when canceled

likawind:
This PR fixes a bug where we failed to render metrics when its upstream is cancelled. Specifically we fixed two issues:
* ensure the `get_artifact_results` handler returns all metadata without the content when an artifact is cancelled.
* Render canceled metric history in list but hide the graph when there's nothing to show.

To make the issue easier to surface in the future, we also added a manual QA workflow that consists of a failed op and some downstream, cancelled metrics and checks.

ENG-2509
## Loom demo (if any)
Screenshot 2023-03-06 at 4 47 06 PM


vsreekanti:
@likawind, this PR seems to depend on the other PR you opened (#1049). Can you either refactor that out or change this to compare against that branch, so we can view the relevant code?

likawind:
@vsreekanti thanks, it's merged now

vsreekanti:
Thanks, @likawind! Just to clarify, does this fix the potential issue where there are some valid runs and some failed runs of the metric? I've actually noticed this as well, so I want to be sure!

likawind:
@vsreekanti do you mean that the same metric's states are different across runs?

I haven't verify yet but I believe so. The fixed behavior should be that, we plot all available results, and we show canceled ones in the list section

vsreekanti:
Yeah, imagine you have:
1. Success
2. Failure
3. Success

Does this now graph correctly? It would be great to verify just in case!

likawind:
@vsreekanti this is how it looks when states are mixed, we plot any metrics that has values and show all of them in the list
Screenshot 2023-03-06 at 9 33 26 PM

vsreekanti:
Okay, that's perfect!

likawind:
@vsreekanti I did find a separate issue here: https://linear.app/aqueducthq/issue/ENG-2578/fix-a-bug-where-metric-history-list-shows-failed-metrics-as-canceled

cw75 (approved):

PR #1061, opened 2023-03-07
Allow registering AWS integration with custom config from the UI

cw75:
See screenshots below:

Screen Shot 2023-03-07 at 1 28 35 AM
Screen Shot 2023-03-07 at 1 28 46 AM


vsreekanti:
Hm, this is quite long now, and I'm not the biggest fan of modals that scroll. We should think about making this a multi-page set up. Something like this: 
1. First, you enter access credentials, you hit Next.
2. Then you get the ability to optionally specify any of the other fields.
3. Then you confirm.

cw75:
@vsreekanti Yeah I like your suggestion. I need to look into how to implement the multi-page version though. Do you think it's a blocker to shipping? If not, I'll create a task and work on other higher priority tasks and then get back to this one.

vsreekanti:
We're not shipping this PR in this release, right?

vsreekanti:
I would not release the feature as-is, but I don't have a strong opinion on the order in which you make the PRs, if that's what you're asking.

cw75:
@vsreekanti Yeah we're planning to ship the project 3/22 and I was asking if the multi-page design is a release blocker. You said "I would not release the feature as-is" so th answer seems to be "yes". I'll work this into my planning.

vsreekanti:
Got it — yes, I wasn't sure if you were asking about merging this PR or releasing the feature. I'm fine with merging this PR, but I would prefer not to release the feature without the multi-page design.

cw75:
@likawind let me know if the state management part makes sense. It seems to work fine during my testing.

src/ui/common/src/components/integrations/dialogs/awsDialog.tsx
```diff
@@ -127,6 +237,15 @@ export const AWSDialog: React.FC<Props> = ({
       </Box>
       {value?.type === AWSCredentialType.AccessKey && accessKeyTab}
       {value?.type === AWSCredentialType.ConfigFilePath && configPathTab}
+      <Box sx={{ borderBottom: 1, borderColor: 'divider', mb: 2 }}>
+        <Tabs value={engineType} onChange={(_, value) => setEngineType(value)}>
```
likawind, line 241: how do you send this field to backend? It appears we didn't call 'onUpdateField' when this option is changed?

src/ui/common/src/components/integrations/dialogs/awsDialog.tsx
```diff
@@ -111,6 +128,99 @@ export const AWSDialog: React.FC<Props> = ({
...
+      />
+      <IntegrationTextInputField
+        spellCheck={false}
+        required={false}
+        label="CPU node type"
+        description="The EC2 instance type of the CPU node group."
+        placeholder={K8sPlaceholders.cpu_node_type}
+        onChange={(event) => {
+          k8sConfigs['cpu_node_type'] = event.target.value;
+          onUpdateField('k8s_serialized', JSON.stringify(k8sConfigs));
```
likawind, line 156: did you verify if this is working smoothly? It's different from what we did in some other places (where we use an additional internal state to track), but I also don't see any issue with your way

Try to type fast, delete fast, and switch between various fields. See if there's any latency when you type, and see if the created integration's fields matches what's in the dialog box. If that works, we can go with this solution.

cw75: I think so yeah it's working.

src/ui/common/src/components/integrations/dialogs/awsDialog.tsx
cw75: Do I need to send this to backend? This var is just to select which tab (we only have k8s for now) to show and it doesn't need to be sent to the server side.

likawind: I see, I thought it's a part of the config, makes sense now

cw75:
@likawind could you stamp this if it looks good?

likawind (approved):
:shipit:

PR #1065, opened 2023-03-08
Fix bug where error occurred during job manager launch is not caught as operator error

cw75:
The original bug was when running with k8s, an extract operator that attempts to read from the local sqlite database will silently fail. The root cause is when error occurs in operator.Launch(), we weren't setting the operator's exec state accordingly.

This PR also fixes a few other smaller issues and comment typos.

@kenxu95 let's sync up tmr on writing a k8s integration test for this.

### DIFF
--- a/integration_tests/sdk/aqueduct_tests/integration_test.py
+++ b/integration_tests/sdk/aqueduct_tests/integration_test.py
@@ -1,10 +1,14 @@
 import pytest
+from aqueduct.constants.enums import ServiceType
 from aqueduct.error import (
+    AqueductError,
     InvalidIntegrationException,
     InvalidUserActionException,
     InvalidUserArgumentException,
 )
 from pydantic import ValidationError
+
+from aqueduct import global_config
 
 from ..shared.data_objects import DataObject
 from .extract import extract
@@ -17,7 +21,7 @@ def test_invalid_source_integration(client):
         client.integration(name="wrong integration name")
 
 
-def test_invalid_destination_integration(client, data_integration):
+def test_invalid_destination_integration(data_integration):
     table_artifact = extract(data_integration, DataObject.SENTIMENT)
     output_artifact = dummy_sentiment_model(table_artifact)
 
@@ -46,3 +50,11 @@ def test_invalid_connect_integration(client):
     # Invalid config raises a pydantic error.
     with pytest.raises(ValidationError):
         client.connect_integration("New Integration", "SQLite", {})
+
+
+@pytest.mark.enable_only_for_engine_type(ServiceType.K8S)
+def test_sqlite_with_k8s(data_integration, engine):
+    """Tests that running an extract operator that reads data from a SQLite database using k8s should fail."""
+    global_config({"engine": engine})
+    with pytest.raises(AqueductError):
+        extract(data_integration, DataObject.SENTIMENT)
--- a/integration_tests/sdk/aqueduct_tests/integration_test.py
+++ b/integration_tests/sdk/aqueduct_tests/integration_test.py
@@ -56,5 +56,5 @@ def test_invalid_connect_integration(client):
 def test_sqlite_with_k8s(data_integration, engine):
     """Tests that running an extract operator that reads data from a SQLite database using k8s should fail."""
     global_config({"engine": engine})
-    with pytest.raises(AqueductError):
+    with pytest.raises(AqueductError, match="Unknown integration service provided SQLite"):
         extract(data_integration, DataObject.SENTIMENT)
--- a/integration_tests/sdk/aqueduct_tests/integration_test.py
+++ b/integration_tests/sdk/aqueduct_tests/integration_test.py
@@ -61,7 +61,7 @@ def test_sqlite_with_k8s(data_integration, engine):
     with pytest.raises(AqueductError, match="Unknown integration service provided SQLite"):
         extract(data_integration, DataObject.SENTIMENT)
 
-        
+
 @pytest.mark.enable_only_for_local_storage()
 def test_compute_integration_without_cloud_storage(client):
     with pytest.raises(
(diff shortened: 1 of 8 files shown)


sdk/aqueduct/constants/enums.py
```diff
@@ -100,6 +100,7 @@ class RelationalDBServices(str, Enum, metaclass=MetaEnum):
 class ExecutionStatus(str, Enum, metaclass=MetaEnum):
     UNKNOWN = "unknown"
     SUCCEEDED = "succeeded"
+    RUNNING = "running"
```
cw75, line 103: We're missing this entry, which is why pydantic was complaining on the SDK side for the preview response.

src/golang/lib/job/errors.go
```diff
@@ -25,7 +25,7 @@ const (
 	// For example, polling a Lambda JobManager will return this error because there is no concept of
 	// fetching a specific job's information from Lambda. The caller must figure out the job status
 	// through other means.
-	Noop = 2
+	Noop = 3
```
cw75, line 28: The original value is duplicated with `JobMissing` above, so I think this is just a bug?

src/golang/lib/job/k8s.go
```diff
@@ -140,7 +140,7 @@ func (j *k8sJobManager) Launch(ctx context.Context, name string, spec Spec) JobE
 
 	containerRepo, err := mapJobTypeToDockerImage(spec, launchGpu)
 	if err != nil {
-		return systemError(err)
+		return userError(err)
```
cw75, line 143: I think this should be a user error; it catches the k8s <> sqlite compatibility issue

src/golang/lib/workflow/operator/base.go
```diff
@@ -165,7 +165,25 @@ func (bo *baseOperator) launch(ctx context.Context, spec job.Spec) error {
 		}
 	}
 
-	return bo.jobManager.Launch(ctx, spec.JobName(), spec)
+	err := bo.jobManager.Launch(ctx, spec.JobName(), spec)
```
cw75, line 168: This is the fix to the core bug.

src/golang/lib/workflow/operator/base.go
```diff
@@ -165,7 +165,25 @@ func (bo *baseOperator) launch(ctx context.Context, spec job.Spec) error {
...
+			)
+		} else {
+			log.Errorf("Unexpected job error code %d", err.Code())
+			bo.UpdateExecState(
+				unknownSystemFailureExecState(err, "Job manager Launch failed due to system error"),
+			)
+		}
+	}
+
+	return err
```
cw75, line 186: This is the fix to the core bug.

integration_tests/sdk/aqueduct_tests/integration_test.py
```diff
@@ -46,3 +50,11 @@ def test_invalid_connect_integration(client):
     # Invalid config raises a pydantic error.
     with pytest.raises(ValidationError):
         client.connect_integration("New Integration", "SQLite", {})
+
+
+@pytest.mark.enable_only_for_engine_type(ServiceType.K8S)
+def test_sqlite_with_k8s(data_integration, engine):
+    """Tests that running an extract operator that reads data from a SQLite database using k8s should fail."""
+    global_config({"engine": engine})
+    with pytest.raises(AqueductError):
```
kenxu95, line 59: can you use the "match=<err msg>" argument here to make it clear what the expected error message is?

src/golang/lib/workflow/operator/base.go
```diff
@@ -165,7 +165,25 @@ func (bo *baseOperator) launch(ctx context.Context, spec job.Spec) error {
...
+				jobManagerUserFailureExecState(err, "Job manager Launch failed due to user error"),
+			)
+		} else if err.Code() == job.System {
+			bo.UpdateExecState(
+				unknownSystemFailureExecState(err, "Job manager Launch failed due to system error"),
+			)
+		} else {
+			log.Errorf("Unexpected job error code %d", err.Code())
+			bo.UpdateExecState(
+				unknownSystemFailureExecState(err, "Job manager Launch failed due to system error"),
```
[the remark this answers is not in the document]
cw75: @kenxu95 but here we are just logging the err message on the server and not returning it to the client side, so I think revealing the detail here is still fine and useful?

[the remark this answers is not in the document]
cw75: @kenxu95 for the unexpected case, what additional detail do you want to add? `unknownSystemFailureExecState` is already logging the error message so I was thinking it should already give the detail?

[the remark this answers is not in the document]
kenxu95: I meant like adding a little more detail to the `unknownSystemFailureExecState()` message, but it's not a big deal.

cw75:
@kenxu95 addressed comments.

src/golang/lib/job/errors.go
```diff
@@ -10,22 +10,22 @@ type JobErrorCode int
 
 const (
 	// System indicates an unexpected system issue that we cannot recover from.
-	System JobErrorCode = 0
+	System JobErrorCode = iota
 
 	// User error code indicates that the issue was the user's fault, and to surface that message
 	// to the user.
-	User = 1
+	User = iota
```
kenxu95, line 17: Wait you don't need to set the other ones to iota right, it goes:
```
System JobErrorCode = iota
User
JobMissing
Noop
```

src/golang/lib/workflow/operator/base.go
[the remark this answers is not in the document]
kenxu95: Can we uncapitalize "Launch" though?

kenxu95 (approved):
Unblocking but a couple last requests.

src/golang/lib/job/errors.go
cw75: got it thx!

cw75:
@kenxu95 I changed to "Job manager's Launch API..."

PR #1066, opened 2023-03-08
Fix bug where compute integration can be registered without configuring cloud storage

cw75:
Also fixes another bug where providing an enum type to `connect_integration` fails.

### DIFF
--- a/integration_tests/sdk/aqueduct_tests/integration_test.py
+++ b/integration_tests/sdk/aqueduct_tests/integration_test.py
@@ -51,8 +51,12 @@ def test_invalid_connect_integration(client):
         client.connect_integration("New Integration", "SQLite", {})
 
 
+@pytest.mark.enable_only_for_local_storage()
 def test_compute_integration_without_cloud_storage(client):
-    with pytest.raises(InvalidRequestError, match="You need to setup cloud storage as metadata store before registering compute integration of type Kubernetes."):
+    with pytest.raises(
+        InvalidRequestError,
+        match="You need to setup cloud storage as metadata store before registering compute integration of type Kubernetes.",
+    ):
         client.connect_integration(
             name="compute integration without cloud storage",
             service=ServiceType.K8S,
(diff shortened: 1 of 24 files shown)


sdk/aqueduct/backend/api_client.py
```diff
@@ -254,10 +254,10 @@ def list_tables(self, integration_id: str) -> List[str]:
         return [x for x in resp.json()["object_names"]]
 
     def connect_integration(
-        self, name: str, service: ServiceType, config: IntegrationConfig
+        self, name: str, service: Union[str, ServiceType], config: IntegrationConfig
     ) -> None:
         integration_service = service
-        if not isinstance(integration_service, str):
```
cw75, line 260: @kenxu95 I had to change this because `if not isinstance(ServiceType.K8S, str):` actually evaluates to False, so this line wasn't working before...

sdk/aqueduct/backend/api_client.py
```diff
@@ -254,10 +254,10 @@ def list_tables(self, integration_id: str) -> List[str]:
...
+        if isinstance(integration_service, ServiceType):
```
cw75, line 260: I verified that this only evaluates to True for enum types, not for str.

sdk/aqueduct/backend/api_client.py
kenxu95: Oh right cus the enum inherits from string...

integration_tests/sdk/aqueduct_tests/integration_test.py
```diff
@@ -46,3 +49,12 @@ def test_invalid_connect_integration(client):
     # Invalid config raises a pydantic error.
     with pytest.raises(ValidationError):
         client.connect_integration("New Integration", "SQLite", {})
+
+
+def test_compute_integration_without_cloud_storage(client):
+    with pytest.raises(InvalidRequestError):
```
kenxu95, line 55: Please add a `match=<err msg>` argument to the `pytest.raises()` call

kenxu95 (approved):
Couple small comments but unblocking

cw75:
@kenxu95 I implemented the new decorator and verified that it's working.

PR #1070, opened 2023-03-10
Fix bug where using S3 integration with non-existent filepath succeeds

Fanjia-Yan:
This PR error handles the case when we get the file(s) in a non-existent S3 folder. There are several cases:

1. S3 folder does not exist, this will prompt an error in server "Given S3 directory path '%s' does not exist"
2. S3 folder exists but empty, this request will succeed and return an empty tuple.
3. S3 file does not exist, this is handled already.

The error detection is done by taking off the last character in `path`(i.e. `\`) symbol and find match that contains `\` as delimiter. 

**Error message when path to folder does not exist:**
Running this block of code:

Screen Shot 2023-03-09 at 6 55 35 PM

The error I got from Jupyter notebook:
Screen Shot 2023-03-09 at 6 55 24 PM

The error I got from Aqueduct server:
Screen Shot 2023-03-09 at 6 55 51 PM

**Empty Folder:** 
The block of code I have run:
Screen Shot 2023-03-09 at 7 25 22 PM

The execution result:
Screen Shot 2023-03-09 at 7 25 27 PM

https://linear.app/aqueducthq/issue/ENG-2477/fix-bug-where-using-s3-integration-with-non-existent-filepath-succeeds


vsreekanti:
@Fanjia-Yan, could you please clarify which screenshots correspond to which part of the app in your first example? In particular, I'd like to know which one is the error message we see in the Jupyter notebook when running this function.

Fanjia-Yan:
> @Fanjia-Yan, could you please clarify which screenshots correspond to which part of the app in your first example? In particular, I'd like to know which one is the error message we see in the Jupyter notebook when running this function.

@vsreekanti Just Updated.

Fanjia-Yan:
I have also asked this on linear for this issue, might as well put it here:
"I notice that most errors are threw in aqueduct_executor. Therefore on Jupyter notebook it will only show `We couldn't execute the provided query. Please double check your query is correct`. While on server it has detailed information such as `Given S3 directory path 'hello/' does not exist`  or `NoSuchKey) when calling the GetObject operation: The specified key does not exist`. Are those expected or should we surface those error to SDK user?"

vsreekanti:
Hm, okay. It seems like we're burying the lede a little bit here. The interesting error message is that the path doesn't exist, but we're just saying we could execute the "query" in the notebook. Not only is query not a good word for reading a file, the error message is not very informative. Could we surface the other error message instead?

Fanjia-Yan:
> Hm, okay. It seems like we're burying the lede a little bit here. The interesting error message is that the path doesn't exist, but we're just saying we could execute the "query" in the notebook. Not only is query not a good word for reading a file, the error message is not very informative. Could we surface the other error message instead?

Agree! I will create a new task on linear on making the error message more informative. As of now, should we get this PR in?

kenxu95 (changes requested):
I need a test for this in `data_integration_tests/s3_test.py`

You can probably just add to `test_s3_basic_fetch_failure()`. Maybe we call it `test_s3_fetch_failure()`

Fanjia-Yan:
Sorry for the confusion, the error message surfacing works, it just my Jupyter is limiting its word character so it is not showing :)
Screen Shot 2023-03-13 at 3 58 55 PM

Fanjia-Yan:
> I need a test for this in `data_integration_tests/s3_test.py`
> 
> You can probably just add to `test_s3_basic_fetch_failure()`. Maybe we call it `test_s3_fetch_failure()`

I added that inside `test_s3_basic_fetch_failure`.

vsreekanti:
Sweet — great work, @Fanjia-Yan!

src/python/aqueduct_executor/operators/connectors/data/s3.py
```diff
@@ -127,7 +127,16 @@ def extract(self, params: extract.S3Params) -> Any:
             if len(path) == 0:
                 raise Exception("S3 file path cannot be an empty string.")
             if path[-1] == "/":
-                # This means the path is a directory, and we will do a prefix search.
+                # Check whether the folder path exists in S3 by just checking the folder
+                # rather than the content in the folder. `MaxKeys = 2` allows file and
+                # directory sharing the same name. The return value will only contains
+                # `CommonPrefixes` if there is a folder match.
```
cw75, line 133: @Fanjia-Yan I'm a bit confused by the comment here. What does "`MaxKeys = 2` allows file and directory sharing the same name" mean? Also what does `CommonPrefixes` mean? Do you have a pointer to the documentation that I can take a look at?

Fanjia-Yan: Here is the doc for the API https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/list_objects.html. So what I am trying to do here is to distinguish empty file folder and non-existing file folder. To do that, say I'm querying on folder `aqueduct/` I'm searching over the its parent directory to see `aqueduct` exist and adding the `delimitter`, the output in `CommonPrefixes`will have `\` which indicates that the folder exists.

integration_tests/sdk/data_integration_tests/s3_test.py
```diff
@@ -422,6 +422,10 @@ def test_s3_basic_fetch_failure(client, data_integration):
     with pytest.raises(AqueductError, match="The specified key does not exist."):
         data_integration.file("asdlkf", artifact_type=ArtifactType.TABLE, format="parquet")
 
+    # Fetch a path to directoy that does not exist will fail.
```
kenxu95, line 425: nit: "directory"

src/python/aqueduct_executor/operators/connectors/data/s3.py
```diff
@@ -127,7 +127,16 @@ def extract(self, params: extract.S3Params) -> Any:
...
+                list_object_dic = self.s3.meta.client.list_objects(
+                    Bucket=self.bucket, Prefix=path[:-1], Delimiter="/", MaxKeys=2
+                )
+                if "CommonPrefixes" not in list_object_dic:
```
kenxu95, line 137: nit: _dict?

src/python/aqueduct_executor/operators/connectors/data/s3.py
kenxu95: Yeah I'm also pretty confused. What do you mean searching over the parent directly? Aren't we just querying the folder directly, as you said in the commend above?

Fanjia-Yan: I think I know what you mean. Sorry for the confusion, let me try if I can get a cleaner way to do this.

Fanjia-Yan:
After some experimentation, this can be really simple. `self.s3.Bucket(self.bucket).objects.filter(Prefix=path)` will return empty with non-existing file path. It will return only one entry with key `directory_name\` upon empty directory.

integration_tests/sdk/data_integration_tests/s3_test.py
```diff
@@ -422,6 +422,10 @@ def test_s3_basic_fetch_failure(client, data_integration):
     with pytest.raises(AqueductError, match="The specified key does not exist."):
         data_integration.file("asdlkf", artifact_type=ArtifactType.TABLE, format="parquet")
 
+    # Fetch a path to directory that does not exist will fail.
+    with pytest.raises(AqueductError, match="Given path to S3 directory 'asdlkf/' does not exist."):
+        data_integration.file("asdlkf/", artifact_type="bytes")
+
```
kenxu95, line 428: Do we have a test case for an empty directory? If not we should add it.

kenxu95 (commented):
Just one last testing request, but the change looks good!

integration_tests/sdk/data_integration_tests/s3_test.py
Fanjia-Yan: I think this piece of testing might be a little bit tricky to setup. If I understand correctly, we set up S3 in `setup_integration.py` by publishing a "save" flow that save all tables from demo_db to s3 storage. However, in order to create a folder, we need a s3 client which I think has been abstracted away in `aqueduct executor`. 

A way I think we can do it is create a S3 client using config info inside `setup_integration.py` and use that to achieve the purpose. I am not sure if that's ok within the testing?

kenxu95: Actually I think we can write the following test case:
```
1) Create an object at some_dir/object
2) Delete the workflow, and include that object in the deletion.
3) some_dir/ should now be an empty directory.
```
This will also help us clarify what our deletion logic too.

Fanjia-Yan: I have included that into the test but there are some discrepancies in test coverage:
 
Behavior 1(Integration Test):
- save an object to path `new_directory\obj_name`, publish the workflow, delete the flow with saved object
- The object is deleted as well as the directory, which we prompted to user `directory does not exist..`

Behavior 2(tried that on jupyter):
- create an empty folder on S3 console with name `existing_directory`
- save an object to path `existing_directory\obj_name`, publish the workflow, delete the flow with saved object
- The directory exists and is an empty directory

I think this test is still good to include but it seems like there is a difference between console created folder and folder created by creating objects.

kenxu95: Oh hmm that is kind of unexpected... why are we deleting the parent directory? @cw75 do you know?

integration_tests/sdk/data_integration_tests/s3_test.py
```diff
@@ -292,6 +295,53 @@ def test_s3_fetch_directory_with_merge(flow_manager, data_integration):
...
+    assert dir_contents.get()[0].equals(hotel_reviews.get())
+
+    # Delete the workflow and include artifact into deletion.
+    tables = flow.list_saved_objects()
+    flow_manager._client.delete_flow(
+        flow_id=str(flow.id()),
+        saved_objects_to_delete=tables,
+        force=True,
+    )
+    # Check if the directory is an empty directory.
```
kenxu95, line 334: This is the unexpected behavior right? If behavior is unexpected let's make a task for it and document that task here with `#TODO(ENG-....): `

kenxu95 (approved):
Unblocking, but I'd really like to get answer to the unexpected directory deletion behavior before you merge. I'm leaving it up you to drive this Fanjia, but please ask for help if you can't figure it out :) 

This isn't going to make this week's release anyways, so there's no rush to merge immediately.

vsreekanti:
@Fanjia-Yan @kenxu95, what's the status of this task?

Fanjia-Yan:
> @Fanjia-Yan @kenxu95, what's the status of this task?

Will work on this and get it in today.

PR #1072, opened 2023-03-10
Fix bug where calling an operator fixes it's execution mode

cw75:
Fixed by moving the global_config check into the execution path, instead of the decorator instantiation path.

### DIFF
--- a/integration_tests/sdk/aqueduct_tests/lazy_execution_test.py
+++ b/integration_tests/sdk/aqueduct_tests/lazy_execution_test.py
@@ -4,6 +4,7 @@ import pandas as pd
 import pytest
 from aqueduct.artifacts.bool_artifact import BoolArtifact
 from aqueduct.artifacts.generic_artifact import GenericArtifact
+from aqueduct.artifacts.numeric_artifact import NumericArtifact
 from aqueduct.constants.enums import ArtifactType
 from aqueduct.error import InvalidUserArgumentException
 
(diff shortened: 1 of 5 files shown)


sdk/aqueduct/decorator.py
```diff
@@ -1017,3 +1024,10 @@ def to_operator(
         ),
     )
     return func_op(func)
+
+
+def _get_execution_mode() -> ExecutionMode:
```
kenxu95, line 1029: nit: `get_global_execution_mode()`?

kenxu95 (approved):

integration_tests/sdk/aqueduct_tests/lazy_execution_test.py
```diff
@@ -238,3 +239,21 @@ def copy_field(df):
...
+        return 1
+
+    result = number()
+    assert isinstance(result, NumericArtifact)
+    assert result.get() == 1
+
+    global_config({"lazy": True})
+
+    result = number()
+    assert isinstance(result, GenericArtifact)
```
kenxu95, line 258: btw the idea is that the data field isn't set on this GenericArtifact right? Until get() is called.

cw75: if the mode is lazy, `result` will be of type `GenericArtifact` because we don't know the exact type yet. This won't be the case when we're evaluating it eagerly.

PR #1077, opened 2023-03-11
Allow users to create table and image parameters from local data

Fanjia-Yan:
This PR adds feature to use local data by creating a parameter `data = client.create_param("data", default="path/to/data.csv", as_type=ArtifactType.TABLE, format="csv")`. The created param will behave normally as a parameter artifact. 

The current size of local data is set to have a maximum of 32MB. The reason for limiting the local data is because our server has a default limit for parsing http request.

## Testing
Add 3 Test suite of integration tests:
1. Basic Functionality of using table local data: Publishing workflow using parameter from local csv, parquet, and json file
2. Basic Functionality of using image local data: Publishing workflow using parameter from local image
3. Error checking test: Check if invalid file path, mismatch artifact type can be detected

## Existing Problem
Currently previewing and publishing workflow using local data parameter is extremely slow around (10 - 20 seconds) for a small(1 MB) of data. We have figured out that was due to serializing using json on Jupiter and json unmarshall the data on server. 

https://linear.app/aqueducthq/issue/ENG-2584/allow-users-to-create-table-and-image-parameters-from-local-data


integration_tests/sdk/data/hotel_reviews.csv
```diff
@@ -0,0 +1,6 @@
+hotel_name,review_date,reviewer_nationality,review
```
kenxu95, line 1: Are these tables supposed to be so small?

sdk/aqueduct/utils/local_data.py
```diff
@@ -0,0 +1,71 @@
...
+    artifact_type: ArtifactType,
+    format: Optional[str] = None,
+) -> LocalData:
+    """Identify the local data which can be passed in as a parameter.
+
+    Args:
+        path:
+            The path to the data.
+        artifact_type:
+                The expected type of the data. The `ArtifactType` class in `enums.py` contains all
```
kenxu95, line 24: This is a user-facing function. Let's not point them to code. Let's just document the supported possibilities right now: ArtifactType.TABLE, ArtifactType.IMAGE

sdk/aqueduct/utils/local_data.py
```diff
@@ -0,0 +1,71 @@
...
+                supported types, except for ArtifactType.UNTYPED.
+        format:
+                If the artifact type is ArtifactType.TABLE, the user has to specify the table format.
+                We currently support JSON, CSV, and Parquet.
```
kenxu95, line 28: Lets put quotes around "JSON", "CSV", etc. so user's know they can just copy those strings.

sdk/aqueduct/utils/local_data.py
```diff
@@ -0,0 +1,71 @@
...
+
+    Returns
+        A `LocalData` object which contains necessary information to create a parameter artifact.
```
kenxu95, line 31: nit: "A configuration object which you can use to reference data in the local filesystem."

sdk/aqueduct/utils/local_data.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+
+    format_enum = _convert_to_local_data_table_format(format)
+    return LocalData(path=path, artifact_type=artifact_type, format=format_enum)
+
+
+def validate_local_data(val: LocalData) -> None:
```
kenxu95, line 38: This validation should be happening in the constructor of LocalData. The extraction of the data itself can happen at create_param() time.

sdk/aqueduct/utils/serialization.py
```diff
@@ -296,3 +347,38 @@ def artifact_type_to_serialization_type(
         "Unimplemented case for artifact type `%s`" % artifact_type
     )
     return serialization_type
+
+
+def extract_val_from_local_data(val: LocalData) -> Any:
+    """Extract value of spcified type in LocalData."""
```
kenxu95, line 353: nit: specified

sdk/aqueduct/artifacts/create.py
```diff
@@ -111,4 +113,10 @@ def create_param_artifact(
         ],
     )
 
-    return to_artifact_class(dag, output_artifact_id, artifact_type, default)
+    if isinstance(default, LocalData):
```
kenxu95, line 116: I'm really confused as to why we're not doing this in `client.create_param()`. And why we also have such an `isinstance()` check in `construct_param_spec()`. Let's extract the local data in one place and then feed it into `create_param_artifact` no?

kenxu95 (changes requested):
Can you create a follow-up task for making this more performant and solving the sizing bug? 1MB is a pretty restrictive bar.

integration_tests/sdk/aqueduct_tests/param_test.py
```diff
@@ -14,7 +15,7 @@
 from pandas._testing import assert_frame_equal
 from PIL import Image
 
-from aqueduct import metric, op
+from aqueduct import Local_Data, metric, op
```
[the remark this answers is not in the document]
Fanjia-Yan: I have been thinking about this for a while and might need your opinion. So right now, I have `aqueduct.local_data(...)` which creates a `LocalData` class object. The function is basically just a wrapper plus some argument validations. Should we only have the `LocalData` class and put the argument validation in the pydantic base class. When we are designing this we are using `aqueduct.LocalData()` but I am not too sure if we should do everything in the Class Object.

integration_tests/sdk/aqueduct_tests/param_test.py
```diff
@@ -586,3 +587,75 @@ def bar(input):
...
+
+        if extension == "csv":
+            input_df = pd.read_csv("data/hotel_reviews." + extension)
+        elif extension == "json":
+            input_df = pd.read_json("data/hotel_reviews." + extension, orient="table")
+        else:
+            input_df = pd.read_parquet("data/hotel_reviews." + extension)
+        assert input_df.equals(data_param.get())
+
+        output = append_row_to_df(data_param, row_to_add)
```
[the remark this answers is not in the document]
Fanjia-Yan: `use_local` is in another linear task https://linear.app/aqueducthq/issue/ENG-2586/add-use-local-flag-to-clientpublish-flow which I think I will do it right after. Should this task come in as a part of this PR?

integration_tests/sdk/data/hotel_reviews.csv
Fanjia-Yan: I will try to put a larger table and see how it goes against test.

sdk/aqueduct/artifacts/create.py
Fanjia-Yan: Yeah, I think putting it at` client.create_param()` save a lot of unnecessary type checking, there is no reason we should put it in `construct_param_spec` Should be fixed here!

integration_tests/sdk/aqueduct_tests/param_test.py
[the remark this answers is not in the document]
kenxu95: Hmm I think we should make our own LocalData class. It can use pydantic within itself, but I think some of our requirements are not so easily enforced in pydantic. For example: the table format should only be supplied if the type is ABLE.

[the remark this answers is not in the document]
kenxu95: A function is also fine... but I like the idea of a class more actually.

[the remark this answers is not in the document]
Fanjia-Yan: Done!

Fanjia-Yan:
> @Fanjia-Yan just confirming that this is still being worked on and doesn't require a review from me yet

There is one thing I would like to try before requesting a review. The File limit is currently set to 1MB which I am getting more and more disturbed by the value and I hope to get this thing straight before merging into main. I observe that the when we are parsing http request from server, large json file(>32MB), the parsing is faulty but no error is returned. This might be some issue of max file size we can set or work around in golang http request library. I am still investigating that.
Screen Shot 2023-03-21 at 3 20 00 PM

sdk/aqueduct/backend/api_client.py
```diff
@@ -403,6 +406,14 @@ def preview(
             "dag": dag.json(exclude_none=True),
         }
 
+        if len(body["dag"]) > MAX_REQUEST_BODY_SIZE and any(
```
Fanjia-Yan, line 409: The reason have a check on the size of dag is that there might be multiple `LocalData` in one workflow/preview which causes the http request parsing to error out.

src/golang/cmd/server/request/payload.go
```diff
@@ -55,7 +55,10 @@ func ExtractHttpPayload(contentType, fileName string, isFile bool, r *http.Reque
 		} else {
 			// If we reach here, it means the value is sent as `String` instead of `File`,
 			// so we use `FormValue` to extract the payload and return its byte form.
-			// The caller can later parse the returned value to string.
+			// The caller can later parse the returned value to string. By default, the
```
Fanjia-Yan, line 58: A reference to how I derive this kind of behavior is from https://cs.opensource.google/go/go/+/refs/tags/go1.20.2:src/net/http/request.go;l=1325
https://cs.opensource.google/go/go/+/refs/tags/go1.20.2:src/mime/multipart/formdata.go;l=31

Essentially, the behavior of `FormValue` is that:
1. request.body size < 42MB, everything is good 
2. request.body size > 42MB, Reading from request is not successful->empty string is returned-> no error is raised-> `json.unmarshal` on empty string -> error.

I am not sure the necessity of taking care of second case, since I am not sure if we have any instance where size of dag is larger than the default 42MB besides `localData`.

src/golang/cmd/server/request/payload.go
```diff
@@ -55,7 +55,10 @@ func ExtractHttpPayload(contentType, fileName string, isFile bool, r *http.Reque
...
+			// request's body size is capped at `defaultMaxMemory = 32 << 20`. In addition
+			// `multipartReader` gives an addition of 10MB upon defaultMaxMemory for
+			// nonfile data. Therefore this adds up to a maximum request body size of 42MB.
```
Fanjia-Yan, line 61: A reference to how I derive this kind of behavior is from https://cs.opensource.google/go/go/+/refs/tags/go1.20.2:src/net/http/request.go;l=1325
https://cs.opensource.google/go/go/+/refs/tags/go1.20.2:src/mime/multipart/formdata.go;l=31

Essentially, the behavior of `FormValue` is that:
1. request.body size < 42MB, everything is good 
2. request.body size > 42MB, Reading from request is not successful->empty string is returned-> no error is raised-> `json.unmarshal` on empty string -> error.

I am not sure the necessity of taking care of second case, since I am not sure if we have any instance where size of dag is larger than the default 42MB besides `localData`.

sdk/aqueduct/backend/api_client.py
```diff
@@ -40,6 +41,8 @@
     SavedObjectUpdate,
 )
 
+MAX_REQUEST_BODY_SIZE = 32 << 20
```
kenxu95, line 44: Please document this. What's are the units? What is the human-understandable value?

sdk/aqueduct/backend/api_client.py
```diff
@@ -403,6 +406,14 @@ def preview(
             "dag": dag.json(exclude_none=True),
         }
 
+        if len(body["dag"]) > MAX_REQUEST_BODY_SIZE and any(
+            artifact_metadata.from_local_data for artifact_metadata in list(dag.artifacts.values())
+        ):
+            raise InvalidUserActionException(
+                "Local Data after serialization are too large. Aqueduct uses json serialization. The maximum size of workflow with local data is %sMB, the current size is %sMB."
```
kenxu95, line 413: nit: "Local data after serialization is too large".

kenxu95 (commented):
I think this is good! I do think this means we have add pyarrow as a requirement then to our SDK right? Want to get Chengang's confirmation on this, but could you add this requirement to our SDK? We're going to need to do it when we transition to parquet from json anyways.

kenxu95 (approved):

sdk/aqueduct/client.py
```diff
@@ -188,19 +196,39 @@ def create_param(self, name: str, default: Any, description: str = "") -> BaseAr
...
             default:
                 The default value to give this parameter, if no value is provided.
-                Every parameter must have a default.
+                Every parameter must have a default. If decided to use local data,
+                a path to the local data file must be specified.
             description:
                 A description of what this parameter represents.
-
+            use_local:
+                whether this parameter uses local data source or not.
```
kenxu95, line 204: Please capitalize.

sdk/aqueduct/client.py
```diff
@@ -178,7 +178,15 @@ def github(self, repo: str, branch: str = "") -> Github:
...
         return Github(repo_url=repo, branch=branch)
 
-    def create_param(self, name: str, default: Any, description: str = "") -> BaseArtifact:
+    def create_param(
+        self,
+        name: str,
+        default: Any,
+        description: str = "",
+        use_local: bool = False,
+        as_type: Optional[ArtifactType] = None,
```
kenxu95, line 187: Can we just call this `as`?

sdk/aqueduct/client.py
```diff
@@ -188,19 +196,39 @@ def create_param(self, name: str, default: Any, description: str = "") -> BaseAr
...
         Returns:
             A parameter artifact.
         """
+        if use_local:
+            try:
+                assert type(default) == str
+            except AssertionError:
+                raise InvalidUserArgumentException(
+                    "the default value must be a path to local data."
+                )
```
kenxu95, line 220: This is really roundabout, and is equivalent to `if not isinstance(default): raise InvalidUser...` Let's rewrite it.

sdk/aqueduct/client.py
```diff
@@ -188,19 +196,39 @@ def create_param(self, name: str, default: Any, description: str = "") -> BaseAr
                 The name to assign this parameter.
             default:
                 The default value to give this parameter, if no value is provided.
-                Every parameter must have a default.
+                Every parameter must have a default. If decided to use local data,
```
kenxu95, line 199: grammer: "if we decide"

sdk/aqueduct/client.py
```diff
@@ -188,19 +196,39 @@ def create_param(self, name: str, default: Any, description: str = "") -> BaseAr
...
+                a path to the local data file must be specified.
             description:
                 A description of what this parameter represents.
-
+            use_local:
+                whether this parameter uses local data source or not.
+            as_type:
+                The expected type of the local data. Currently Local Data has support for ArtifactType.TABLE
```
kenxu95, line 206: Can we replace the second sentence with "Only supported types are ArtifactType.TABLE and ArtifactType.IMAGE

sdk/aqueduct/client.py
```diff
@@ -188,19 +196,39 @@ def create_param(self, name: str, default: Any, description: str = "") -> BaseAr
...
             A parameter artifact.
         """
+        if use_local:
+            try:
+                assert type(default) == str
+            except AssertionError:
+                raise InvalidUserArgumentException(
+                    "the default value must be a path to local data."
+                )
+            default = extract_val_from_local_data(default, as_type, format)
```
kenxu95, line 221: Can you check that the `as` argument is one of the two supported types? And throw an exception if it isn't?

sdk/aqueduct/client.py
```diff
@@ -188,19 +196,39 @@ def create_param(self, name: str, default: Any, description: str = "") -> BaseAr
...
+            is_local_data = True
+        else:
+            is_local_data = False
         return create_param_artifact(
             self._dag,
             name,
             default,
             description,
             explicitly_named=True,
+            is_local_data=is_local_data,
```
kenxu95, line 231: Can't we just use `use_local` here?

sdk/aqueduct/client.py
```diff
@@ -188,19 +196,39 @@ def create_param(self, name: str, default: Any, description: str = "") -> BaseAr
...
                 A description of what this parameter represents.
-
+            use_local:
+                whether this parameter uses local data source or not.
+            as_type:
+                The expected type of the local data. Currently Local Data has support for ArtifactType.TABLE
+                and ArtifactType.IMAGE.
+            format:
+                If local data type is ArtifactType.TABLE, the user has to specify the table format.
+                We currently support "JSON", "CSV", and "Parquet".
```
kenxu95, line 210: Let's keep the captialization consistent between the three examples: eg: "json", "csv", "parquet"

kenxu95 (changes requested):
It's very close, but not quite production quality yet. The big issue I have is with the overly aggressive restriction on `pyarrow` version. Let's always set this to the latest if we can get away with it.

sdk/aqueduct/client.py
Fanjia-Yan: I think `as` is a python keyword, overriding it with a variable might be a bit tricky. Will something like `_as` or `astype` suffice?

sdk/aqueduct/client.py
Fanjia-Yan: Added in` validate_local_data` in `utils/local_data.py`

sdk/aqueduct/client.py
```diff
@@ -188,19 +197,34 @@ def create_param(self, name: str, default: Any, description: str = "") -> BaseAr
...
+            format:
+                If local data type is ArtifactType.TABLE, the user has to specify the table format.
+                We currently support "json", "csv", and "parquet".
         Returns:
             A parameter artifact.
         """
+        if use_local:
+            if not isinstance(default, str):
+                raise InvalidUserArgumentException(
+                    "the default value must be a path to local data."
```
kenxu95, line 217: super nit: Capitalize "the"

kenxu95 (approved):
Cool! lgtm just one super nit :)

PR #1079, opened 2023-03-13
ENG 2410 improve metadata migration notice looking

likawind:
This PR improves the rendering of metadata migration alerts by properly assign width and margin to this component.

ENG-2410

## Loom demo (if any)
Before:
image

After:
Screenshot 2023-03-12 at 6 27 16 PM

### DIFF
--- a/src/ui/common/src/components/integrations/addIntegrations.tsx
+++ b/src/ui/common/src/components/integrations/addIntegrations.tsx
@@ -38,7 +38,7 @@ const AddIntegrations: React.FC<Props> = ({
             setShowMigrationDialog(false);
           }}
           severity="info"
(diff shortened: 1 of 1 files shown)


vsreekanti:
Is there a max width on the container that has the logos? If so, can we set it to be the same as that?

likawind:
@vsreekanti the container size is dynamic and we always put as many logos as possible in each row. The run over is caused by the remainder space in container that doesn't fit a logo. We do ensure the alert is the same size as the container, but I'm not sure if there's an easy way to compute the size of all logos as it's not an explicit value:

Screenshot 2023-03-13 at 5 12 43 PM

agiron123 (approved):

PR #1092, opened 2023-03-15
Allow creating multiple AWS integrations

cw75:
This PR enables users to create multiple AWS integrations. Each integration has its own Terraform directory that manages a cluster.


src/golang/cmd/server/server/execute_handler.go
```diff
@@ -25,15 +25,7 @@ func HandleError(
...
-	var externalMsg string
-	dbxErr, ok := err.(errors.DropboxError)
-	if ok {
-		// Only return the top-level error message to clients.
-		externalMsg = dbxErr.GetMessage()
-	} else {
-		externalMsg = err.Error()
-	}
-	response.SendErrorResponse(w, externalMsg, statusCode)
+	response.SendErrorResponse(w, errors.GetMessage(err), statusCode)
```
cw75, line 28: @kenxu95 I found that `errors.GetMessage(err)` has the same functionality as the original code, with the added benefit that it gets errors at all levels instead of just the top-level, which I found to be much more useful both to ourselves and to the users.

sdk/aqueduct/client.py
```diff
@@ -427,7 +427,7 @@ def integration(
                 metadata=integration_info,
             )
         elif integration_info.service == ServiceType.AWS:
-            dynamic_k8s_integration_name = "%s:k8s" % name
+            dynamic_k8s_integration_name = "%s:aqueduct_ondemand_k8s" % name
```
kenxu95, line 430: Curious why this change?

src/golang/cmd/executor/executor/dynamic_teardown.go
```diff
@@ -31,52 +33,62 @@ func (ex *DynamicTeardownExecutor) Run(ctx context.Context) error {
...
+						dynamicIntegration,
+						ex.IntegrationRepo,
+						ex.Database,
+					); err != nil {
+						log.Error(errors.Wrap(err, "Unable to delete dynamic k8s integration"))
+						return
+					}
+				} else {
+					log.Info("Have not reached keepalive threshold, not tearing down the cluster.")
+				}
```
kenxu95, line 87: Now that it's a goroutine, should we have some timeout mechanism? Actually, was there a timeout mechanism before, or do our executors just run unbounded.

src/golang/cmd/server/handler/cloud_integration_utils.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/aqueducthq/aqueduct/lib/dynamic"
+	"github.com/aqueducthq/aqueduct/lib/errors"
+	"github.com/aqueducthq/aqueduct/lib/lib_utils"
+	"github.com/aqueducthq/aqueduct/lib/models/shared"
+	"github.com/aqueducthq/aqueduct/lib/workflow/operator/connector/auth"
+	"github.com/google/uuid"
+)
+
+// setupCloudIntegration sets up the cloud integration's Terraform directory, registers a k8s
+// integration and run `terraform init` to initialize the Terraform workspace.
```
kenxu95, line 21: Can you add a note here about where this terraform directory exists?

src/golang/cmd/server/server/execute_handler.go
kenxu95: Actually since this is such a useful and scoped changed, can you make a new PR with just this? I can stamp it immediately (might want it for my error handling work)

src/golang/lib/dynamic/engine.go
```diff
@@ -30,9 +32,12 @@ const (
 	K8sClusterUpdateAction k8sClusterActionType = "update"
 )
 
-const stateLockErrMsg = "Error acquiring the state lock"
+const (
+	stateLockErrMsg      = "Error acquiring the state lock"
+	K8sClusterNameSuffix = "aqueduct_ondemand_k8s"
```
kenxu95, line 37: This is the name of the K8s integration right? Not the eks cluster name.

kenxu95:
Looks good but left a couple questions!

sdk/aqueduct/client.py
cw75: just to prevent this name from colliding with user-defined names. `aqueduct_ondemand_k8s` is much less likely to collide than `k8s`

src/golang/cmd/executor/executor/dynamic_teardown.go
cw75: Our executor runs unbounded, but in this case Terraform itself has timeout built in so we won't run forever even during the error case.

src/golang/cmd/server/handler/cloud_integration_utils.go
cw75: added a note to `setupTerraformDirectory`

src/golang/lib/dynamic/engine.go
cw75: yeah, updated to `K8sIntegrationNameSuffix`

cw75:

src/golang/cmd/executor/executor/dynamic_teardown.go
kenxu95: got it. Can you note that at the beginning of the goroutine? It's one of the first things that comes to my mind when reading unbounded goroutines (usually I've been taught to use ErrGroups where all goroutines share the same context).

kenxu95 (approved):

PR #1093, opened 2023-03-15
Update the naming scheme for artifacts in our SDK

kenxu95:
Implements what we have been discussing for the last week about naming in the SDK. There are two other decisions I made in this PR that I want to surface:
- Metrics and checks are special in that we used to be able to edit them. That wouldn't work with a naive record-everything approach. Moreover, metrics and checks are implicitly included by default in a published flow, so having extraneous copies of metrics/checks floating around probably isn't a good idea. The solution in this PR is to allow operator replacement for this particular edge case. If a user defines a metric/check with the same name on the **same** artifact twice, then we will delete the original one.
- This PR sanitizes our artifact names by stripping out any leading or trailing whitespace immediately. This seemed like a sensible restriction, and also makes the suffix bumping code slightly simpler. 

ENG-2520

## Loom demo (if any)
(I said its a quick demo but it's not really that quick)
https://www.loom.com/share/bcb12cf71b0a4c5d97ca5fd84fff2fae


integration_tests/sdk/aqueduct_tests/checks_test.py
```diff
@@ -84,27 +89,43 @@ def test_check_on_multiple_mixed_inputs(client, flow_name, data_integration, eng
...
+    assert not fail.get()
+
+    flow = publish_flow_test(client, artifacts=[pass2, fail], engine=engine, name=flow_name())
+    flow_run = flow.latest()
+    assert flow_run.artifact("foo artifact").get()
+
+    # TODO(2629): Fetching a warning check should not raise an exception.
+    with pytest.raises(
+        ArtifactNeverComputedException,
+        match="This artifact was part of an existing flow run but was never computed successfully",
```
cw75, line 126: why will this raise this exception?

integration_tests/sdk/aqueduct_tests/metrics_test.py
```diff
@@ -1,6 +1,6 @@
 import pandas as pd
 import pytest
-from aqueduct.error import AqueductError
+from aqueduct.error import AqueductError, ArtifactNeverComputedException, ArtifactNotFoundException
```
cw75, line 3: do we need "ArtifactNeverComputedException"?

integration_tests/sdk/aqueduct_tests/naming_test.py
```diff
@@ -33,87 +67,258 @@ def test_extract_with_default_name_collision(client, data_integration):
...
-        "reviewer_nationality",
-        "review",
-        "newcol",
-    ]
-    assert fn_df.shape[0] == 100
+    @op
+    def foo():
+        return 123
+
+    @op(outputs=["foo artifact"])
```
cw75, line 184: If I explicitly name my artifact sth with a suffix "foo (1)", does the naming resolution still work? Unlikely to happen but just curious.

cw75 (commented):
@kenxu95 LGTM! Left a few questions and comments.

sdk/aqueduct/models/dag.py
```diff
@@ -47,21 +50,113 @@ class DAG(BaseModel):
...
+
+        # For some determinism around naming, we sort the starting operators by name and then id.
+        starting_ops: List[Operator] = sorted(
+            [op for op in self.operators.values() if len(op.inputs) == 0],
+            key=lambda op: (op.name, op.id),
+        )
+
+        seen_artifact_names: Set[str] = set()
+        self._validate_explicitly_named_artifacts(starting_ops, seen_artifact_names)
+        self._resolve_artifact_names(starting_ops, seen_artifact_names)
```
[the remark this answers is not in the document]
kenxu95: The main reason is that we want the first pass to only handle explicitly named artifacts so that those artifacts can claim their names first. Now that I think about it, only the second pass needs to be topologically ordered, since that's the one that's bumping the suffix. Can fix.

sdk/aqueduct/models/dag.py
```diff
@@ -127,44 +222,71 @@ def set_engine_config(
...
+        match = [
+            op
+            for op in self.operators.values()
+            if (
+                get_operator_type(op) == get_operator_type(candidate_op)
+                and set(op.inputs) == set(candidate_op.inputs)
+                and op.name == candidate_op.name
+            )
+        ]
+        assert len(match) <= 2, (
```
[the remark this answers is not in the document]
kenxu95: yes nice catch :)

sdk/aqueduct/models/dag.py
```diff
@@ -329,35 +451,24 @@ def update_artifact_type(self, artifact_id: uuid.UUID, artifact_type: ArtifactTy
...
 
     def update_artifact_name(self, artifact_id: uuid.UUID, new_name: str) -> None:
-        self.must_get_artifact(artifact_id).name = new_name
-
-    def update_operator_name(self, op_id: uuid.UUID, new_name: str) -> None:
-        # Update the name -> operator map.
-        old_name = self.must_get_operator(op_id).name
-        self.operator_by_name[new_name] = self.operator_by_name[old_name]
-        del self.operator_by_name[old_name]
+        """Assumption: the artifact being altered is being explicitly named."""
```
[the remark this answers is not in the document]
kenxu95: Yep! Lemme rewrite the comment to be more clear.

integration_tests/sdk/aqueduct_tests/checks_test.py
kenxu95: It shouldn't. That's what ENG-2629 addresses. For some reason, a failed warning check is hidden by the backend, I believe.

integration_tests/sdk/aqueduct_tests/naming_test.py
kenxu95: Yeah it should! See `test_bump_artifact_suffix`. Just to be clear though, are you trying to collide an artifact explicitly named `foo artifact (1)` with another one that is implicitly named `foo artifact (1)`? I can test it out but that's a three artifact situation.

integration_tests/sdk/aqueduct_tests/checks_test.py
cw75: Oh interesting, seems bad!

sdk/aqueduct/models/dag.py
```diff
@@ -451,7 +418,9 @@ def update_artifact_type(self, artifact_id: uuid.UUID, artifact_type: ArtifactTy
         self.must_get_artifact(artifact_id).type = artifact_type
 
     def update_artifact_name(self, artifact_id: uuid.UUID, new_name: str) -> None:
-        """Assumption: the artifact being altered is being explicitly named."""
+        """Updates an artifact to have a user-specified name. THis means the artifact is
```
cw75, line 421: typo: THis @kenxu95

integration_tests/sdk/aqueduct_tests/naming_test.py
cw75: Yeah that's what I was thinking. Will a explicitly named `foo artifact (1)` and our suffix bumping logic to generate an implicit name `foo artifact (1)` work, and yes it's a 3 artifact situation.

cw75 (approved):
Unblocking but spotted a typo, and if you can add the 3 artifact naming test case that'd be great!

PR #1095, opened 2023-03-16
Add `use_local` flag to client.publish_flow()

Fanjia-Yan:
Change the `publish_flow()`, add a boolean parameter `use_local` such that when use_local is `True`, the user can publish workflow using local data. `use_local` is default to `False`. When it is `False`, the `publish_flow()` will error out.

https://linear.app/aqueducthq/issue/ENG-2586/add-use-local-flag-to-clientpublish-flow

### DIFF
--- a/integration_tests/sdk/aqueduct_tests/param_test.py
+++ b/integration_tests/sdk/aqueduct_tests/param_test.py
@@ -625,17 +625,14 @@ def test_local_table_data_parameter(client, flow_name, engine):
         output_artifact_list.append(output)
         input_data_list.append(input_df)
 
-    with pytest.raises(
-        InvalidUserActionException,
-        match="Cannot create a flow with local data."
-    ):
+    with pytest.raises(InvalidUserActionException, match="Cannot create a flow with local data."):
         flow = client.publish_flow(
             name=flow_name(),
             artifacts=output_artifact_list,
             engine=engine,
         )
     flow = publish_flow_test(
-        client, artifacts=output_artifact_list, name=flow_name(), engine=engine, use_local = True
+        client, artifacts=output_artifact_list, name=flow_name(), engine=engine, use_local=True
     )
     flow_run = flow.latest()
     assert flow_run.artifact("output_csv").get().equals(input_data_list[0])
@@ -684,5 +681,5 @@ def test_local_image_data_parameter(client, flow_name, engine):
             image_output,
         ],
         engine=engine,
-        use_local = True
+        use_local=True,
     )
--- a/integration_tests/sdk/aqueduct_tests/param_test.py
+++ b/integration_tests/sdk/aqueduct_tests/param_test.py
@@ -629,8 +629,10 @@ def test_local_table_data_parameter(client, flow_name, engine):
         InvalidUserActionException,
         match="Cannot create a flow with local data."
     ):
-        flow = publish_flow_test(
-            client, artifacts=output_artifact_list, name=flow_name(), engine=engine
+        flow = client.publish_flow(
+            name=flow_name(),
+            artifacts=output_artifact_list,
+            engine=engine,
         )
     flow = publish_flow_test(
         client, artifacts=output_artifact_list, name=flow_name(), engine=engine, use_local = True
--- a/integration_tests/sdk/aqueduct_tests/param_test.py
+++ b/integration_tests/sdk/aqueduct_tests/param_test.py
@@ -625,8 +625,15 @@ def test_local_table_data_parameter(client, flow_name, engine):
         output_artifact_list.append(output)
         input_data_list.append(input_df)
 
+    with pytest.raises(
+        InvalidUserActionException,
+        match="Cannot create a flow with local data."
+    ):
+        flow = publish_flow_test(
+            client, artifacts=output_artifact_list, name=flow_name(), engine=engine
+        )
     flow = publish_flow_test(
-        client, artifacts=output_artifact_list, name=flow_name(), engine=engine
+        client, artifacts=output_artifact_list, name=flow_name(), engine=engine, use_local = True
     )
     flow_run = flow.latest()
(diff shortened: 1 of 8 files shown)


integration_tests/sdk/aqueduct_tests/param_test.py
```diff
@@ -625,8 +625,15 @@ def append_row_to_df(df, row):
         output_artifact_list.append(output)
         input_data_list.append(input_df)
 
+    with pytest.raises(
+        InvalidUserActionException,
+        match="Cannot create a flow with local data."
+    ):
+        flow = publish_flow_test(
+            client, artifacts=output_artifact_list, name=flow_name(), engine=engine
```
kenxu95, line 633: Let's use `client.publish_flow()` if we expect the flow to error before sending anything to the server.

sdk/aqueduct/client.py
```diff
@@ -552,6 +556,9 @@ def publish_flow(
             source_flow:
                 Used to identify the source flow for this flow. This can be identified
                 via an object (Flow), name (str), or id (str or uuid).
+            use_local:
+                Used to identify if local data is used. In order to use `LocalData`, this has to be set
```
kenxu95, line 560: Maybe rephrase: "Must be set if any artifact in the flow is derived from local data."

sdk/aqueduct/artifacts/base_artifact.py
```diff
@@ -14,6 +14,7 @@ class BaseArtifact(ABC):
     _dag: DAG
     _content: Any
     _from_flow_run: bool
+    _from_local_data: bool = False
```
kenxu95, line 17: I would rather have this information on the underlying DAG's ArtifactMetadata object. See how this was done in https://github.com/aqueducthq/aqueduct/pull/1093/files, specifically `sdk/aqueduct/models/artifact.py`. 

The reason being that it's easier to think about all the construction fields are on the same data plane. `_from_flow_run` is different since the underlying DAG is fetched from the server. `_from_operator_type` is in the wrong place imo, so let's not follow that example.

sdk/aqueduct/client.py
```diff
@@ -223,6 +223,9 @@ def create_param(self, name: str, default: Any, description: str = "") -> BaseAr
         check_explicit_param_name(self._dag, name)
         if isinstance(default, LocalData):
             default = extract_val_from_local_data(default)
+            artifact = create_param_artifact(self._dag, name, default, description)
+            artifact._from_local_data = True
```
kenxu95, line 227: Following the advice above, add a `from_local_data` argument to `create_param_artifact()`.

kenxu95 (changes requested):
Lemme know if my comment isn't clear! Almost there though

Fanjia-Yan:
Changes after reviews:

- Move `from_local_data` to ArtifactMetadata field that will be put into dag
- Only mark this field to True when the Artifact is a param artifact with local data input. 
- In `publish_flow`, after we construct the DAG subgraph, if there is any artifact in the subgraph that is a local data, raise Exception when `use_local` parameter is False which is default value.

sdk/aqueduct/models/artifact.py
```diff
@@ -8,3 +8,4 @@ class ArtifactMetadata(BaseModel):
     id: uuid.UUID
     name: str
     type: ArtifactType
+    from_local_data: bool = False
```
kenxu95, line 11: You need to exclude this from the json config. See the reference PR I noted in my previous comment.

src/golang/lib/models/artifact.go
```diff
@@ -20,10 +20,11 @@ const (
...
 type Artifact struct {
-	ID          uuid.UUID           `db:"id" json:"id"`
-	Name        string              `db:"name" json:"name"`
-	Description string              `db:"description" json:"description"`
-	Type        shared.ArtifactType `db:"type" json:"type"`
+	ID            uuid.UUID           `db:"id" json:"id"`
+	Name          string              `db:"name" json:"name"`
+	Description   string              `db:"description" json:"description"`
+	Type          shared.ArtifactType `db:"type" json:"type"`
+	FromLocalData bool                `db:"from_local_data" json:"from_local_data"`
```
kenxu95, line 27: oh we definitely shouldn't be propagating this field to the golang side. This is an SDK-only concept.

kenxu95 (changes requested):

src/golang/lib/models/artifact.go
Fanjia-Yan: Make sense that this is a SDK only object. Should be addressed!

sdk/aqueduct/client.py
```diff
@@ -501,6 +508,7 @@ def publish_flow(
         k_latest_runs: Optional[int] = None,
         config: Optional[FlowConfig] = None,
         source_flow: Optional[Union[Flow, str, uuid.UUID]] = None,
+        use_local: Optional[bool] = False,
```
kenxu95, line 511: You might need to rebase. `list_params()` is gone now.

sdk/aqueduct/client.py
```diff
@@ -552,6 +560,8 @@ def publish_flow(
             source_flow:
                 Used to identify the source flow for this flow. This can be identified
                 via an object (Flow), name (str), or id (str or uuid).
+            use_local:
+                Used to identify if local data is used. Must be set if any artifact in the flow is derived from local data.
```
kenxu95, line 564: nit: Let's scrape the first sentence. The second sentence is very clear already.

kenxu95 (approved):
Yay! Please address the nits before merging, and gl with the rebase let me know if you need help :)

sdk/aqueduct/client.py
Fanjia-Yan: I will be doing the rebase in when `eng-2584` merge into `main` since I will be using from_local_data in that branch as well

PR #1105, opened 2023-03-17
Surface error message to the user when setting the `stream` flag to true

cw75:
Previously, when we set the `stream` flag to true, we are only streaming the error message to the server log but not returning the error to the user. This PR make it so that the error log is also captured and returned to the user (instead of just saying "Error waiting for command to complete" which is not useful).

### DIFF
--- a/src/golang/lib/lib_utils/utils.go
+++ b/src/golang/lib/lib_utils/utils.go
@@ -98,6 +98,7 @@ func RunCmd(command string, args []string, dir string, stream bool) (string, str
 		go func() {
 			var sb strings.Builder
 			for stderrScanner.Scan() {
+				log.Errorf("stderr: %s", stderrScanner.Text())
 				sb.WriteString(stderrScanner.Text())
 				sb.WriteString("\n")
 			}
--- a/src/golang/lib/lib_utils/utils.go
+++ b/src/golang/lib/lib_utils/utils.go
@@ -262,10 +262,12 @@ func ParseAWSConfig(conf auth.Config) (*shared.AWSConfig, error) {
 	}
 
 	var c struct {
-		AccessKeyId     string `json:"access_key_id"`
-		SecretAccessKey string `json:"secret_access_key"`
-		Region          string `json:"region"`
-		K8sSerialized   string `json:"k8s_serialized"`
+		AccessKeyId       string `json:"access_key_id"`
+		SecretAccessKey   string `json:"secret_access_key"`
+		Region            string `json:"region"`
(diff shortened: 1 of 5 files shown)


src/golang/lib/lib_utils/utils.go
```diff
@@ -84,21 +84,33 @@ func RunCmd(command string, args []string, dir string, stream bool) (string, str
...
+
 		// start separate goroutines to stream the output from each scanner
 		go func() {
 			for stdoutScanner.Scan() {
 				log.Infof("stdout: %s", stdoutScanner.Text())
 			}
 		}()
 		go func() {
+			var sb strings.Builder
 			for stderrScanner.Scan() {
```
kenxu95, line 100: I don't know too much about the bufio.Scanner class, but this is guaranteed to return even if nothing is ever logged to stderr? How does it know that...

src/golang/lib/lib_utils/utils.go
```diff
@@ -84,21 +84,33 @@ func RunCmd(command string, args []string, dir string, stream bool) (string, str
...
 				log.Errorf("stderr: %s", stderrScanner.Text())
+				sb.WriteString(stderrScanner.Text())
+				sb.WriteString("\n")
 			}
+			ch <- sb.String()
 		}()
 
-		// wait for the command to complete
+		// Wait for the stderr goroutine to finish and receive the stderr from the channel.
+		stderrMsg = <-ch
```
kenxu95, line 109: hmm I would expect this to block forever until something is written to ch... is something guaranteed to be written?

kenxu95 (commented):
Hmm this channel thing looks a little fishy to me. Does this really work if nothing is ever written to ch?

src/golang/lib/lib_utils/utils.go
cw75: Yes, when the program exits, the scanner will reach the end and break out of the for loop.

src/golang/lib/lib_utils/utils.go
cw75: and yes you can send an empty string to channel and the listener blocked on the channel will get unblocked.

cw75:
@kenxu95 I can write a couple comments clarifying my response if that's helpful.

kenxu95:
@cw75 yeah that would be helpful. I was just confused because:
1) I had no idea when the Scan() would terminate
2) I did not know sending an empty string to the channel would unblock it.

cw75:
@kenxu95 addressed comments.

kenxu95 (approved):

PR #1106, opened 2023-03-17
Allow registering AWS integration via credential file path and profile

cw75:

### DIFF
--- a/sdk/aqueduct/integrations/connect_config.py
+++ b/sdk/aqueduct/integrations/connect_config.py
@@ -184,6 +184,9 @@ class DynamicK8sConfig(BaseConnectionConfig):
 
 
 class AWSConfig(BaseConnectionConfig):
+    # Either 1) all of access_key_id, secret_access_key, region, or 2) both config_file_path and
+    # config_file_profile need to be specified. Any other cases will be rejected by the server's
+    # config validation process.
     access_key_id: str = ""
     secret_access_key: str = ""
     region: str = ""
--- a/sdk/aqueduct/integrations/connect_config.py
+++ b/sdk/aqueduct/integrations/connect_config.py
@@ -184,16 +184,20 @@ class DynamicK8sConfig(BaseConnectionConfig):
 
 
 class AWSConfig(BaseConnectionConfig):
-    access_key_id: str
-    secret_access_key: str
-    region: str
+    access_key_id: str = ""
+    secret_access_key: str = ""
+    region: str = ""
+    config_file_path: str = ""
+    config_file_profile: str = ""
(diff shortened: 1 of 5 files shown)


likawind (approved):

likawind:
Do we need UI change?

cw75:
@likawind nah the UI part is already there it just doesn't work before. :)

sdk/aqueduct/integrations/connect_config.py
```diff
@@ -184,16 +184,20 @@ class Config:
...
 
 class AWSConfig(BaseConnectionConfig):
-    access_key_id: str
-    secret_access_key: str
-    region: str
+    access_key_id: str = ""
+    secret_access_key: str = ""
+    region: str = ""
+    config_file_path: str = ""
+    config_file_profile: str = ""
```
kenxu95, line 191: Oh it would be really helpful to know what fields are required and which ones are optional here. Eg. What happens if all the fields are set? Which configuration is picked? We either set the key/secret directory or we use a profile path right? We should enforce that.

src/golang/lib/models/shared/integration_config.go
```diff
@@ -192,10 +192,12 @@ func (config *DynamicK8sConfig) Update(newConfig *DynamicK8sConfig) {
...
-	AccessKeyId     string            `json:"access_key_id"`
-	SecretAccessKey string            `json:"secret_access_key"`
-	Region          string            `json:"region"`
-	K8s             *DynamicK8sConfig `json:"k8s"`
+	AccessKeyId       string            `json:"access_key_id"`
+	SecretAccessKey   string            `json:"secret_access_key"`
+	Region            string            `json:"region"`
+	ConfigFilePath    string            `json:"config_file_path"`
+	ConfigFileProfile string            `json:"config_file_profile"`
+	K8s               *DynamicK8sConfig `json:"k8s"`
```
kenxu95, line 200: Same comment as the one in the SDK about documentation. I really want to push us to explain our data structures more. It's low-effort and super beneficial.

sdk/aqueduct/integrations/connect_config.py
cw75: @kenxu95 Yeah so this is enforced at the backend's Golang authentication helper. We either require all first 3 fields or the last 2, and any other possibilities are rejected. I think adding the check in the backend is more robust because it guard against both UI, SDK, and other REST calls. Should we just add comments here explaining what we allow or do you want to add the duplicate check here as well?

cw75:
@kenxu95 updated comment to make the validation criteria clear.

PR #1107, opened 2023-03-17
Add a workflow-level error banner on the UI for errors that occur outside the workflow's context

kenxu95:

Also fixes a bug in `get_artifact_result` where the server could panic in certain failure conditions.

This is what it looks like. Test that it works against historical workflows too.
image


src/ui/common/src/components/pages/workflow/id/index.tsx
```diff
@@ -482,6 +482,25 @@ const WorkflowPage: React.FC<WorkflowPageProps> = ({
           </Box>
         )}
 
+        {/*Show any workflow-level errors at the top of the workflow details page.*/}
+        {workflow.selectedResult.exec_state.error !== null && (
```
vsreekanti, line 486: Tiny nit: You can just do `workflow.selectedResult.exec_state.error && (` — Typescript treats `null` and `undefined` as `False`.

src/ui/common/src/components/pages/workflow/id/index.tsx
```diff
@@ -482,6 +482,25 @@ const WorkflowPage: React.FC<WorkflowPageProps> = ({
...
+          <Box
+            sx={{
+              backgroundColor: theme.palette.red[100],
+              color: theme.palette.red[600],
+              p: 2,
+              paddingBottom: '16px',
+              paddingTop: '16px',
+              height: 'fit-content',
+              width: !drawerIsOpen ? `calc(100% - 148px)` : 'calc(100% - 32px)',
```
vsreekanti, line 495: What are these numbers coming from? Presumably they should be `const`s somewhere if they're not already defined?

src/golang/cmd/server/handler/get_workflow.go
```diff
@@ -187,10 +191,28 @@ func (h *GetWorkflowHandler) Perform(ctx context.Context, interfaceArgs interfac
 
 	workflowDagResults := make([]workflowDagResult, 0, len(dagResults))
 	for _, dagResult := range dagResults {
+		var dagExecState shared.ExecutionState
+		if !dagResult.ExecState.IsNull {
```
vsreekanti, line 195: What is the scenario in which the DAG result would have a different status vs. execution state? That seems very strange.

vsreekanti: It also doesn't seem like we're doing anything other than logging that information — is that intentional?

vsreekanti (approved):
Mostly looks good. A few stylistically things inline and a question about error handling on the backend.

src/golang/cmd/server/handler/get_workflow.go
kenxu95: yeah it's mostly just a passive check. It definitely shouldn't happen but we don't technically validate this anywhere. Once https://linear.app/aqueducthq/issue/ENG-2665/remove-redundant-status-field-in-workflowdagresult-responses is addressed we won't need to think about this at all.

src/ui/common/src/components/pages/workflow/id/index.tsx
kenxu95: I think they're hardcoded in this file. Can comment with a more thorough explanation.

src/golang/cmd/server/handler/get_workflow.go
vsreekanti: Got it. Should we tag this to be removed once ENG-2665 is addressed?

kenxu95: yes, can do

kenxu95:
Skipping tests since only a comment added.

PR #1108, opened 2023-03-18
[ENG-2656, WIP] Adds workflow history route, store

vsreekanti:

This PR adds an API route that returns the version history (version ID, timestamp, status) for a particular workflow. This route is hooked up to the version selector on the frontend. This should also be used for the workflow header soon, but I wanted to keep this PR tactical to start since we aren't deprecating any code yet.

This is an example implementation of the types of API routes that I'd like to see us moving towards in the future. I implemented this in about ~2.5 hours including debugging*, which gives me reasonable confidence that these smaller routes are both easier to implement and easier to maintain. 

The full set of new code that's added is ~250 LOC, which includes a bunch of boilerplate, and I think we'll be able to deprecate a lot of the really complex reducers that we have in the codebase right now.

*To be fair, I haven't added any tests.

ENG-2656


src/golang/cmd/server/handler/get_workflow_history.go
```diff
@@ -0,0 +1,119 @@
...
+//	Headers:
+//		`api-key`: user's API Key
+// Response:
+//	Body:
+//		serialized `workflowHistoryResponse`, a list of all the versions of this workflow, their timestamps, and their statuses
+
+type getWorkflowHistoryResponse struct {
+	Id       uuid.UUID                 `json:"id"`
+	Name     string                    `json:"name"`
+	Versions []workflowVersionResponse `json:"versions"`
```
kenxu95, line 32: Since we're documenting the request, it might be good to document our responses too. This route is fairly straightforward but we should definitely cover assumptions like what is the ordering of the versions in the list.

kenxu95 (commented):
Going to leave UI review to Andre. But having a historical version endpoint makes a ton of sense to me! I imagine we'll have a similar one for artifacts too?

src/ui/common/src/components/workflows/version_selector.tsx
```diff
@@ -28,17 +28,27 @@ export const VersionSelector: React.FC = () => {
...
   const selectedResult = workflow.selectedResult;
 
+  const workflowHistory = useSelector(
+    (state: RootState) => state.workflowHistoryReducer
+  );
   const dispatch = useDispatch();
 
   const [menuAnchor, setMenuAnchor] = useState<HTMLButtonElement | null>(null);
-  const [selectedResultIdx, setSelectedResultIdx] = React.useState(0);
+  const [selectedResultIdx, setSelectedResultIdx] = useState(0);
```
agiron123, line 37: nit: `useState<number>(0)`

src/ui/common/src/components/workflows/version_selector.tsx
```diff
@@ -48,7 +58,7 @@ export const VersionSelector: React.FC = () => {
 
       let defaultBackground, hoverBackground, selectedBackground;
 
-      switch (r.status) {
+      switch (r.exec_state.status) {
```
agiron123, line 61: do we need to add null checking for exec_state here? I'm not entirely sure just by looking at the code.

src/ui/common/src/reducers/workflowHistory.ts
```diff
@@ -0,0 +1,55 @@
+import { createSlice } from '@reduxjs/toolkit';
+
+import { handleGetWorkflowHistory } from '../handlers/getWorkflowHistory';
+import { WorkflowHistoryResponse } from '../handlers/responses/workflowHistory';
+import { LoadingStatus, LoadingStatusEnum } from '../utils/shared';
+
+// TODO(vikram): Do we ever need to capture multiple workflows' histories at once?
```
agiron123, line 7: File a linear task here?

agiron123 (approved):
UI Code looks good here and follows the handlers pattern that we're trying to move our requests to.

Left some minor comments, but should be good to go after resolving.

likawind (approved):
@vsreekanti is this for showing the version selector in workflow headers? (nvm, saw this in your PR comment)

likawind:
Another suggestion is to look at this file: https://github.com/aqueducthq/aqueduct/blob/main/src/golang/lib/workflow/dag/response.go there may be a few structs we can reuse here, but not a blocker for this PR

vsreekanti:
Addressed the feedback here. Just waiting for confirmation on the DB guarantees and then will merge.

src/ui/common/src/components/workflows/version_selector.tsx
vsreekanti: Confirmed with Saurav that this is correct.

PR #1110, opened 2023-03-18
Fix a bug where engine icon is not properly shown in list workflow page

likawind:
This PR fixes the engine icon listing issue in list workflow page. Now we properly show the engine icon based on operator engine types together with the dag engine type.

To fix this, we added a new field in `list_workflows` route to include all operator engine types. We added unit test for the corresponding query.

ENG-2633

## Loom demo (if any)
Verified single-engine is shown properly.
Multiple-engine case:
Screenshot 2023-03-17 at 7 27 14 PM


src/golang/lib/repos/tests/operator.go
```diff
@@ -222,7 +222,7 @@ func (ts *TestSuite) TestOperator_GetUnusedCondaEnvNames() {
 	workflows := ts.seedWorkflowWithUser(1, userIDs)
 	workflowIDs := sampleWorkflowIDs(1, workflows)
 	dags := ts.seedDAGWithWorkflow(2, []uuid.UUID{workflowIDs[0], workflowIDs[0]})
-	historicalOp := ts.seedOperatorAndDAG(artifactID, dags[0].ID, users[0].ID, operator.FunctionType)
+	historicalOp := ts.seedOperatorAndDAGOperatorToArtifact(artifactID, dags[0].ID, operator.FunctionType)
```
likawind, line 225: this is related to https://linear.app/aqueducthq/issue/ENG-2667/disable-the-random-behavior-of-seedoperatoranddag

src/ui/common/src/utils/workflows.tsx
```diff
@@ -212,13 +213,21 @@ export function getWorkflowEngineTypes(dag: WorkflowDag): EngineType[] {
...
+
+export function reduceEngineTypes(
+  dagType: EngineType,
+  opTypes: EngineType[]
+): EngineType[] {
   if (
     dagType === EngineType.Aqueduct &&
-    allOpTypes.length > 0 &&
-    allOpTypes.every((t) => t !== EngineType.Aqueduct)
+    opTypes.length > 0 &&
```
agiron123, line 226: Just want to make sure that I'm understanding this right, but this function first filters out all non Aqueduct engines and makes sure that they are all unique (new Set part on line 232), then puts them into an array?

So the final result here is an Array of unique engine types excluding Aqueduct?

src/ui/common/src/components/pages/workflows/index.tsx
```diff
@@ -66,52 +67,58 @@ const WorkflowsPage: React.FC<Props> = ({ user, Layout = DefaultLayout }) => {
...
+      let checks = [];
+      if (workflow.checks) {
+        checks = workflow.checks.map((check) => {
+          const value =
+            check.result?.exec_state.status === 'succeeded' ? 'True' : 'False';
+
+          return {
+            checkId: check.id,
+            name: check.name,
+            level: check.spec?.check?.level ?? CheckLevel.Warning,
```
agiron123, line 99: Are we defaulting to warning level if there is no level defined for the check?

I find that kind of odd, figured most people would go to error, but I suppose this is to keep the workflow from failing if the check fails?

src/ui/common/src/components/pages/workflows/index.tsx
```diff
@@ -66,52 +67,58 @@ const WorkflowsPage: React.FC<Props> = ({ user, Layout = DefaultLayout }) => {
...
-          value: metric.result?.content_serialized ?? '',
-          status: metric.result?.exec_state?.status ?? ExecutionStatus.Unknown,
-        };
-      });
-    }
+  const workflowElements: PaginatedSearchTableRow[] = workflows.map(
+    (workflow) => {
+      const engines = reduceEngineTypes(
+        workflow.engine,
+        workflow.operator_engines.filter((x) => !!x)
```
agiron123, line 74: Does a null/undefined operator engine here imply that the engine is Aqueduct?

src/ui/common/src/utils/workflows.tsx
likawind: yes, the conditioning ensures we show only operator types if no operator runs on aqueduct. This is possible as we are still orchestrating using aqueduct (and the dag type is Aq)

agiron123 (approved):
I left some comments to make sure that I'm properly understanding some frontend code here, but the UI code looks good on my end. You can address these comments and merge when ready.

We could probably use another set of eyes on the backend code just to be sure things are good to go 🙂

src/ui/common/src/components/pages/workflows/index.tsx
likawind: No, that means the engine is inheriting the dag's setting. But there's actually a bug here that we should actually map operator's engine to the dag's engine in the list, so that the conditioning in `reduceEngineTypes` is working properly

src/ui/common/src/components/pages/workflows/index.tsx
likawind: I think this is more a safeguard to show something and the UI doesn't crash. There shouldn't be practical use case where this value is empty. If this actually happens the backend is likely to complain and more places in UI will error.

PR #1113, opened 2023-03-19
ENG-2620 DAG Line Crossing Fix

agiron123:
This PR removes elk and brings back the old positioning algorithm.

Edge crossings should be much less likely, however they might seem a little long when there is a collapsed DAG node.

ENG-2620

## Loom demo (if any)
Screehshots:
image

image

image

### DIFF
--- a/src/ui/common/src/utils/reactflow.ts
+++ b/src/ui/common/src/utils/reactflow.ts
@@ -45,7 +45,7 @@ export function getOperatorNode(
   op: Operator,
   pos: NodePos,
   onChange: () => void,
-  onConnect: (any) => void,
+  onConnect: (any) => void
 ): Node<ReactFlowNodeData> {
   return {
     id: op.id,
@@ -58,8 +58,6 @@ export function getOperatorNode(
       nodeId: op.id,
       label: op.name,
     },
-    // Give an initial position. We will reposition this node later.
-    //position: { x: 0, y: 0 },
     position: pos,
   };
 }
@@ -68,7 +66,7 @@ export function getArtifactNode(
   artf: Artifact,
   pos: NodePos,
   onChange: () => void,
-  onConnect: (any) => void,
+  onConnect: (any) => void
 ): Node<ReactFlowNodeData> {
   return {
     id: artf.id,
@@ -81,9 +79,7 @@ export function getArtifactNode(
       nodeId: artf.id,
       label: artf.name,
     },
-    // Give an initial position. We will reposition this node later.
-    //position: { x: 0, y: 0 },
-    position: pos
+    position: pos,
   };
 }
 
--- a/src/ui/common/src/utils/reactflow.ts
+++ b/src/ui/common/src/utils/reactflow.ts
@@ -43,8 +43,9 @@ type NodePos = { x: number; y: number };
 
 export function getOperatorNode(
   op: Operator,
+  pos: NodePos,
   onChange: () => void,
-  onConnect: (any) => void
+  onConnect: (any) => void,
 ): Node<ReactFlowNodeData> {
   return {
     id: op.id,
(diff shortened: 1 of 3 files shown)


src/ui/common/src/utils/reactflow.ts
```diff
@@ -58,14 +59,16 @@ export function getOperatorNode(
       label: op.name,
     },
     // Give an initial position. We will reposition this node later.
-    position: { x: 0, y: 0 },
+    //position: { x: 0, y: 0 },
```
agiron123, line 62: remove comment

vsreekanti:
@agiron123 @likawind, it shouldn't be too difficult to modify the inputs to the algorithm to take into account metrics and checks, right? The algorithm doesn't really care about whether it's positioning an artifact or an operator, correct? If so, with a little bit of preprocessing, we should be able to "collapse" the nodes before sending the API call.

agiron123:
> @vsreekanti I think what we can do here is add Artifacts to the request body that's sent out when doing positioning so that we can check the Artifact type and remove edge from the DAG on the backend before the positioning algorithm is run.

agiron123:
Not sure why this linter issue is failing. When I run linters locally, it says that there are no outstanding changes:

```
➜  aqueduct git:(eng-2620-dag-line-crossing-fix) python3 scripts/run_linters.py --golang
Linting Go code...
WARN [linters context] rowserrcheck is disabled because of generics. You can track the evolution of the generics support by following the https://github.com/golangci/golangci-lint/issues/2649. 
WARN [linters context] sqlclosecheck is disabled because of generics. You can track the evolution of the generics support by following the https://github.com/golangci/golangci-lint/issues/2649. 
WARN [linters context] structcheck is disabled because of generics. You can track the evolution of the generics support by following the https://github.com/golangci/golangci-lint/issues/2649. 
WARN [linters context] wastedassign is disabled because of generics. You can track the evolution of the generics support by following the https://github.com/golangci/golangci-lint/issues/2649. 
➜  aqueduct git:(eng-2620-dag-line-crossing-fix) git status
On branch eng-2620-dag-line-crossing-fix
Your branch is up to date with 'origin/eng-2620-dag-line-crossing-fix'.

nothing to commit, working tree clean
```

vsreekanti:
Makes sense! I'll let @likawind comment on the implementation because he knows the code better than I do. 

re: linting, is your linter up to date with what's on GitHub Actions?

likawind (approved):
The code LGTM! @vsreekanti the positioning algorithm takes operator / artifacts as input, so it will be some work to take account collapsed metrics / checks. Given the current solution solves most problem, I think we can do the collapsing in a separatly?

likawind:
@agiron123 could you help removing the `var err error = nil` line and see if the GH action still complaints? I don't know why this is not caught before but the warning seems to be reasonable

agiron123:
> @agiron123 could you help removing the `var err error = nil` line and see if the GH action still complaints? I don't know why this is not caught before but the warning seems to be reasonable

@vsreekanti Looks like I was on version 1.48, when the latest version for golangci-lint is 1.52.

vsreekanti:
Yeah, that's a bit of a pain sometimes — that's usually the first thing to check if linters don't work.

vsreekanti:
Hm, it looks like we merged this without addressing the challenges around node spacing. Have we created a separate task for tracking that?

agiron123:
> Hm, it looks like we merged this without addressing the challenges around node spacing. Have we created a separate task for tracking that?

Sorry, I thought @likawind 's comment below addressed that we should be doing this in another PR:
The code LGTM! @vsreekanti the positioning algorithm takes operator / artifacts as input, so it will be some work to take account collapsed metrics / checks. Given the current solution solves most problem, I think we can do the collapsing in a separatly?

I've created a Linear task here to track:
https://linear.app/aqueducthq/issue/ENG-2692/dag-node-spacing

vsreekanti:
Awesome, thanks for tracking it!

PR #1122, opened 2023-03-21
Set up demo codespace

likawind:
This PR adds necessary change to deploy a demo codespace that consists of all example notebooks. User can browse these workflows in UI and also play around with the notebooks. We also added a readme to instruct the user to do so.

Caveats:
* I need to remove `sentiment` as it requires torch. The resulting img will be 5G and it crashes container build (I believe the dkr pull exceed the size limit).
* Some workflows are deployed with a schedule. This means they will be triggered at codespace start time as the server will treat these runs as missing

ENG-2546

## Loom demo (if any)
This link should work: https://github.com/codespaces/new?hide_repo_select=true&ref=demo_codespace&repo=496844646


vsreekanti:
@likawind, a couple questions: 
1. Will we need to keep this branch up to date on every release? or only when the demo DB changes? 
2. Can we turn off the schedule? Seems unnecessary for this.

likawind:
@vsreekanti yes, we should keep it up to date. I will update release script / checklist once the PR is pushed.

The thing is we run the entire example notebooks to publish them in the codespace, so the schedule follows whatever in these notebooks.

I'm not sure if we should change the notebooks since they serve for other purposes. The easiest thing to do is to add a server flag so that we skip backfilling skipped workflows on start time?

vsreekanti:
I'd be okay with removing the schedules from the example notebooks. If people are running those, there's no reason that they would want it to actually update every day. 🙂

likawind:
@vsreekanti updated all schedules, this PR should be good to go

vsreekanti (approved):

PR #1123, opened 2023-03-21
Remove gathering requirements from user environment

jpurusho65:
Remove gathering module prerequisites from environment using pip freeze command
This is error prone, and different user environment may have different modules, and leads to inconsistent runtime behavior.
This change raises a RequirementsFileMissingError exception if either requirement is not provided as a list parameter or requirement.txt file is missing in the current working directory of the imported python function.

ENG 2551

### DIFF
--- a/integration_tests/backend/setup/flow_with_failure.py
+++ b/integration_tests/backend/setup/flow_with_failure.py
@@ -11,12 +11,12 @@ def setup_flow_with_failure(client: aqueduct.Client, integration_name: str) -> s
     n_runs = 1
     integration = client.integration(name=integration_name)
 
-    @aqueduct.op()
+    @aqueduct.op
     def bad_op(df):
         x = y  # intentional buggy code
         df["new"] = df["review"]
 
-    @aqueduct.op()
+    @aqueduct.op
     def bad_op_downstream(df):
         return df
 
--- a/integration_tests/no_concurrency/sdk/requirements_test.py
+++ b/integration_tests/no_concurrency/sdk/requirements_test.py
@@ -6,7 +6,7 @@ from aqueduct.error import AqueductError, InvalidUserArgumentException
 from transformers_model.model import sentiment_prediction_using_transformers
 from utils import SENTIMENT_SQL_QUERY, get_integration_name
 
-from aqueduct import infer_requirements, op
+from aqueduct import op
 
 INVALID_REQUIREMENTS_PATH = "~/random.txt"
 VALID_REQUIREMENTS_PATH = "transformers_model/requirements.txt"
@@ -46,42 +46,12 @@ def _install_transformers_package():
     _run_shell_command(f"{sys.executable} -m pip install transformers")
 
 
-def _check_infer_requirements(transformers_exists: bool):
-    """Runs our supplied `infer_requirements` method and checks for the presence of `transformers`."""
-    inferred_reqs = infer_requirements()
-    if transformers_exists:
-        assert any(req_str.startswith("transformers") for req_str in inferred_reqs)
-    else:
-        assert all(not req_str.startswith("transformers") for req_str in inferred_reqs)
-
-
-@op()
+@op
 def sentiment_prediction_without_reqs_path(reviews: pd.DataFrame) -> pd.DataFrame:
     import transformers
 
     model = transformers.pipeline("sentiment-analysis")
     return reviews.join(pd.DataFrame(model(list(reviews["review"]))))
-
-
-def test_infer_requirements(client):
-    if _transformers_package_exists():
-        _uninstall_transformers_package()
-
-    # If no requirements are supplied, our inference will not pick up the transformers package.
-    db = client.integration(name=get_integration_name())
-    table = db.sql(query=SENTIMENT_SQL_QUERY)
-    with pytest.raises(AqueductError):
-        sentiment_prediction_without_reqs_path(table)
-
-    _check_infer_requirements(transformers_exists=False)
-    _install_transformers_package()
-
-    # After installing the transformers package in the environment, check that we now
-    # are inferring the `transformers` package. There's not a great way of checking this exactly,
-    # since the SDK and backend are expected to share the same python environment currently.
-    # TODO(ENG-1471): Once installation isolation is provided, we can rewrite this to test case
-    #  to simply check that preview works.
-    _check_infer_requirements(transformers_exists=True)
 
 
 @op(requirements=INVALID_REQUIREMENTS_PATH)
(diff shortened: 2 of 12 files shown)


sdk/aqueduct/utils/function_packaging.py
```diff
@@ -74,8 +78,7 @@ def serialize_function(
             Can be either a path to the requirements.txt file or a list of pip requirements specifiers.
             (eg. ["transformers==4.21.0", "numpy==1.22.4"]. If not supplied, we'll first
             look for a `requirements.txt` file in the same directory as the decorated function
-            and install those. Otherwise, we'll attempt to infer the requirements with
-            `pip freeze`.
+            and install those. Otherwise FileNotFoundError exception will be raised.
```
kenxu95, line 81: What do you think about using a custom RequirementsNotFoundError class instead? I think it may make more sense since requirements can be defined in code (@op) or in a file (requirements.txt)

sdk/aqueduct/tests/decorator_test.py
```diff
@@ -55,6 +55,14 @@ def test_handle_reserved_file_dependencies():
         )
 
 
+def test_missing_requirements_file():
+    with pytest.raises(FileNotFoundError):
```
kenxu95, line 59: Can you add the "pytest.raises(..., match=...)" argument so we also confirm the error message here?

sdk/aqueduct/tests/decorator_test.py
```diff
@@ -55,6 +55,14 @@ def test_handle_reserved_file_dependencies():
...
+        _package_files_and_requirements(
+            python_function,
+            dir_path=os.path.join(os.getcwd(), "test_files"),
+        )
```
kenxu95, line 63: Why does this fail? Isn't there a requirements.txt file in `test_files/python_function`? I guess I'm confused as to why there's a requirements.txt file in test_files/python_function at all.

sdk/aqueduct/utils/function_packaging.py
```diff
@@ -211,13 +218,10 @@ def _package_files_and_requirements(
             shutil.copy(REQUIREMENTS_FILE, packaged_requirements_path)
 
         else:
-            # No requirements have been provided, so we use `pip freeze` to infer.
-            logger().info(
-                "%s: No requirements.txt file detected, self-generating file by inferring package dependencies."
-                % func.__name__
+            # Do not infer package dependencies from environment.
+            raise FileNotFoundError(
+                "A valid requirements list or requirements.txt file must be provided"
```
kenxu95, line 223: Let's give the user a little bit more guidance here. Can you add that the requirements.txt file should be in the same directory as the function definition? And that you can alternatively add the requirements directly in the @op/@metric/@check decorator.

kenxu95 (changes requested):
Changes look mostly good! Just a couple small requests/questions

Can we also add one integration test that tries to run an @op that doesn't have requirements? Very similar to the unit test but at a higher level to capture user behavior.

sdk/aqueduct/tests/decorator_test.py
jpurusho65: done

integration_tests/backend/setup/flow_with_failure.py
```diff
@@ -11,12 +11,12 @@ def setup_flow_with_failure(client: aqueduct.Client, integration_name: str) -> s
...
     integration = client.integration(name=integration_name)
 
-    @aqueduct.op()
+    @aqueduct.op(requirements=[])
     def bad_op(df):
         x = y  # intentional buggy code
         df["new"] = df["review"]
 
-    @aqueduct.op()
+    @aqueduct.op(requirements=[])
```
kenxu95, line 19: Any reason we can't just add `requirements.txt` to `integration_tests/backend/`? This seems like a hassle.

integration_tests/no_concurrency/sdk/requirements_test.py
```diff
@@ -46,44 +46,14 @@ def _install_transformers_package():
...
-    """Runs our supplied `infer_requirements` method and checks for the presence of `transformers`."""
-    inferred_reqs = infer_requirements()
-    if transformers_exists:
-        assert any(req_str.startswith("transformers") for req_str in inferred_reqs)
-    else:
-        assert all(not req_str.startswith("transformers") for req_str in inferred_reqs)
-
-
-@op()
+@op(requirements=[])
```
kenxu95, line 49: Same comment about adding `requirements.txt` to `integration_tests/no_concurrency/sdk`.

sdk/aqueduct/decorator.py
```diff
@@ -825,8 +825,7 @@ def check(
             Can be either a path to the requirements.txt file or a list of pip requirements specifiers.
             (eg. ["transformers==4.21.0", "numpy==1.22.4"]. If not supplied, we'll first
             look for a `requirements.txt` file in the same directory as the decorated function
-            and install those. Otherwise, we'll attempt to infer the requirements with
-            `pip freeze`.
+            and install those. Otherwise, the method raises FileNotFoundError exception.
```
kenxu95, line 828: We raise a `RequirementsFileMissingError` right?

sdk/aqueduct/error.py
```diff
@@ -130,3 +130,8 @@ class InvalidGithubQueryError(Error):
 # Exception raised when client fails to validate the server.
 class ClientValidationError(Error):
     pass
+
+
+# Exception raised when requirements.txt file is missing
+class RequirementsFileMissingError(Error):
```
kenxu95, line 136: Sorry it's a super small language nit but I wanted this to just be `RequirementsMissingError` without the "file" part, since requirements can be specified in multiple ways.

sdk/aqueduct/utils/function_packaging.py
```diff
@@ -204,20 +207,20 @@ def _package_files_and_requirements(
...
         else:
-            # No requirements have been provided, so we use `pip freeze` to infer.
-            logger().info(
-                "%s: No requirements.txt file detected, self-generating file by inferring package dependencies."
-                % func.__name__
+            # Do not infer package dependencies from environment.
+            raise RequirementsFileMissingError(
+                "A valid requirements.txt file must be provided "
+                "and must be in the same directory as the function "
+                "definition or alternatively add the requirements "
```
kenxu95, line 221: nit: "definition. Alternatively, add ..."

kenxu95 (approved):
Happy to unblock but please address the comments before merging :) It would be best if we could add `requirements.txt` to our test folders, instead of forcing every @op defined in them to have `requirements=[]`.

PR #1125, opened 2023-03-21
ENG-2577 follow up

agiron123:
This PR show a workflow's status badge in workflow list page as warning status if there is a warning level check that fails in the workflow.

ENG-2577

## Loom demo (if any)

See warning_bad_check workflow below:
image

### DIFF
--- a/src/ui/common/src/components/pages/workflows/index.tsx
+++ b/src/ui/common/src/components/pages/workflows/index.tsx
@@ -94,9 +94,13 @@ const WorkflowsPage: React.FC<Props> = ({ user, Layout = DefaultLayout }) => {
           const value =
             check.result?.exec_state.status === 'succeeded' ? 'True' : 'False';
           const level = check.spec?.check?.level ?? CheckLevel.Warning;
-          const status = check.result?.exec_state?.status ?? ExecutionStatus.Unknown;
+          const status =
+            check.result?.exec_state?.status ?? ExecutionStatus.Unknown;
 
-          if (status === ExecutionStatus.Failed && level === CheckLevel.Warning) {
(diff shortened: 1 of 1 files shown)


src/ui/common/src/components/pages/workflows/index.tsx
```diff
@@ -87,19 +87,30 @@ const WorkflowsPage: React.FC<Props> = ({ user, Layout = DefaultLayout }) => {
         });
       }
 
+      let containsWarning = false;
       let checks = [];
       if (workflow.checks) {
         checks = workflow.checks.map((check) => {
           const value =
             check.result?.exec_state.status === 'succeeded' ? 'True' : 'False';
+          const level = check.spec?.check?.level ?? CheckLevel.Warning;
```
vsreekanti, line 96: Is there a world in which this wouldn't be set?

vsreekanti (approved):
LGTM! Thanks for the quick turnaround. Just one question in line.

src/ui/common/src/components/pages/workflows/index.tsx
agiron123: I'm not entirely sure 🤔 Just moved around what was there. 

@likawind Could this happen when there is a check that hasn't run yet?
Or maybe a situation where the spec isn't set?

likawind: We shouldn't reach the fall-back in practice. I think this is just a safeguard where the type is completely missing, and in this case something is seriously wrong in backend.

I've seen multiple complaints about this line being confusing, alternatively we can just remove the check level icon if the level is undefined

PR #1129, opened 2023-03-23
Eng 2441 Update node statuses to cancel when workflow run is finished

eunice-chan:

Currently, node statuses are not updated to cancel when the workflow fails in the Golang level.
When updating the workflow run status, this fix checks to see if the workflow run is in the complete state (succeeded/canceled/failed) and marks all pending or running nodes as canceled.

ENG 2441

## Loom demo (if any)
Before - workflow done but nodes marked pending
Screen Shot 2023-03-22 at 7 54 21 PM
After - all nodes marked as cancelled after workflow run gets a final status (succeeded or failed)
Screen Shot 2023-03-22 at 7 53 44 PM

### DIFF
--- a/src/golang/lib/workflow/utils/database.go
+++ b/src/golang/lib/workflow/utils/database.go
@@ -349,6 +349,8 @@ func UpdateDAGResultMetadata(
 	dagResultID uuid.UUID,
 	execState *shared.ExecutionState,
 	dagResultRepo repos.DAGResult,
+	artifactResultRepo repos.ArtifactResult,
+	operatorResultRepo repos.OperatorResult,
 	workflowRepo repos.Workflow,
 	notificationRepo repos.Notification,
 	DB database.Database,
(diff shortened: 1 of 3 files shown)


src/golang/lib/workflow/utils/database.go
```diff
@@ -374,6 +376,57 @@ func UpdateDAGResultMetadata(
 		return err
 	}
 
+	// Check if DAG has finished running.
+	if execState.Status == shared.CanceledExecutionStatus || execState.Status == shared.FailedExecutionStatus || execState.Status == shared.SucceededExecutionStatus {
```
saurav-c, line 380: If the DAG status is succeeded, do we need to perform the following checks? The artifact result and op results should all be succeeded already?

src/golang/lib/workflow/utils/database.go
```diff
@@ -374,6 +376,57 @@ func UpdateDAGResultMetadata(
 		return err
 	}
 
+	// Check if DAG has finished running.
```
saurav-c, line 379: can you update the comment for this function so it describes the fact that the artifact and op results are also updated if the dag status changes to cancelled or failed

saurav-c (approved):
LGTM :shipit:

PR #1144, opened 2023-03-28
Eng 2331 Fix bug where workflow run can stay in pending states forever

likawind:
This PR fixes the bug where the workflow run can stay in pending states forever. This happens when the server is killed when there are actively running workflows.

The solution modifies server starting code by adding a step that marks all `pending` and `running` artf / op / dag results to `canceled` with current timestamp.

ENG-2331

## Testing
All queries have corresponding unit tests.
Verified by killing the server with active workflow and restart:
Before:
Screenshot 2023-03-27 at 11 02 33 PM

After:
Screenshot 2023-03-27 at 11 06 08 PM


src/golang/lib/database/stmt_preparers/stmt_preparer.go
```diff
@@ -15,4 +21,18 @@ type StmtPreparer interface {
...
 	PrepareUpdateWhereWithReturnAllStmt(table string, columns []string, predicateColumn string, allColumns string) string
+	// This helper method returns a query fragment that updates exec state blob
+	// with the given status and timestamp.
+	// This is useful to update the state without deserializing the content.
+	// Example: PrepareUpdateExecStateStmt('integration.execution_state', 'succeeded', time.Now())
+	// -> '`integration.execution_state = CAST(
+	//  json_set(json_set(integration.execution_state, '$.status', 'succeeded'),
+	//    '$.timestamps.finished_at', '2023-03-27 14:13PM'
+	//  ) AS BLOB)`
+	PrepareUpdateExecStateStmt(
```
saurav-c, line 32: From a code organization standpoint, I would like to keep `lib/database` free of any Aqueduct application logic. If there is a query that is reused between different `sqlite repos`, lets create a shared helper underneath `repos/sqlite`.

src/golang/lib/database/stmt_preparers/update_execution_state.go
```diff
@@ -0,0 +1,44 @@
...
+	status shared.ExecutionStatus,
+	timestamp time.Time,
+	offset int,
+) (fragment string, args []interface{}, err error) {
+	timestampField, err := shared.ExecutionTimestampsJsonFieldByStatus(status)
+	if err != nil {
+		return "", nil, err
+	}
+
+	// (likawind) If passing the timestamp object directly to query, it somehow bypass the
```
saurav-c, line 21: hmm thats interesting because we insert `time.Time` objects directly elsewhere in our codebase.

src/golang/lib/models/shared/execution_timestamps.go
```diff
@@ -1,10 +1,38 @@
...
+)
 
 type ExecutionTimestamps struct {
 	RegisteredAt *time.Time `json:"registered_at"`
 	PendingAt    *time.Time `json:"pending_at"`
 	RunningAt    *time.Time `json:"running_at"`
 	FinishedAt   *time.Time `json:"finished_at"`
 }
+
+func ExecutionTimestampsJsonFieldByStatus(
```
saurav-c, line 16: Can you add a comment explaining what this method is returning?

src/golang/cmd/server/server/cron.go
```diff
@@ -44,12 +54,77 @@ func (s *AqServer) triggerMissedCronJobs(
 	}
 }
 
-// RunMissedCronJobs first gets the latest workflow run timestamp of all deployed workflows that are
+// backfillKilledJobs backfills all pending and running op/artf/DAG _results
+// and mark them as canceled.
+func (s *AqServer) backfillKilledJobs(ctx context.Context) error {
```
saurav-c, line 59: How does this work for non-Aqueduct orchestrated workflows, such as Airflow. In that case we don't want to turn running operators and artifacts to cancelled.

saurav-c (changes requested):

src/golang/lib/database/stmt_preparers/update_execution_state.go
likawind: I think it's because we insert it as a part of `json_set` and it somehow doesn't work 😭

src/golang/cmd/server/server/cron.go
likawind: Interesting..how does the sync with airflow works? I think we can either bypass this backfill for these engines or runs the sync as a part of this method. I'd prefer to do the second option if there's an easy way to do so

src/golang/lib/database/stmt_preparers/update_execution_state.go
likawind: Elsewhere we mainly insert directly to the column

src/golang/cmd/server/server/cron.go
saurav-c: But as a first cut you can also just bypass Airflow DAGs for now, since that might be easiest to get you unblocked. This is more of an optimization anyways, as we would sync with Airflow anytime the user goes to the UI.

likawind:
@saurav-c all comments should get addressed!

saurav-c (approved):
Awesome LGTM!

PR #1146, opened 2023-03-28
[Backend] Improve lifecycle management of the artifact storage migration 

kenxu95:
Lead reviewers: @saurav-c and @likawind.

This is a WIP process PR, but the critical backend parts have been changed and tested. I would like some preliminary feedback on 1) the data model and 2) the updated logic in connect_integration.go to make sure there are no surprises when the rest of the work is fleshed out.

Testing:
- Tested that I'm able to migration to an S3 bucket, and from that S3 bucket to another S3 bucket, and that the storage migration table is in the appropriate state afterwards.

I also updated `configure_storage.go` later following the same pattern. Once the latter is updated we'll be able to migrate from S3 back to local.

There will be a follow-up for the UI changes, as well as any additional backend routes we need to support those UI changes.

Known issues that I will address (or make tasks for):
- [x] We should perform the deletion of contents in the old storage layer AFTER updating the global storage config. Otherwise, this can lead to an inconsistent and broken state. It's not a huge deal if we accidentally leave some files behind.
- [ ] There's currently no way of testing s3 -> local filesystem migration through the UI or SDK? (Task created)
- [ ] We should probably lock the `storage_migration` table so that only one migration can be ongoing at a time. (Task created)
- [ ] There should be an enforced timeout for the migration process. (Task created)
- [x] Disallow deleting an integration that is also being used as the current metadata store.
- [ ] [Nice-to-have] Migrating to the same S3 bucket, or migrating local->local should error. This currently works for some reason, but we can probably disallow it earlier.

ENG-2604

## Loom demo (if any)
https://www.loom.com/share/e04f228644f24c2eb0b0b583503c8abd


src/golang/cmd/server/handler/connect_integration.go
```diff
@@ -202,58 +207,241 @@ func (h *ConnectIntegrationHandler) Perform(ctx context.Context, interfaceArgs i
 		// This integration should be used as the new storage layer.
 		// In order to do so, we need to migrate all content from the old store
 		// to the new store. This requires pausing the server and then restarting it.
-		// All of this logic is performed asynchronously so that the user knows that
-		// the connect integration request has succeeded and that the migration is now
-		// under way.
+
```
likawind, line 210: Should we put these to a designated `lib` package like `storage_migration`, and call stuffs like `storage_migration.Prepare()`, `storage_migration.Migrate()`?

src/golang/cmd/server/handler/connect_integration.go
```diff
@@ -202,58 +207,241 @@ func (h *ConnectIntegrationHandler) Perform(ctx context.Context, interfaceArgs i
...
+			var err error
 			defer func() {
-				if err := lock.Unlock(); err != nil {
-					log.Errorf("Unexpected error when unlocking workflow execution lock: %v", err)
+				if err != nil {
+					// Regardless of whether the migration succeeded, we should track the finished timestamp.
+					finishedAt := time.Now()
+					execState.Timestamps.FinishedAt = &finishedAt
+
+					execState.UpdateWithFailure(
```
likawind, line 243: Interesting, so we are not updating timestamp in this `UpdateWithFailure` call?

src/golang/cmd/server/handler/connect_integration.go
```diff
@@ -202,58 +207,241 @@ func (h *ConnectIntegrationHandler) Perform(ctx context.Context, interfaceArgs i
...
+		storageMigrationID,
+		updates,
+		db,
+	)
+	if err != nil {
+		return errors.Wrap(err, "Unexpected error when updating storage migration execution state.")
+	}
+
+	// Only need to do this if we're committing a transaction.
+	if txn, ok := db.(database.Transaction); ok {
```
likawind, line 364: What does this block do? It feels like an antipattern to me - all transaction owner (whoever calls txn := db.BeginTx()) should be responsible for calling `Commit()` and `TxnRollBack()`. I'm confused why we need to access some transaction this helper didn't create itself.

likawind (commented):
Great work! Left a few comments

src/golang/cmd/server/handler/connect_integration.go
```diff
@@ -202,58 +207,241 @@ func (h *ConnectIntegrationHandler) Perform(ctx context.Context, interfaceArgs i
...
 			}
 
 			log.Info("Successfully migrated the storage layer!")
+			finishedAt := time.Now()
+			execState.Timestamps.FinishedAt = &finishedAt
+			execState.Status = shared.SucceededExecutionStatus
+
+			// The update of the storage config and storage migration entry should happen together.
+			// While we don't enforce this atomically, we can make the two update together to minimize the risk.
+			err = h.updateStorageMigrationExecState(ctx, storageMigrationObj.ID, &execState)
```
saurav-c, line 289: As we discussed a couple days ago, do we have a task that handles combining this logic into one, so the `config` pkg always reads from the DB to get the latest storage config.

src/golang/cmd/server/handler/connect_integration.go
```diff
@@ -202,58 +207,241 @@ func (h *ConnectIntegrationHandler) Perform(ctx context.Context, interfaceArgs i
...
+	if execState.Status == shared.SucceededExecutionStatus {
+		updates[models.StorageMigrationCurrent] = true
+
+		// If there was a previous storage migration, update that entry to be `current=False`.
+		oldStorageMigrationObj, err := h.StorageMigrationRepo.Current(ctx, h.Database)
+		if err == nil {
+			// Updating the old storage migration entry must be done in the same transaction
+			// as the update of the new storage migration entry, so that there is at most one
+			// current=True entry in the storage_migration table.
+			txn, err := h.Database.BeginTx(context.Background())
```
saurav-c, line 328: I would actually just create a `txn` at the start of this function. The read call of `StorageMigrationRepo.Current` should be inside of this transaction just from a database consistency standpoint. We wouldn't want to read from invalid state because a different operation modified the table between the read and when this transaction performs the update. 

Even if there is higher level concurrency control that prevents multiple storage migrations from happening at the same time, this would be good practice. It also lets us remove the semi-confusing logic of overwriting the `db` variable depending on if there was a previous storage migration or not.

saurav-c (commented):
Left initial comments, still working through the rest of the PR.

src/golang/cmd/server/routes/headers.go
```diff
@@ -12,6 +12,11 @@ const (
 	IntegrationConfigHeader  = "integration-config"
 	IntegrationIDsHeader     = "integration-ids"
 
+	// Storage Migration Headers
```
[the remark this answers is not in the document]
kenxu95: I don't think any of these are json'd though?

src/golang/lib/repos/tests/storage_migration.go
```diff
@@ -0,0 +1,3 @@
+package tests
+
+// TODO
```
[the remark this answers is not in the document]
kenxu95: I will do this as part of this PR.

src/golang/lib/storage_migration/list.go
```diff
@@ -0,0 +1,75 @@
...
+	DB database.Database,
+) ([]models.StorageMigration, error) {
+	migrations, err := storageMigrationRepo.List(ctx, DB)
+	if err != nil {
+		return nil, errors.Wrap(err, "Failed to list storage migrations.")
+	}
+
+	// Apply the filters in the following order: status, completedSince, then limit.
+	if status != nil {
+		migrations = filterStorageMigrationsByStatus(migrations, *status)
```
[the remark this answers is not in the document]
kenxu95: It's not too hard but it is a little tricky and I felt like there probably aren't enough storage migrations for listing them all to matter.

src/golang/cmd/server/handler/connect_integration.go
kenxu95: I refactored the migration code into it's own file. Lemme know if that works.

src/golang/cmd/server/handler/connect_integration.go
kenxu95: We are not. Do you think we should?

src/golang/cmd/server/handler/connect_integration.go
kenxu95: Yes will add.

src/golang/cmd/server/handler/connect_integration.go
kenxu95: It's a little confusing, I agree. Refactored it so that the transaction definition happens at the top of the function.

src/golang/cmd/server/handler/connect_integration.go
kenxu95: Actually just went ahead and did it.

src/golang/cmd/server/routes/headers.go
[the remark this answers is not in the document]
likawind: actually you are right, but it looks like they are query parameters (`/storage-migrations?limit=5`) tough?

cc @jpurusho65 as you may have better experience on API pattern

src/golang/cmd/server/handler/connect_integration.go
kenxu95: https://linear.app/aqueducthq/issue/ENG-2725/have-the-get-config-route-always-read-from-the-storage-migration-table

kenxu95:
@saurav-c @likawind ok this PR should have all the components ready now. The UI part is here: https://github.com/aqueducthq/aqueduct/pull/1155

src/golang/cmd/server/handler/connect_integration.go
```diff
@@ -199,61 +204,49 @@ func (h *ConnectIntegrationHandler) Perform(ctx context.Context, interfaceArgs i
...
-				log.Errorf("Unexpected error when setting the new storage layer: %v", err)
-			}
+		newStorageConfig, err := storage.ConvertIntegrationConfigToStorageConfig(args.Service, confData)
+		if err != nil {
+			return emptyResp, http.StatusBadRequest, errors.Wrap(err, "Integration config is malformed.")
+		}
 
-			log.Info("Successfully migrated the storage layer!")
-		}()
+		err = storage_migration.PerformStorageMigration(
```
likawind, line 217: @saurav-c my understanding is we don't need to name method repeatedly after package name? In this case, would `Perform()` be sufficient here?

likawind (approved):
Thanks for addressing comments, nothing major from my end!

src/golang/cmd/server/handler/connect_integration.go
saurav-c: Yeah, that would be good practice

saurav-c (approved):
Left a couple minor comments, otherwise LGTM

src/golang/cmd/server/handler/v2/list_storage_migrations.go
```diff
@@ -0,0 +1,113 @@
...
+
+	Database             database.Database
+	StorageMigrationRepo repos.StorageMigration
+}
+
+type listStorageMigrationsArgs struct {
+	*aq_context.AqContext
+
+	// See the route description above for what each of these filters mean.
+	status         *string
```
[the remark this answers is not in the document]
kenxu95: Those are optional fields that will be left as nil if the request doesn't want to apply those filters

kenxu95:
Pre-merge tests already ran, so skipping them when merging in UI changes.

PR #1121, opened 2023-03-21
[ENG-2676] Show operator context on nodes

vsreekanti:
## Describe your changes and why you are making these changes

This PR updates the node design to show context about operators on each node. In particular, it replaces the generic icon at the top of each node with:
1. the engine used by compute operators
2. the data system used by load & save operators.

This PR also takes the opportunity to refactor a bunch of code that was very redundant — we've removed all the duplicate node definitions and collapsed them into a single Node class with a little bit of branching logic.

It also generalizes what was previously called `EngineItem` to `ResourceItem` to also support data systems in addition to compute systems. 

## Related issue number (if any)

ENG-2676, ENG-2648

## Loom demo (if any)

https://www.loom.com/share/2101e9a838e648c592251e36529403e3

## Checklist before requesting a review
- [x] I have created a descriptive PR title. The PR title should complete the sentence "This PR...".
- [x] I have performed a self-review of my code.
- [x] I have included a small demo of the changes. For the UI, this would be a screenshot or a Loom video.
- [N/A] If this is a new feature, I have added unit tests and integration tests.
- [N/A] I have run the integration tests locally and they are passing.
- [x] I have run the linter script locally (See `python3 scripts/run_linters.py -h` for usage).
- [x] All features on the UI continue to work correctly.
- [x] Added one of the following CI labels:
    - `run_integration_test`: Runs integration tests
    - `skip_integration_test`: Skips integration tests (Should be used when changes are ONLY documentation/UI)

PR #1150, opened 2023-03-29
[ENG-2718] Fix bug where loading different workflow DAGs causes the UI to crash

vsreekanti:

#1121 introduced a somewhat insidious bug. As of that PR, when we render the workflow DAG, we don't just depend on the node position state but also on the workflow DAG state. However, the selected DAG state seems to always load after the node position state loads. 

Since the node position state loads first, we trigger a re-render of the ReactFlow canvas, but we don't have the right node state yet since the `selectedDag` is out of date (still pointing at the old version).

This PR adds two state synchronization checks to make sure that everything has loaded correctly before we render the DAG to avoid crashes.

ENG-2718

### DIFF
--- a/src/ui/common/src/components/workflows/ReactFlowCanvas.tsx
+++ b/src/ui/common/src/components/workflows/ReactFlowCanvas.tsx
@@ -33,6 +33,22 @@ const ReactFlowCanvas: React.FC<ReactFlowCanvasProps> = ({
   );
 
   const { edges, nodes } = dagPositionState.result ?? { edges: [], nodes: [] };
+  if (edges.length === 0 || nodes.length === 0) { // The DAG position state is still loading.
+    return null;
+  }
+
+  // This is a bit of a tricky check; when we switch between workflow versions, the selected DAG
+  // does not load in sync with the DAG positioning. As a result, we need to ensure 
+  // that we only render the graph once the two sets of Redux state are in sync before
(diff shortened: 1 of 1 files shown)


src/ui/common/src/components/workflows/ReactFlowCanvas.tsx
```diff
@@ -33,6 +33,26 @@ const ReactFlowCanvas: React.FC<ReactFlowCanvasProps> = ({
...
+    // The DAG position state is still loading.
+    return null;
+  }
+
+  // This is a bit of a tricky check; when we switch between workflow versions, the selected DAG
+  // does not load in sync with the DAG positioning. As a result, we need to ensure
+  // that we only render the graph once the two sets of Redux state are in sync before
+  // proceeding; otherwise, our node IDs will be mismatched. Here, we simply check to see
+  // if the UUIDs for one of the nodes exists in the selected DAG. If it doesn't, that
+  // means the state has synced yet, so we return null and wait for it to sync.
```
cw75, line 46: Do you mean has "not" synced?

cw75 (approved):
Left a bit otherwise LGTM

src/ui/common/src/components/workflows/ReactFlowCanvas.tsx
vsreekanti: Yep, good catch.

PR #1152, opened 2023-03-30
Check for data in response payload for failed execution

jpurusho65:
Check for data in failed workflow execution response payload. If there is valid data, it is deserialized and returned as artifact result. This is helpful for boolean operator that returns True or False

ENG-2629

### DIFF
--- a/integration_tests/sdk/aqueduct_tests/checks_test.py
+++ b/integration_tests/sdk/aqueduct_tests/checks_test.py
@@ -122,13 +122,7 @@ def test_edit_check(client, data_integration, engine, flow_name):
     flow = publish_flow_test(client, artifacts=[pass2, fail], engine=engine, name=flow_name())
     flow_run = flow.latest()
     assert flow_run.artifact("foo artifact").get()
-
-    # TODO(ENG-2629): Fetching a warning check should not raise an exception.
-    with pytest.raises(
-        ArtifactNeverComputedException,
-        match="This artifact was part of an existing flow run but was never computed successfully",
-    ):
-        assert not flow_run.artifact("foo artifact (1)").get()
+    assert not flow_run.artifact("foo artifact (1)").get()
 
 
 def test_delete_check(client, data_integration):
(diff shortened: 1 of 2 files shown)


integration_tests/sdk/aqueduct_tests/checks_test.py
```diff
@@ -123,13 +123,6 @@ def foo(table):
     flow_run = flow.latest()
     assert flow_run.artifact("foo artifact").get()
 
-    # TODO(ENG-2629): Fetching a warning check should not raise an exception.
-    with pytest.raises(
-        ArtifactNeverComputedException,
-        match="This artifact was part of an existing flow run but was never computed successfully",
-    ):
-        assert not flow_run.artifact("foo artifact (1)").get()
```
kenxu95, line 131: Wait we should definitely keep this assertion in, since it should work now right?

sdk/aqueduct/backend/api_client.py
```diff
@@ -537,13 +537,17 @@ def get_artifact_result_data(
...
-            execution_status,
-        )
+
+        return_value = None
+        if "data" in parsed_response:
+            return_value = deserialize(serialization_type, artifact_type, parsed_response["data"])
+
+        if execution_status != ExecutionStatus.SUCCEEDED:
+            if return_value is not None:
+                print(f"Artifact execution failed with return value {return_value}")
```
kenxu95, line 549: Hmm let's remove this print statement. The user already gets the value back, they don't need it printed for them.

sdk/aqueduct/backend/api_client.py
```diff
@@ -537,13 +537,17 @@ def get_artifact_result_data(
...
+            else:
+                print("Artifact result unavailable due to unsuccessful execution.")
```
kenxu95, line 551: Can you use the aqueduct logger().warn(..) method here instead?

kenxu95 (changes requested):

sdk/aqueduct/backend/api_client.py
jpurusho65: done

sdk/aqueduct/backend/api_client.py
jpurusho65: done

integration_tests/sdk/aqueduct_tests/checks_test.py
jpurusho65: Agree, get should fetch some value now, running the tests again.

kenxu95 (approved):
Looks good!

PR #1153, opened 2023-03-30
Allow users to create all types of Artifact from Local Data

Fanjia-Yan:


integration_tests/sdk/aqueduct_tests/param_test.py
```diff
@@ -518,17 +520,90 @@ def test_invalid_local_data(client):
...
+        Unable to check that the input is picklable, since `pickle.loads()`
+        complains about `import of module 'param_test' failed`.
+        """
+        print(input)
+        if input != ArtifactType:
+            raise Exception("Expected Class.")
+        return input
+
+    with open("data/test.pickle", "wb") as file:
+        pickle.dump(ArtifactType, file)
```
Fanjia-Yan, line 536: To test pickle local data. I save class `ArtifactType` into a pickable file, load it as local data parameter and check for equality. 

In `test_all_param_types`, we dynamically created an `EmptyClass`, put it in `aq.create_param()`, and check for equality.

I didn't do what `test_all_param_types` did is because `pickle.load` complain a lot in pytest. Would that be enough of coverage.

integration_tests/sdk/aqueduct_tests/param_test.py
```diff
@@ -518,17 +520,90 @@ def test_invalid_local_data(client):
...
 
-def test_local_image_data_parameter(client, flow_name, engine):
+def test_all_local_data_types(client, flow_name, engine):
+    @op
+    def must_be_picklable(input):
+        """
+        Unable to check that the input is picklable, since `pickle.loads()`
+        complains about `import of module 'param_test' failed`.
+        """
+        print(input)
```
saurav-c, line 530: Lets remove this print statement

integration_tests/sdk/aqueduct_tests/param_test.py
saurav-c: Can you share the error you were seeing after doing what `test_all_param_types` does.

saurav-c: I'd rather we try to test this the same way of creating an empty class

sdk/aqueduct/constants/enums.py
```diff
@@ -241,6 +241,11 @@ class LocalDataSerializationType(str, Enum, metaclass=MetaEnum):
     JSON_TABLE = "json_table"
     PARQUET_TABLE = "parquet_table"
     IMAGE = "image"
+    JSON = "json"
+    PICKLE = "pickle"
+    STRING = "string"
+    BYTES = "bytes"
+    TF_KERAS = "tensorflow-keras-model"
```
saurav-c, line 248: Can we also add a test case for `TF_KERAS`

sdk/aqueduct/utils/serialization.py
```diff
@@ -74,23 +74,48 @@ def _read_string_content(content: bytes) -> str:
...
 
 
-def _read_local_image_content(path: str) -> Any:
+def _read_local_image_content(path: str) -> Image.Image:
     return Image.open(path)
 
 
+def _read_local_json_content(path: str) -> Any:
+    with open(path, mode="rb", encoding=DEFAULT_ENCODING) as file:
+        return json.load(file.read())
```
saurav-c, line 95: You should be able to do `json.load(file)` directly

saurav-c (commented):
Looks good for the most, but left a few comments that we should get resolved before merging

integration_tests/sdk/aqueduct_tests/param_test.py
Fanjia-Yan: Screen Shot 2023-04-07 at 3 09 09 PM
I have been  researching on this error and a lot of the complaints are coming from using multiprocessing package. I suspect that pytest is also using some sort of multiprocessing which causes the pickle.dump to fail?

Fanjia-Yan: I have created https://linear.app/aqueducthq/issue/ENG-2774/investigate-the-limitation-of-artifacttypepicklable-from-create-param. Feel free to give it a read to see if the issue makes sense

saurav-c (approved):

PR #1163, opened 2023-04-03
Enhance flow_run.describe() to also print failed execution state

kenxu95:
This is not only good for `flow_run.describe()`, but we also plug it into our integration test suite, so that whenever a workflow fails when it's supposed to succeed, we will call `flow_run.describe()` on the failed flow and get any workflow-level or operator-level errors and logs printed out. 

This should both save us time debugging, and also bring some parity between the UI and SDK wrt to error surfacing. The errors that are surfaced on the workflows page will also be accessible through `flow_run.describe()` now.

This is what an failure in expected workflow status now looks on the console:
Screen Shot 2023-04-03 at 4 19 51 PM

This is what the following workflow now looks like when fetched from the SDK:
```
@op
def foo():
    return 123

@op
def bar(input):
    print("HELLO IM IN BAR")
    5/0
    return input

output = bar.lazy(foo.lazy())
client.publish_flow("Test Flow", artifacts=output)
```
Screen Shot 2023-04-03 at 3 26 17 PM

IMPLEMENTATION:
- This required calling `get_workflow_dag_result` route from the SDK, instead of trying to parse the incomplete fields from `get_workflow.go`. I think having flow_run rely only on that route makes everything a lot easier to reason about now too.

ENG-2642

### DIFF
--- a/sdk/aqueduct/models/response_models.py
+++ b/sdk/aqueduct/models/response_models.py
@@ -12,7 +12,7 @@ from aqueduct.constants.enums import (
 from aqueduct.models.artifact import ArtifactMetadata
 from aqueduct.models.dag import Metadata, RetentionPolicy, Schedule
 from aqueduct.models.operators import LoadSpec, Operator, OperatorSpec
-from aqueduct.utils.utils import human_readable_timestamp
+from aqueduct.models.utils import human_readable_timestamp
 from pydantic import BaseModel
 
 
--- a/sdk/aqueduct/models/response_models.py
+++ b/sdk/aqueduct/models/response_models.py
@@ -168,12 +168,10 @@ class WorkflowDagResultResponse(BaseModel):
 
     """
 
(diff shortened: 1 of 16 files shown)


sdk/aqueduct/models/response_models.py
```diff
@@ -1,4 +1,5 @@
 import uuid
```
likawind, line 1: I think it's fine to iterate on the dir structure later. Ideally I'd like to follow the pattern between UI code and backend code to have clear file-level mappings between response definitions (like https://github.com/aqueducthq/aqueduct/blob/main/src/golang/lib/response/workflow.go and https://github.com/aqueducthq/aqueduct/blob/main/src/ui/common/src/handlers/responses/workflow.ts ). But we are still iterating on the plan, and we can change the python side altogether once we start to migrate SDK endpoints to v2.

PR #1164, opened 2023-04-03
[1/4] Eng 2711 Implement operator and artifact node views to capture structure information

likawind:
Base on our [API 2.0 Design](https://www.notion.so/aqueducthq/Aqueduct-2-0-REST-Interfaces-e6cfd5b011014697897770ce603beaa8?pvs=4), we will return up / downstream information when querying operator / artifacts. As a result, we are foreseeing the need of having a DB view that consolidate this inputs / outputs information so that we can save lots of application logic on map manipulation.

This PR introduces two views, `operator_node` and `artifact_node` that additionally add `dag_id`, `inputs` and `outputs` for op and artf objects. The idea is to use `GROUP BY` to aggregate all dag edges starting from / ends at op / artf to construct the list. In subsequent PR, we will consume these two views in our DB query library.

One thing to note is that since `json_group_array` has non-deterministic behavior on ordering, we explicitly capture the `idx` of each field and will push the sorting to go side.

ENG 2711

## Loom demo (if any)
Subsequent PRs will include database unit test


src/golang/cmd/migrator/versions/000026_add_operator_artifact_node_views/up_sqlite.go
```diff
@@ -0,0 +1,141 @@
+package _000026_add_operator_artifact_node_views
+
+const upSqliteScript = `
```
saurav-c, line 3: Views in SQLite are just stored SQL statements, so there isn't any performance benefit of having precomputed values ready to go. I feel like it may be more flexible to just define this as a regular query inside of a method in one of our repos instead of defining a database view for it. 

That way if we need to make changes to the query it can have no overhead as opposed to updating a view with another migration step.

likawind: I see, would you suggest by pre-running these `CREATE VIEW` while executing the queries?

likawind: In this case, do you see any draw-back if this view creation is running multiple times?

saurav-c: I would remove the `CREATE VIEW` part completely and just run the query, because SQLite won't update the view when the underlying tables change: https://www.sqlite.org/lang_createview.html.

This is different from something like Postgres which has support for updateable views, but we don't support Postgres right now, so we should optimize for Postgres when we do add support for it: https://www.percona.com/blog/postgresql-updatable-views-performing-schema-updates-with-minimal-downtime/#:~:text=PostgreSQL%20has%20introduced%20automatically%20updatable,as%20on%20a%20regular%20table.

saurav-c: I think the main thing is running `CREATE VIEW` doesn't add any value for us right now, and running the view creation multiple times is extra overhead for SQLite, as opposed to just running the query.

likawind: Thanks! So should we just do `WITH (%s) AS operator_nodes` and put this in a helper?

saurav-c: You can start from `WITH edge_ordered AS...` and keep everything else the same. I don't think you need to define `operator_nodes` since that is what you are returning

likawind: Cool, this PR and #1165 should be ready!

src/golang/lib/models/views/artifact_node.go
```diff
@@ -0,0 +1,59 @@
+package views
+
+const ArtifactNodeViewSubQuery = `
```
eunice-chan, line 3: Can you add a comment to the top of these files so describe what it does?

A little confused about what the json_group_array contains. For the outputs json_group_array, does value=all operators that take in this artifact as input and  the inputs json_group_array take in all the operators that create the artifact as output? If so, shouldn't it just be one operator that generates the artifact?

eunice-chan (approved):

src/golang/lib/models/views/artifact_node.go
saurav-c, line 3: I'd like us to only keep struct definitions in `lib/models/views`. Any SQL queries should go under the proper `repos` directory. Something like this should belong in the `artifact` repo.

saurav-c (changes requested):

src/golang/lib/models/views/artifact_node.go
likawind: Thanks! Would you mind reviewing #1165 and #1199 as well? Now that the stack is high and there will be lots of rebases, so I'd like to get all comments and address them all at once

saurav-c: Yup reviewing rn

likawind: Yeah for artifact inputs we could make it unique. In general `json_group_array` operates on `group by` clause and put together elements that are 'grouped' as an json array. We make use of it to put together inputs / outputs list which were separate rows before `group by`.

likawind: Done, updated query with unique artf input

likawind: query changes will follow up in #1165

likawind: Done :)

src/golang/lib/repos/sqlite/artifact.go
```diff
@@ -15,6 +15,57 @@ import (
 	"github.com/google/uuid"
 )
 
+const ArtifactNodeViewSubQuery = `
```
likawind, line 18: This is public because linter complaints if private const is not used

saurav-c (approved):

PR #1165, opened 2023-04-04
[2/4] Implement queries on operator and artifact nodes

likawind:
This PR implements DB queries that reads from `opeator_node` and `artifact_node` based on DAG ID. Besides common `repo` patterns, we introduced `NullIndexedList` that pushes index-sorting of inputs / outputs as a part of driver deserialization.

ENG-2711

## Testing (if any)
Unit test included using a structured workflow.


src/golang/lib/repos/sqlite/operator.go
```diff
@@ -84,6 +97,23 @@ func (*operatorReader) GetByDAG(ctx context.Context, dagID uuid.UUID, DB databas
 	return getOperators(ctx, DB, query, args...)
 }
 
+func (*operatorReader) GetNodesByDAG(
+	ctx context.Context,
+	dagID uuid.UUID,
+	DB database.Database,
+) ([]views.OperatorNode, error) {
+	query := fmt.Sprintf(
```
saurav-c, line 105: This is where I think we should put the query. We can also get rid of `views.OperatorNodeCols()` since the columns are specific to this query only.

src/golang/lib/repos/artifact.go
```diff
@@ -22,12 +23,20 @@ type artifactReader interface {
...
+	// GetNode returns theArtifactNode view given the artifact ID.
+	GetNode(ctx context.Context, ID uuid.UUID, DB database.Database) (*views.ArtifactNode, error)
+
 	// GetBatch returns the Artifacts with IDs.
 	GetBatch(ctx context.Context, IDs []uuid.UUID, DB database.Database) ([]models.Artifact, error)
 
 	// GetByDAG returns the Artifacts created by the workflow DAG with ID dagID.
 	GetByDAG(ctx context.Context, dagID uuid.UUID, DB database.Database) ([]models.Artifact, error)
 
+	// GetNodesByDAG returns the OperatorNodes in the specified DAG.
```
saurav-c, line 35: nit: typo, change `OperatorNodes` to `ArtifactNodes`

src/golang/lib/models/shared/list.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"database/sql/driver"
+	"sort"
+
+	"github.com/aqueducthq/aqueduct/lib/models/utils"
+	"github.com/dropbox/godropbox/errors"
+)
+
+var ErrReadOnlyFieldType = errors.New("The cannot serialized read-only field type.")
```
saurav-c, line 11: We don't want to use this pattern for defining errors anymore. See [here](https://github.com/aqueducthq/aqueduct/blob/main/src/golang/lib/database/database.go#L15) for an example of how we should define this. Also we should use our `lib//errors` package instead of the dropbox package.

saurav-c: Also typo in the error message

saurav-c (commented):
Generally looks good, just followed up based on my feedback on #1164

src/golang/lib/repos/sqlite/operator.go
likawind: It's reused in `GetNode()` so I think some variables should be helpful here. Also since `opView` and `artfView` follows how node objects are defined in API endpoints, it makes sense to assume we will have more queries on these views.

But I agree we should move subquery here as it's language specific and is a part of query implementation

saurav-c (approved):

PR #1177, opened 2023-04-05
Enable previews in our integration tests for compute engines

hsubbaraj-spiral:
Enable previews in our integration tests for compute engines
https://linear.app/aqueducthq/issue/ENG-2720/enable-previews-in-our-integration-tests-to-also-run-on-external


integration_tests/sdk/aqueduct_tests/test_functions/simple/model.py
```diff
@@ -3,29 +3,28 @@
 
 @op()
 def dummy_sentiment_model(df):
-    df["positivity"] = 123
+    columns = df.columns
```
hsubbaraj-spiral, line 6: These changes to the dummy models are so both pandas and Spark dataframes can use the same dummy models. I'll make the changes to the other test functions in another PR.

integration_tests/sdk/setup_integration.py
```diff
@@ -325,6 +325,24 @@ def _fetch_integration_credentials(section: str, name: str) -> Dict[str, Any]:
...
+    Checks config to see if given compute integration should have lazy exeuction.
+    Returns True if if given compute integration has `enable_previews` set.
+    Returns False otherwise.
+    """
+    test_credentials = _parse_credentials_file()
+
+    assert "compute" in test_credentials, "compute section expected in test-credentials.yml"
+    assert name in test_credentials["compute"].keys(), "%s not in test-credentials.yml." % name
+
+    if "enable_previews" in test_credentials["compute"][name].keys():
```
kenxu95, line 339: Maybe we can just `return "enable_previews" not in test_credentials["compute"][name].keys()`. It should be pretty clear what is being returned from the docstring.

integration_tests/sdk/conftest.py
```diff
@@ -126,10 +127,14 @@ def engine(request, pytestconfig):
...
-
     # Test cases process the aqueduct engine as None. We do the conversion here
     # because fixture parameters are printed as part of test execution.
-    return request.param if request.param != "aqueduct_engine" else None
+    if request.param != "aqueduct_engine":
+        lazy_config = lazy_configuration(request.param)
+        global_config({"engine": request.param, "lazy": lazy_config})
+        return request.param
+    else:
+        return None
```
kenxu95, line 137: I was actually thinking we could make a separate fixture that is autoused for this (so the user doesn't have to explicitly include `engine` for it to trigger). This fixture would essentially perform this logic here, yield, and then reset the global configuration at the end.

This makes the fixture more understandable I think, since it's focus is just to handle the global config lifecycle. It also makes it so that tests that don't include `engine` can benefit from this.

integration_tests/sdk/setup_integration.py
```diff
@@ -325,6 +325,24 @@ def _fetch_integration_credentials(section: str, name: str) -> Dict[str, Any]:
     return test_credentials[section][name]
 
 
+def lazy_configuration(name: str) -> bool:
```
kenxu95, line 328: The logical jump from `enable_previews` to "should have lazy execution" actually took me a while to get. I think it would be good to explain in the fixture why external compute engines need to have their executions set to lazy for previews.

I think part of the confusion was that the docstrings and name of this function attempt to explain things beyond the scope of it. I've prefer if that explanation was in the caller, and this method was focused on just checking if `enable_preview` was set - probably renaming this to `is_preview_enabled()` or something.

integration_tests/sdk/conftest.py
kenxu95: Also, if we don't reset at the end, we may not be able to handle parallelized tests with multiple compute engines (eg. aqueduct_engine and k8s), since the former must have previews enabled and the second may not.

kenxu95 (changes requested):

integration_tests/sdk/conftest.py
hsubbaraj-spiral: By this, you mean just the global_config({"lazy"}) part? So the new fixture would just focus on lazy execution and I'd keep the logic for the engine here.

kenxu95: mmm I was thinking everything, both the engine and config. Essentially, this fixture translates what we have in `test-config.yml` to the appropriate global_config() state. I think it's better to separate all that from the engine fixture, which has to be included explicitly by the user.

kenxu95 (approved):
Cool! But please address the last few polish comments :)

PR #1182, opened 2023-04-07
Eng 1986 warn user when there is a python version mismatch between server and client

eunice-chan:
Warn user when there is a python version mismatch between server and client as it may result in unexpected errors when running the workflow on the server.

ENG 1986

## Loom demo (if any)
Commented out most of the warning display condition to show what is the warning. I don't have a Python version mismatch so it won't show otherwise.
image

Displays warning there is a mismatch but still allows the workflow to be published (as seen in the next line, there is a URL).


sdk/aqueduct/models/response_models.py
```diff
@@ -69,9 +69,12 @@ class RegisterWorkflowResponse(BaseModel):
     Attributes:
         id:
             The uuid if of the newly registered workflow.
+        python_version:
+            The Python version in the engine in string format "major_version.minor_version.patch_level".
```
kenxu95, line 73: Can you just provide an example here instead? Like `3.8.11`.

src/golang/cmd/server/handler/register_workflow.go
```diff
@@ -355,5 +359,21 @@ func (h *RegisterWorkflowHandler) Perform(ctx context.Context, interfaceArgs int
 		}
 	}()
 
-	return registerWorkflowResponse{Id: workflowId}, http.StatusOK, nil
+	var version strings.Builder
```
kenxu95, line 362: Can you note up here that the python version check is best-effort?

sdk/aqueduct/client.py
```diff
@@ -662,7 +663,20 @@ def publish_flow(
         else:
             if run_now is None:
                 run_now = True
-            flow_id = globals.__GLOBAL_API_CLIENT__.register_workflow(dag, run_now).id
+            register_metadata = globals.__GLOBAL_API_CLIENT__.register_workflow(dag, run_now)
```
kenxu95, line 666: nit: name `registered_metadata`?

kenxu95 (approved):
I noticed the error message in your screenshot has the same version for client and server? We only want to print this warning when they mismatch right - was this just a demo oversight?

eunice-chan:
@kenxu95 Good eye! For purposes of demonstration, I removed the condition checking the values are different. When the condition is put back in place, the warning does not show up if there is no mismatch, as expected.

PR #1191, opened 2023-04-10
Add support for subfolders for aws s3

kenxu95:
When registering an S3 integration as the storage layer, the user can specify a subfolder that then acts as the root of artifact storage. The vault/ folder is created inside this subfolder. This subfolder can be defined with optional leading/trailing slashes - the backend will sanitize it into the consistent form path/to/dir before writing it to the database.

ENG-2738

## Loom demo (if any)
https://www.loom.com/share/3dbf7fd6fa394117a79d5b77661af3d3

### DIFF
--- a/src/golang/lib/models/shared/storage.go
+++ b/src/golang/lib/models/shared/storage.go
@@ -37,7 +37,7 @@ type S3Config struct {
 	Bucket string `yaml:"bucket" json:"bucket"`
 
 	// If set, use this directory as the root, otherwise we use the root of the bucket.
-	RootPath string `yaml:"root_dir" json:"root_dir"`
+	RootDir string `yaml:"root_dir" json:"root_dir"`
 
 	CredentialsPath    string `yaml:"credentialsPath" json:"credentials_path"`
 	CredentialsProfile string `yaml:"credentialsProfile"  json:"credentials_profile"`
@@ -48,6 +48,9 @@ type S3Config struct {
 type S3ConfigPublic struct {
 	Region string `yaml:"region" json:"region"`
 	Bucket string `yaml:"bucket" json:"bucket"`
+
+	// If set, use this directory as the root, otherwise we use the root of the bucket.
+	RootPath string `yaml:"root_dir" json:"root_dir"`
 }
 
 type FileConfig struct {
@@ -81,8 +84,9 @@ func (s *StorageConfig) ToPublic() (*StorageConfigPublic, error) {
 		storageConfigPublic.FileConfig = s.FileConfig
 	case S3StorageType:
 		storageConfigPublic.S3ConfigPublic = &S3ConfigPublic{
-			Region: s.S3Config.Region,
-			Bucket: s.S3Config.Bucket,
+			Region:   s.S3Config.Region,
+			Bucket:   s.S3Config.Bucket,
+			RootPath: s.S3Config.RootDir,
 		}
 	case GCSStorageType:
 		storageConfigPublic.GCSConfigPublic = &GCSConfigPublic{
--- a/src/golang/lib/models/shared/storage.go
+++ b/src/golang/lib/models/shared/storage.go
@@ -33,8 +33,12 @@ type StorageConfigPublic struct {
 }
 
 type S3Config struct {
-	Region             string `yaml:"region" json:"region"`
-	Bucket             string `yaml:"bucket" json:"bucket"`
+	Region string `yaml:"region" json:"region"`
+	Bucket string `yaml:"bucket" json:"bucket"`
+
+	// If set, use this directory as the root, otherwise we use the root of the bucket.
+	RootPath string `yaml:"root_dir" json:"root_dir"`
+
 	CredentialsPath    string `yaml:"credentialsPath" json:"credentials_path"`
 	CredentialsProfile string `yaml:"credentialsProfile"  json:"credentials_profile"`
 	AWSAccessKeyID     string `yaml:"awsAccessKeyId"  json:"aws_access_key_id"`
--- a/src/golang/lib/models/shared/storage.go
+++ b/src/golang/lib/models/shared/storage.go
@@ -36,7 +36,8 @@ type S3Config struct {
 	Region string `yaml:"region" json:"region"`
 	Bucket string `yaml:"bucket" json:"bucket"`
 
-	// If set, use this directory as the root, otherwise we use the root of the bucket.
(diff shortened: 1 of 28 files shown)


kenxu95:
@vsreekanti should this subfolder thing also apply when the uses the integration as a data integration? Eg. should the file paths be relative to the subdirectory also?

src/golang/lib/models/shared/storage.go
```diff
@@ -44,6 +49,10 @@ type S3Config struct {
 type S3ConfigPublic struct {
 	Region string `yaml:"region" json:"region"`
 	Bucket string `yaml:"bucket" json:"bucket"`
+
+	// Use this directory in the bucket as the root. If not set, we default to the root of the bucket.
+	// Expected to be santizied into the format "path/to/dir/" (without a leading slash, but with a trailing one).
+	RootPath string `yaml:"root_dir" json:"root_dir"`
```
cw75, line 55: Is there a reason we're using different variable name here? It's `RootDir` above.

src/golang/lib/storage/convert.go
```diff
@@ -23,8 +23,9 @@ func convertS3IntegrationtoStorageConfig(c *shared.S3IntegrationConfig) (*shared
 	storageConfig := &shared.StorageConfig{
 		Type: shared.S3StorageType,
 		S3Config: &shared.S3Config{
-			Bucket: fmt.Sprintf("s3://%s", c.Bucket),
-			Region: c.Region,
+			Bucket:  fmt.Sprintf("s3://%s", c.Bucket),
+			Region:  c.Region,
+			RootDir: c.RootDir,
```
cw75, line 28: same comment for the variable naming.

src/python/aqueduct_executor/operators/connectors/data/s3.py
```diff
@@ -37,6 +38,15 @@ def authenticate(self) -> None:
                 % str(e)
             )
 
+        # Check that any user-supplied root directory exists.
+        if self.root_dir != "":
+            # If nothing is returned by this filter call, then the directory does not exist.
+            if len(list(self.s3.Bucket(self.bucket).objects.filter(Prefix=self.root_dir))) == 0:
```
cw75, line 44: If there's a lot of objects in this root directory, this call may take very long to finish. I wonder if there's a lower overhead way of checking if a s3 directory exists.

src/python/aqueduct_executor/operators/connectors/data/config.py
```diff
@@ -54,8 +54,11 @@ class S3Config(models.BaseConfig):
     config_file_profile: str = ""
 
     bucket: str = ""
-
     region: str = ""
+
+    # Expected to be in the format "path/to/dir/" (without a leading slash, but with a trailing one).
+    root_dir: str = ""
```
cw75, line 60: Is this used in data connection? I thought it's only relevant for S3 used as storage.

src/ui/common/src/components/integrations/dialogs/s3Dialog.tsx
```diff
@@ -193,6 +194,19 @@ export const S3Dialog: React.FC<Props> = ({
         disableReason={editMode ? readOnlyFieldDisableReason : undefined}
       />
 
+      <IntegrationTextInputField
+        spellCheck={false}
+        required={false}
+        label="Directory"
+        description="An optional path to a directory in the S3 bucket to be used as the root. Defaults to the bucket's root."
```
cw75, line 201: Let's make a note here that clarifies this is only relevant when used as an artifact store, not relevant when used as a regular data integration.

src/python/aqueduct_executor/operators/utils/storage/s3.py
```diff
@@ -63,14 +66,33 @@ def exists(self, key: str) -> bool:
...
+
+def parse_s3_bucket_and_key_prefix(s3_bucket_path: str, root_dir_path: str) -> Tuple[str, str]:
+    """The root_dir_path is assumed to be in the format `path/to/dir/`."""
+    path_parts = s3_bucket_path.replace("s3://", "").split("/")
     bucket = path_parts.pop(0)
-    key = "/".join(path_parts)
-    return bucket, key
+
+    if root_dir_path != "":
+        path_parts += [_sanitize_path(root_dir_path)]
```
cw75, line 95: if root_dir_path is assumed to be in the format `path/to/dir/`, why do we need to sanitize it again?

cw75 (commented):
@kenxu95 looks good overall, left a few questions and comments.

src/golang/lib/models/shared/storage.go
kenxu95: nope, I went back and forth on this name so it probably was just left behind.

src/golang/lib/storage/convert.go
kenxu95: Leaving as rootdir.

src/python/aqueduct_executor/operators/connectors/data/config.py
kenxu95: It has to be here because the model forbids extra fields. I'll add a note that it's actually unused.

src/python/aqueduct_executor/operators/utils/storage/s3.py
kenxu95: It is assumed, but we don't check it here. I didn't want to be too strict with the assertions in case there was a codepath I missed. Agree, it's confusing - maybe I can just remove the comment for now and make no assumptions about the root path.

src/python/aqueduct_executor/operators/connectors/data/s3.py
kenxu95: I'm not sure, let's discuss in https://aqueducthq.slack.com/archives/C01KF131001/p1681168125986639? We already use this method in our `extract` implementation when fetching entire directories.

src/python/aqueduct_executor/operators/utils/storage/s3.py
cw75: reminder to remove the comment.

src/python/aqueduct_executor/operators/connectors/data/config.py
cw75: reminder to add the note.

cw75 (approved):
LGTM. See reminders inline to add the notes.

PR #1199, opened 2023-04-12
[3/4] Add handlers for Node reading routes

likawind:
This PR adds server handlers for node reading routes. We added 3 handlers:
* `/wf/<id>/dag/<id>/nodes` that returns all nodes on a given DAG
* `/wf/<id>/dag/<id>/node/operator/<id>` returns the given operator
* `/wf/<id>/dag/<id>/node/artifact/<id>` returns the given artifact

We also added a `parser` library to standardize request parameters parsing. We can extend this library to parse query parameters and form body.

ENG-2711
## Testings
Compiles. Integration tests should be sufficient to validate no regression occurred.


src/golang/cmd/server/request/parser/dag_id.go
```diff
@@ -0,0 +1,26 @@
...
+	"net/http"
+
+	"github.com/aqueducthq/aqueduct/cmd/server/routes"
+	"github.com/dropbox/godropbox/errors"
+	"github.com/google/uuid"
+)
+
+type DagIDParser struct{}
+
+func (DagIDParser) Parse(r *http.Request) (uuid.UUID, error) {
```
saurav-c, line 14: Can we just have a generic ID parser instead of defining one for each ID type

src/golang/cmd/server/handler/v2/dag_get.go
```diff
@@ -55,16 +54,14 @@ func (h *DAGGetHandler) Prepare(r *http.Request) (interface{}, int, error) {
 		return nil, statusCode, err
 	}
 
-	workflowIDStr := chi.URLParam(r, routes.WorkflowIDParam)
-	workflowID, err := uuid.Parse(workflowIDStr)
+	workflowID, err := (parser.WorkflowIDParser{}).Parse(r)
```
saurav-c, line 57: See comment below about having a generic ID parser so we don't need to initialize unused IDParsers here.

saurav-c (approved):
Left a couple nit, otherwise LGTM

kenxu95 (changes requested):
Want to take a look in a sec, putting a block up so it doesn't get merged yet.

src/golang/cmd/server/request/parser/dag_id.go
likawind: This is actually something inspired by OpenAPI that each parameter has its own definition, so that I feel it makes sense to create one specific parser in application logic. In this way the caller only need to worry about 'parsing a parameter'. I think 'type parsing' is a lower level abstraction which we can certainly iterate further, but I don't see a strong motivation getting that abstraction refined now

likawind: Reference: https://github.com/aqueducthq/aqueduct/pull/1186/files look for `parameters` in `aqueduct.json`

src/golang/lib/response/node.go
```diff
@@ -3,28 +3,43 @@ package response
...
 
-	// upstream operator ID, must be unique.
-	From uuid.UUID `json:"from"`
+	// upstream operator ID. It should be unique in practice, but we do not
+	// enforce the assumption here.
+	Inputs []uuid.UUID `json:"inputs"`
 
 	// downstream operator IDs, could be multiple or empty.
-	To []uuid.UUID `json:"to"`
+	Outputs []uuid.UUID `json:"outputs"`
```
kenxu95, line 25: super nit: capitalize beginning of comments

kenxu95 (approved):
looks good to me

PR #1202, opened 2023-04-12
ENG-2759 Installs dependencies for mac mysql connection

agiron123:

This PR installs dependencies needed to run mysql clients on Mac.

For more background on the fix, please see StackOverflow post here:
https://stackoverflow.com/questions/39574813/error-loading-mysqldb-module-no-module-named-mysqldb

ENG-2759

## Loom demo (if any)
Loom Demo:
https://www.loom.com/share/4a292575335644a98eb91f4b88ceb9d7

### DIFF
--- a/src/python/aqueduct_executor/operators/connectors/data/execute.py
+++ b/src/python/aqueduct_executor/operators/connectors/data/execute.py
@@ -1,6 +1,6 @@
+import platform
 import sys
 from typing import Any
-import platform
 
 from aqueduct_executor.operators.connectors.data import common, config, connector, extract
 from aqueduct_executor.operators.connectors.data.spec import (
@@ -65,13 +65,10 @@ def run(spec: Spec) -> None:
         utils.write_exec_state(storage, spec.metadata_path, exec_state)
     except ExecFailureException as e:
         # We must reconcile the user logs here, since those logs are not captured on the exception.
-        from_exception_exec_state = ExecutionState.from_exception(
-            e, user_logs=exec_state.user_logs)
+        from_exception_exec_state = ExecutionState.from_exception(e, user_logs=exec_state.user_logs)
 
-        print(
-            f"Failed with error. Full Logs:\n{from_exception_exec_state.json()}")
-        utils.write_exec_state(storage, spec.metadata_path,
-                               from_exception_exec_state)
+        print(f"Failed with error. Full Logs:\n{from_exception_exec_state.json()}")
+        utils.write_exec_state(storage, spec.metadata_path, from_exception_exec_state)
         sys.exit(1)
     except MissingConnectorDependencyException as e:
         exec_state.mark_as_failure(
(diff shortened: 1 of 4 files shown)


src/python/bin/aqueduct
```diff
@@ -501,27 +532,34 @@ def install_mysql():
...
-            execute_command(["sudo", "yum", "install", "-y", "python3-devel", "mysql-devel"])
+            execute_command(["sudo", "yum", "install", "-y",
+                            "python3-devel", "mysql-devel"])
         else:
             print("Unsupported distribution:", distro.id())
     elif system == "Darwin":
         cmd = ["brew", "install", "mysql"]
-        architecture = subprocess.Popen(["which", "-a", "brew"], stdout=subprocess.PIPE ).communicate()[0]
-        if architecture.startswith(b"/opt/homebrew"): # Using arm verison of brew
+        architecture = subprocess.Popen(
```
kenxu95, line 541: Shouldn't we add the Mac mysql installation here instead?

agiron123: I didn't actually touch that code. Looks like the linter just reformatted things.

What I added here was the pip_cmd and the execute_cmd below (lines 547 and 548)

kenxu95: yeah I mean move what you added in `src/python/aqueduct_executor/operators/connectors/data/execute.py` here?  I seems better to have all the installation be done at `aqueduct install` time rather than in `setup_connector()`?

agiron123: Actually come to think of it, we're going to need to run this for Linux as well. I've now moved the installation step for the python package to the bottom. Now each platform installs their prerequisites, then installs the pymysql package.

As far as the import stuff goes, we need to run this when we import the MySQLdb module, so that should probably stay in the execute.py step. I've included some comments there to explain what's going on as well.

src/python/aqueduct_executor/operators/connectors/data/execute.py
```diff
@@ -267,28 +268,23 @@ def setup_connector(
...
+            # MySQLdb is a C-based library
+            import pymysql
+
+            # Implementation can be found here: https://github.com/PyMySQL/PyMySQL/blob/main/pymysql/__init__.py
+            pymysql.install_as_MySQLdb()
             import MySQLdb
         except:
             raise MissingConnectorDependencyException(
-                "Unable to initialize the MySQL connector. Have you run `aqueduct install mysql`?"
+                "Unable to initialize the MySQL\/MariaDB connector. Have you run `aqueduct install mysql`?"
```
kenxu95, line 282: Can we delete the now useless MariaDB file then?

kenxu95 (approved):

src/python/aqueduct_executor/operators/connectors/data/execute.py
agiron123: I'm not sure I follow which file you are referring to.

PR #1205, opened 2023-04-13
Fix creating local tensorflow model data fails with module import and tensorflow versioning

Fanjia-Yan:
This PR contains two changes regarding local tensorflow data parameter test:
1. Latest Tensorflow does not support Python 3.7. With older tensorflow, it does not have package module `keras.saving.load_model`. Therefore, we change the loading method to `keras.models.load_model` which presents in both new and older version of tensorflow.
2. Local tensorflow model parameter tests fail with our Periodic Integration test. The reason is that we didn't have tensorflow as a part of param docker image. I added the image with latest version of tensorflow and mark the failed test inactive to k8s which should be reverted next release. 

### DIFF
--- a/.github/actions/setup-server/action.yml
+++ b/.github/actions/setup-server/action.yml
@@ -62,7 +62,7 @@ runs:
 
     - name: Install packages needed for testing
       shell: bash
-      run: python3 -m pip install nltk pytest-xdist tensorflow
+      run: python3 -m pip install nltk pytest-xdist
 
     - name: Wait for server again
       shell: bash
--- a/sdk/aqueduct/utils/serialization.py
+++ b/sdk/aqueduct/utils/serialization.py
@@ -114,7 +114,7 @@ def _read_local_bytes_content(path: str) -> bytes:
 def _read_local_tf_keras_model(path: str) -> Any:
     from tensorflow import keras
 
-    return keras.saving.load_model(path)
+    return keras.models.load_model(path)
 
 
 # Returns a tf.keras.Model type. We don't assume that every user has it installed,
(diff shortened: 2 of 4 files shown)


sdk/aqueduct/utils/serialization.py
```diff
@@ -114,7 +114,7 @@ def _read_local_bytes_content(path: str) -> bytes:
 def _read_local_tf_keras_model(path: str) -> Any:
     from tensorflow import keras
 
-    return keras.saving.load_model(path)
+    return keras.models.load_model(path)
```
hsubbaraj-spiral, line 117: If this is backwards compatible (works with multiple TF versions), do we need to pin the TF version in the dockerfile?

Fanjia-Yan: Lower bounded it at the latest TF release.

kenxu95: I think we'll want to just run the parameter operators locally as a long-term solution for K8s.

.github/actions/setup-server/action.yml
```diff
@@ -62,7 +62,7 @@ runs:
 
     - name: Install packages needed for testing
       shell: bash
-      run: python3 -m pip install nltk pytest-xdist tensorflow
```
Fanjia-Yan, line 65: Remove tensorflow as a requirement for `setup-server` since we are not testing tensorflow models curently.

Fanjia-Yan:
I disabled the integration test for tf_keras parameter artifact. Since we don't have `tensorflow` as dependencies, previewing the artifact and executing the workflow with keras model will fail.

hsubbaraj-spiral (approved):

PR #1209, opened 2023-04-13
[2/2] Implement RTK routes for node result and content routes

likawind:
This PR implements RTK routes for a few node results and content related routes.

We also added a few improvements in codebase:
* Moved all v2 typescript handlers to a subfolder so that it's clear which ones are from V2 and which ones are from v1
* Ensures all v2 typescript handlers has correct .go file name in their starting comments
* removed unnecessary assignments for `transformErrorResp`
* removed unnecessary imports in `index.ts`
## Loom demo (if any)
Verified existing features keeps working with a full rebuild.


src/golang/cmd/server/handler/v2/dag_get.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 // This file should map directly to
-// src/ui/common/src/handlers/DagGet.tsx
+// src/ui/common/src/handlers/v2/DagGet.tsx
```
agiron123, line 18: Curious, when you run godoc, does this actually link to the file?
Or are IDEs able to to follow this and go directly there?

src/ui/common/src/handlers/parameters/ApiKey.ts
```diff
@@ -0,0 +1,3 @@
+export type APIKeyParameter = {
```
agiron123, line 1: Should these all be in separate files, or should they all be combined into one index.ts file inside of the parameters folder?

agiron123 (approved):
LGTM, left small comments but this can be merged after they are addressed.

src/ui/common/src/handlers/parameters/ApiKey.ts
likawind: I like the idea of combining them. I think we can split them into `header` `path` and `query` to clearly indicate where the parameter applies

src/golang/cmd/server/handler/v2/dag_get.go
likawind: nope there's no automated navigation :(

PR #1210, opened 2023-04-14
Update RoBERTa doc to include minimum storage required

cw75:
I verified that everything worked fine with 50GB of storage in each node.


vsreekanti:
@cw75, are we sure this isn't going to affect any other workloads? It seems a little narrow to only document it here to me. 

Have we tried a "vanilla" `torch` example? What about a `transformers` one?

cw75:
@vsreekanti by vanilla torch do you mean we just assume the requirements only contains `torch` and see what's the minimum disk space required? For `transformers`, I think it largely depends on which model the user wants to use right? Could be as big as Dolly.

vsreekanti:
Yeah, I agree — I'm wondering if we should put this in the Kubernetes documentation rather than this notebook.

PR #1216, opened 2023-04-14
Eng 2142 add ability to run the conda integration test

likawind:
This PR adds conda to in-PR integration tests

* added server environment to `delete_wf` and `register_wf` so that we skip env GC to preserve cached environments.
* adds proper parsings to test configs to connect to conda accordingly.

ENG-2142

## Loom demo (if any)
Will test new GH action on top of this branch


.github/workflows/integration-tests.yml
```diff
@@ -26,7 +26,7 @@ jobs:
         with:
           aws_access_key_id: ${{ secrets.KENNY_AWS_ACCESS_KEY_ID }}
           aws_secret_access_key: ${{ secrets.KENNY_AWS_SECRET_ACCESS_KEY }}
-          s3_test_config_path: premerge-basic-test-config.yml
+          s3_test_config_path: premerge-basic-test-config-wei.yml
```
kenxu95, line 29: note: remove "Wei" before merging.

.github/workflows/integration-tests.yml
```diff
@@ -57,6 +57,44 @@ jobs:
         with:
           prefix: Basic
 
+  run-tests-conda:
```
kenxu95, line 60: Hmm what do you think the risks of conda breaking are? If we don't expect it to happen a lot I wonder if we should save on costs and move it to periodic.

.github/workflows/integration-tests.yml
```diff
@@ -57,6 +57,44 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: ./.github/actions/setup-server
+        timeout-minutes: 5
+
+      - uses: conda-incubator/setup-miniconda@v2
+        with:
+          auto-update-conda: true
+          python-version: ${{ matrix.python-version }}
```
kenxu95, line 73: where does this `matrix.python-version` come from?

.github/workflows/integration-tests.yml
```diff
@@ -57,6 +57,44 @@ jobs:
...
+
+      - name: Install conda build
+        run: conda install conda-build
+
+      # TODO(ENG-2537): Use our separate GH actions credentials.
+      - uses: ./.github/actions/fetch-test-config
+        with:
+          aws_access_key_id: ${{ secrets.KENNY_AWS_ACCESS_KEY_ID }}
+          aws_secret_access_key: ${{ secrets.KENNY_AWS_SECRET_ACCESS_KEY }}
+          s3_test_config_path: premerge-conda-test-config-wei.yml
```
kenxu95, line 83: Another note to remove "-wei" here.

.github/workflows/integration-tests.yml
```diff
@@ -57,6 +57,44 @@ jobs:
...
+        working-directory: integration_tests/sdk
+        run: python3 run_tests.py -n 8
+
+      - name: Set the API key as an env variable.
+        run: echo "API_KEY=$(aqueduct apikey)" >> $GITHUB_ENV
+
+      - uses: ./.github/actions/upload-artifacts
+        if: always()
+        with:
+          prefix: Basic
```
kenxu95, line 96: This prefix should be something like "Conda"?

.github/workflows/pre-merge-ci.yml
```diff
@@ -1,13 +1,13 @@
...
+on:
   pull_request:
-    types: [ labeled, opened, synchronize, reopened ]
+    types: [labeled, opened, synchronize, reopened]
 
 jobs:
   trigger-integration-tests:
     if: contains(github.event.pull_request.labels.*.name, 'run_integration_test')
-    uses: aqueducthq/aqueduct/.github/workflows/integration-tests.yml@main
+    uses: aqueducthq/aqueduct/.github/workflows/integration-tests.yml@eng-2142-add-ability-to-run-the-e2e-integration
```
kenxu95, line 10: note about removing the branch name.

integration_tests/sdk/conftest.py
```diff
@@ -80,6 +80,7 @@ def _parse_flags_and_setup_data_integrations(config, client: Client):
     if should_skip:
         return
 
+    # TODO: create client here.
```
kenxu95, line 83: Is this still relevant?

integration_tests/sdk/test-config-example.yml
```diff
@@ -28,3 +31,4 @@ data:
 # `aqueduct_engine`, which is our native orchestration engine.
 compute:
   aqueduct_engine:
+    conda: test_conda
```
kenxu95, line 34: And there doesn't need to be an entry for conda in `test-credentials.yml` right? Can we document that here actually.

kenxu95 (changes requested):
I need to run now so didn't get to the core of the logic yet. I'll take another look when I'm back!

.github/workflows/integration-tests.yml
likawind: sounds good, it's easier for me to test in-PR here. Will move once all test succeeded

kenxu95 (commented):
So I'm a little confused, I thought we were taking the approach outlined here https://linear.app/aqueducthq/issue/ENG-2664/delete-all-workflows-at-the-end-of-test-suite to avoid the cleanup of Conda until the very end? I'm a little wary of having the server behavior completely differently in test vs prod mode, if it can be easily avoided.

From ChatGPT:
image

likawind:
@kenxu95 obviously I forgot my own slack discussion lol. I will resolve the test runner issue and amend this change!

integration_tests/sdk/aqueduct_tests/conftest.py
```diff
@@ -31,3 +35,9 @@ def test_relational_data_integrations_only(data_integration):
                 "Skipped for data integration `%s`, since it is not of type `%s`."
                 % (data_integration.name(), ",".join(enabled_data_integration_types))
             )
+
+
+def pytest_sessionfinish(session, exitstatus):
```
likawind, line 40: @kenxu95 could you confirm if these changes make sense?

likawind:
@kenxu95  this PR is ready given these tests are passing: https://github.com/aqueducthq/aqueduct/actions/runs/4759059667

integration_tests/sdk/data_integration_tests/conftest.py
```diff
@@ -60,3 +64,9 @@ def flow_manager(client, flow_name, engine):
     as it is mostly a mechanism by which data can be saved.
     """
     return FlowManager(client, flow_name, engine)
+
+
+def pytest_sessionfinish(session, exitstatus):
+    if not hasattr(session.config, "workerinput"):
+        client = aq.Client(*get_aqueduct_config())
+        delete_all_flows(client)
```
kenxu95, line 72: Why do we have two copies of this method? Shouldn't it just go in the parent directory's `conftest.py` once? The subfolders should inherit conftest.py hooks from the parent directories.

integration_tests/sdk/aqueduct_tests/conftest.py
kenxu95: Not sure what the `hasattr(session.config, "workerinput")` is about, can you add a comment explaining?

src/Makefile
```diff
@@ -1,7 +1,7 @@
 .DEFAULT_GOAL = all
 
 # By default, turn off the Makefile practice of printing each command before
-# you run it.
+# you run  it.
```
kenxu95, line 4: tiny nit: extra space :P

integration_tests/sdk/aqueduct_tests/conftest.py
kenxu95: Do you know if this is only ever called once, even when the tests are run with concurrency?

integration_tests/sdk/conftest.py
```diff
@@ -282,22 +282,6 @@ def get_new_flow_name():
...
-                        saved_objects_to_delete=client.flow(
-                            flow_name=flow_name
-                        ).list_saved_objects(),
-                    )
-                except Exception as e:
-                    print("Error deleting workflow %s with exception: %s" % (flow_name, e))
-                else:
-                    print("Successfully deleted workflow %s" % flow_name)
-
-    request.addfinalizer(cleanup_flows)
```
kenxu95, line 300: Actually this is huge. The entire point of the `flow_name()` fixture was to register each new flow  at the beginning of each test, and then tear it down at the end. Now that we're using your `delete_all_flows()` method, this fixture really should come useless. Can you make a task for this and tag it on this fixture so we can deprecate?

integration_tests/sdk/conftest.py
```diff
@@ -282,22 +282,6 @@ def get_new_flow_name():
         flow_names.append(flow_name)
         return flow_name
 
-    def cleanup_flows():
-        if not pytestconfig.getoption("keep_flows"):
```
kenxu95, line 286: This `--keep-flows` behavior is actually really useful during debugging. I think we're going to want to keep this! I think you'll just need to pass `pytestconfig` into `pytest_finish...` and skip the step if this is set.

integration_tests/sdk/aqueduct_tests/test_metrics/constant/requirements.txt
```diff
@@ -0,0 +1 @@
+pytest
```
kenxu95, line 1: For each of these requirements.txts, can you add a comment that these only need to be added due to Conda? Otherwise it's hard to know since user's typically don't do this.

integration_tests/sdk/aqueduct_tests/conftest.py
likawind: No this calls when and after every worker ends, the exact use of the line is to only run after all workers ends

integration_tests/sdk/data_integration_tests/conftest.py
likawind: Hmmm it didn't work when I call that in parent directory, so I put them here

integration_tests/sdk/conftest.py
likawind: I actually think it's still very useful to avoid duplicated names across test cases?

integration_tests/sdk/conftest.py
likawind: ohoh yeah good catch, I intended to move it to the `sessionfinish` fixture

integration_tests/sdk/conftest.py
kenxu95: yeah but it doesn't need to be a fixture anymore, we can just use a helper which is a lot easier (eg. publish_flow_test() can take an optional flow_name, and if it isn't set, it'll generate a random flow name for you).

likawind: Makes sense. Let's do this separately as it seems to require us change every single test case https://linear.app/aqueducthq/issue/ENG-2840/remove-flow-name-fixture

likawind:
@kenxu95 any other questions on this PR?

integration_tests/sdk/aqueduct_tests/conftest.py
kenxu95: Can you add that as a comment here? Since it's not intuitive

integration_tests/sdk/setup_integration.py
```diff
@@ -260,24 +260,53 @@ def setup_compute_integrations(client: Client, filter_to: Optional[str] = None)
...
         # Only connect to integrations that don't already exist.
         if integration_name not in connected_integrations.keys():
-            integration_config = _fetch_integration_credentials("compute", integration_name)
+            print(f"Connecting to {integration_name}")
+            if "aqueduct_engine" in compute_integrations:
+                # Connect to conda if specified, otherwise, do nothing for aq engine.
+                aq_config = _parse_config_file()["compute"][integration_name]
+                if aq_config and "conda" in aq_config:
+                    client.connect_integration(
+                        integration_name,
```
kenxu95, line 276: wait so this means the conda integration will always be named "aqueduct_engine"? There might be a collision here once we add "aqueduct_engine" to the startup script the same way we do "aqueduct_demo". Can we use the user supplied name ala "conda: <conda integration name>"?

kenxu95 (commented):
It looks good! I mainly am just blocking on the one issue with the conda integration name.

integration_tests/sdk/setup_integration.py
likawind: I see..I tried to use `conda` but we have a few assumptions in our testing framework that makes things trickier to change. We have a few places that assumes `integration_name` are keys of the "compute" field, and we check for things like global engine config for non-aqueduct engines, which is not applicable for conda.

likawind: Overall, there are a lot works entailed in the framework to use a different name here

kenxu95: Wait I think there might be a misunderstanding. I'm trying to propose this:

```
compute:
    aqueduct_engine:
          conda: <conda integration name> # use this integration name instead
```

This is mostly just a config file parsing change right?

likawind: sg! https://github.com/aqueducthq/aqueduct/pull/1216/commits/5372a2f19b8f41178a6813ff629f2caa5827d037 this should cover it

likawind:
skipping tests as earlier versions are passing, and test configs on `main` will timeout

integration_tests/sdk/aqueduct_tests/conftest.py
kenxu95: Bump on adding a comment here about the `hasattr(..., "workflowinput")` thingy

kenxu95 (approved):
Looks good! Just one nit about the comment.

PR #1222, opened 2023-04-18
Reduce concurrency of k8s tests to 1

hsubbaraj-spiral:
Reduces concurrency of K8s tests to 1 for stability reasons. Sets the timeout to 180 min to account for test time. Created a [task](https://linear.app/aqueducthq/issue/ENG-2815/investigate-why-k8s-tests-fail-with-increased-concurrency) to track the investigation of the test failures with higher concurrency.
Eng-2811
## Loom demo (if any)
https://github.com/aqueducthq/aqueduct/actions/runs/4726107721 Run of the periodic tests.

### DIFF
--- a/.github/workflows/periodic-integration-tests.yml
+++ b/.github/workflows/periodic-integration-tests.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   run-k8s-tests:
     runs-on: ubuntu-latest-4-cores
-    timeout-minutes: 600
+    timeout-minutes: 180
     name: SDK Integration Tests against K8s Compute
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


cw75 (approved):

.github/workflows/periodic-integration-tests.yml
```diff
@@ -8,7 +8,7 @@ on:
 jobs:
   run-k8s-tests:
     runs-on: ubuntu-latest-4-cores
```
kenxu95, line 10: Can we actually try to downgrade this to ubuntu-latest? I don't think we need anything that powerful if we're just doing concurrency 1. We might need to bump the timeout to something slightly more generous for the first run though.

kenxu95 (approved):
noice - just one comment about downgrading the machine power too?

PR #1223, opened 2023-04-18
[SDK] Fail when using any integrations that aren't connected yet

kenxu95:
@likawind and @hsubbaraj-spiral as lead reviewers.

A couple SDK improvements in this:

The `integration.describe()` method now shows the status and the error if the integration failed to connect:

image

Methods on integrations that are decorated with @validate_is_connected() will not be allowed to be executed if the integration is not properly connected:

image

The same thing applies for when we set the engine of an operator to an invalid integration.

image
image


sdk/aqueduct/artifacts/preview.py
```diff
@@ -50,13 +57,6 @@ def preview_artifacts(
         ],
         make_copy=True,
     )
-
-    global_engine_config: Optional[EngineConfig] = None
-    if globals.__GLOBAL_CONFIG__.engine is not None:
-        global_engine_config = generate_engine_config(
-            globals.__GLOBAL_API_CLIENT__.list_integrations(),
-            globals.__GLOBAL_CONFIG__.engine,
-        )
```
kenxu95, line 59: non-functional change. It just looks better to have the two subgraph statements next to each other.

sdk/aqueduct/integrations/validation.py
```diff
@@ -0,0 +1,19 @@
+from typing import Any, Callable
+
+from aqueduct.utils.integration_validation import validate_integration_is_connected
+
+AnyFunc = Callable[..., Any]
+
+
+def validate_is_connected() -> Callable[[AnyFunc], AnyFunc]:
```
likawind, line 8: I'd prefer exposing this as a method in integration base class and have all methods use `@self.validate_is_connected()`. In this way the method is tied to the integration and it's clear that we are ensuring 'this' integration is connected. It also avoids passing around `self` to an externally imported method (I'm not super sure if it's an antipattern but it feels weird to read :) )

src/golang/lib/lambda/utils.go
```diff
@@ -78,7 +78,11 @@ func CreateLambdaFunction(ctx context.Context, functionsToShip []LambdaFunctionT
 	defer close(pushImageChannel)
 	AddFunctionTypeToChannel(functionsToShip[:], pullImageChannel)
 
-	for i := 0; i < MaxConcurrentDownload; i++ {
+	numPullWorkers := MaxConcurrentDownload
```
likawind, line 81: what are these for?

likawind (approved):
Overall is good, left one suggestion to improve the ownership of `validate_is_connected` decorator

sdk/aqueduct/models/integration.py
```diff
@@ -76,6 +77,34 @@ def name(self) -> str:
...
+            raise IntegrationFailedToConnect(
+                "The attempt to connect to %s has failed with error: %s\n%s\n\n "
+                "Please see the /integrations page on the UI for more details."
+                % (
+                    self.name(),
+                    self._metadata.exec_state.error.tip,
+                    self._metadata.exec_state.error.context,
+                )
+            )
+        else:
```
hsubbaraj-spiral, line 102: Is there an explicit `ExecutionStatus.Pending` status to check for in progress, or is it just that after 2813, all statuses that are `None` are considered in progress?

src/golang/lib/lambda/utils.go
hsubbaraj-spiral: We pull the lambda Docker images from Dockerhub in parallel, this is making sure that the number of workers isn't greater than necessary i believe. @kenxu95 can confirm here?

kenxu95: yeah that's what it's doing. It's just leftover from when I was trying out the lambda stuff, but seemed useful.

sdk/aqueduct/models/integration.py
kenxu95: Nah I don't think this has to do with ENG-2813 - that's just saying that the None case will no longer exist. I don't think we ever have a pending state, we just go directly to running. Can make that a harder assertion here.

sdk/aqueduct/integrations/validation.py
kenxu95: hmm I don't think I can use the `@self.validate_is_connected()` syntax.. My intellij errors and also https://stackoverflow.com/questions/11731136/class-method-decorator-with-self-arguments#:~:text=You%20can't.,use%20a%20different%20method%20entirely.

There's also a bit of a layering issue, where I'd have to move the `validate_integration_is_connected()` down to the models/ layer from utils/, which I feel doesn't make too much sense. This is all because the integration classes are defined in integrations/ and the base integration class is defined at a much lower level in `models`. I think these layering issues in the SDK should be addressed, but probably not in this PR. 

What do you think?

kenxu95:
@likawind re-requesting review to sync on a path forward for the decorator.

sdk/aqueduct/models/integration.py
kenxu95: Actually, I decided to simply document this assumption. The assertion seems too aggressive since this isn't a code path we are really testing right now.

likawind (approved):

sdk/aqueduct/integrations/validation.py
likawind: yeah the current way looks fine then!

sdk/aqueduct/models/integration.py
hsubbaraj-spiral: sounds good

hsubbaraj-spiral (approved):

PR #1231, opened 2023-04-20
Refactor Spark Python Executor

hsubbaraj-spiral:
This PR refactors the spark python executors and reduces the maintenance overhead.

The technical complexity was how to share code with similar definitions but one additional parameter, a SparkSession. We also want to avoid importing pyspark in our normal codepaths that don't execute in Spark environments. To do this, we pass in the differing functions as parameters (`read_artifacts`, `write_artifact`, `infer_type_artifact`, and `setup_connector`) while sending the `SparkSession` object as a `kwarg`. This way the generic function passed in can be called in the same manner by both Spark and non-Spark codepaths.


src/python/aqueduct_executor/operators/connectors/data/execute.py
```diff
@@ -48,11 +48,49 @@ def run(spec: Spec) -> None:
...
+        write_artifact_func=utils.write_artifact,
+        setup_connector_func=setup_connector,
+        is_spark=False,
+    )
+
+
+def run_helper(
+    spec: Spec,
+    read_artifacts_func: Any,
+    write_artifact_func: Any,
```
likawind, line 63: If we are already parsing function objects, does it make sense to pass different function objects based on whether it's spark or not? In this way we don't even need to pass `is_spark` and other stuff as arguments.

For example, could we do something like
```
if is_spark:
 run_helper(spec, read_artifact_func=utils.read_spark_artifacts, write_artifact_func=utils.write_spark_artifacts, ...)
```

likawind (commented):
Looks great, this is a huge improvements! I have some minor comments but I don't think we have to spend too much time iterating on this

src/python/aqueduct_executor/operators/connectors/data/execute.py
```diff
@@ -48,11 +48,49 @@ def run(spec: Spec) -> None:
...
+    setup_connector_func: Any,
+    is_spark: bool,
```
likawind, line 65: If we have to pass this arg around, does it makes sense to explicitly pass a single `Optional[spark.Session]` object to decide if spark is enabled? Also I'd like to remove `**kwargs` and since it's not clear what to expect and how it's used. This pattern is more useful in cases like decorators where we support arbitrary inputs, which is not the case here

hsubbaraj-spiral: This is the correct pattern, however it requires that we import `pyspark.sql` in the regular code path, which we want to avoid. The kwargs pattern was to avoid the import in non-Spark environments.

kenxu95 (approved):
Just some stylistic/readability concerns from me!

PR #1233, opened 2023-04-21
Improve placeholder values for integrations

hsubbaraj-spiral:
Improved the placeholder values for the Spark, Databricks, and Lambda integrations.
https://linear.app/aqueducthq/issue/ENG-2837/improve-the-placeholder-values-for-integration-connections

### DIFF
--- a/src/ui/common/src/components/integrations/dialogs/lambdaDialog.tsx
+++ b/src/ui/common/src/components/integrations/dialogs/lambdaDialog.tsx
@@ -5,7 +5,8 @@ import { LambdaConfig } from '../../../utils/integrations';
 import { IntegrationTextInputField } from './IntegrationTextInputField';
 
 const Placeholders: LambdaConfig = {
-  role_arn: '<my lambda role ARN>',
+  role_arn: 'arn:aws:iam::123:role/lambda-function-role-arn',
+  exec_state: '',
 };
 
 type Props = {
--- a/src/ui/common/src/components/integrations/cards/databricksCard.tsx
+++ b/src/ui/common/src/components/integrations/cards/databricksCard.tsx
@@ -22,6 +22,10 @@ export const DatabricksCard: React.FC<DatabricksCardProps> = ({
         <strong>Access Token: </strong>
         {config.access_token}
       </Typography>
+      <Typography variant="body2">
+        <strong>Instance Pool ID: </strong>
+        {config.instance_pool_id}
+      </Typography>
     </Box>
   );
 };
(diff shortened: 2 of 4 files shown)


src/ui/common/src/components/integrations/dialogs/lambdaDialog.tsx
```diff
@@ -5,7 +5,8 @@ import { LambdaConfig } from '../../../utils/integrations';
 import { IntegrationTextInputField } from './IntegrationTextInputField';
 
 const Placeholders: LambdaConfig = {
-  role_arn: '<my lambda role ARN>',
+  role_arn: 'arn:aws:iam::123:role/lambda-function-role-arn',
+  exec_state: '',
```
kenxu95, line 9: What is this even for? I was a little confused when reading this.

kenxu95 (approved):

src/ui/common/src/components/integrations/dialogs/lambdaDialog.tsx
hsubbaraj-spiral: We require they provide a role with permissions for us to create Lambda functions.

agiron123: @hsubbaraj-spiral Can you explain what we need the exec_state field here for?

agiron123 (approved):

src/ui/common/src/components/integrations/dialogs/lambdaDialog.tsx
kenxu95: Yeah I meant the exec_state.

hsubbaraj-spiral: I believe `exec_state` was added when we switched to connecting with Lambda asynchronously. VSCode threw an error saying the placeholder was missing a value so I added an empty placeholder for it. 

@Fanjia-Yan do we need the exec_state as part of the UI LambdaConfig? If only the backend uses the `exec_state`, we can make that field optional and remove it from the UI.

Fanjia-Yan: The exec_state is to update the status of Lambda Connection on UI so I think the `exec_state` should be there at least when we receive the "lambda_config" from server to UI. 

I think we can either hardcoded the input to lambda config exec_state as null or we can create a new something like `LambdaExecStateConfig` to receive the output from server that contains the `exec_state`?

hsubbaraj-spiral: I think it's fine to leave it as is, since we are using the value on the UI. It'll be fine to set the value to empty string on the UI at connection time.

kenxu95: I can get rid of this as part of my integration refactor. All integrations now have an exec_state on their config, so we don't need a specialized field anymore.

PR #1234, opened 2023-04-24
Implement `use` argument for LLM for decorators

cw75:
This PR implements the `use` argument that allows the user to specify the intent to use our LLM library. Under the hood, we include the `use_llm` flag into the resource struct and use this as a flag to select Docker images that have LLM Python library dependencies and Cuda drivers pre-installed.


sdk/aqueduct/__init__.py
```diff
@@ -7,6 +7,8 @@
 from aqueduct.flow import Flow
 from aqueduct.schedule import DayOfMonth, DayOfWeek, Hour, Minute, daily, hourly, monthly, weekly
 
+llm = "llm"
```
hsubbaraj-spiral, line 10: Can we declare this as a constant?

src/golang/lib/models/shared/operator/operator.go
```diff
@@ -47,6 +47,7 @@ type ResourceConfig struct {
 	MemoryMB        *int               `json:"memory_mb,omitempty"`
 	GPUResourceName *string            `json:"gpu_resource_name,omitempty"`
 	CudaVersion     *CudaVersionNumber `json:"cuda_version,omitempty"`
+	UseLlm          *bool              `json:"use_llm,omitempty"`
```
hsubbaraj-spiral, line 50: Change to `UseLLM` just to be consistent with our naming scheme.

hsubbaraj-spiral (approved):
Few small nits, but otherwise looks great!

sdk/aqueduct/__init__.py
cw75: removed this as it's not needed

cw75:
@hsubbaraj-spiral updated the implementation to add the `llm_op` API shown in [this](https://www.loom.com/share/28f7388e677e42f4a7b3ed1419b1da74) demo. If you can take a look at `llm_wrapper.py` that'd be great.

sdk/aqueduct/llm_wrapper.py
```diff
@@ -0,0 +1,220 @@
...
+
+        module = __import__("aqueduct_llm", fromlist=[llm_name])
+        llm = getattr(module, llm_name)
+
+        if "prompt" in parameters:
+            prompt = parameters["prompt"]
+            if not isinstance(prompt, str):
+                raise ValueError("The 'prompt' parameter must be a string.")
+
+            if "{text}" not in prompt:
```
hsubbaraj-spiral, line 59: Can you explain the significance of `"{text}"` is this a keyword in LLM prompts?

hsubbaraj-spiral (approved):
Yep, the new API makes sense to me, just a quick questions about the prompt.

sdk/aqueduct/llm_wrapper.py
cw75: "If the prompt contains {text}, we will replace {text} with the input string(s) before sending to the LLM. If the prompt does not contain {text}, we will prepend the prompt to the input string(s) before sending to the LLM."

PR #1242, opened 2023-04-25
Fix bug where on-demand k8s integration registration fails on Mac

cw75:
This is due to `env --version` does not work on Mac, but since `env` should be pre-installed on all common OSes, we can simply skip the check.

In addition, `cp -T` is not supported on Mac, so we implement a workaround instead. 

Tested that the new changes work both on Linux and on Mac.

### DIFF
--- a/src/golang/cmd/server/handler/cloud_integration_utils.go
+++ b/src/golang/cmd/server/handler/cloud_integration_utils.go
@@ -125,7 +125,7 @@ func setupTerraformDirectory(dst string) error {
 		return err
 	}
 
-	_, stdErr, err := lib_utils.RunCmd("cp", []string{"-rT", dynamic.TerraformTemplateDir, dst}, "", false)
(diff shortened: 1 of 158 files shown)


src/golang/cmd/server/handler/cloud_integration_utils.go
```diff
@@ -125,7 +125,16 @@ func setupTerraformDirectory(dst string) error {
 		return err
 	}
 
-	_, stdErr, err := lib_utils.RunCmd("cp", []string{"-rT", dynamic.TerraformTemplateDir, dst}, "", false)
+	_, stdErr, err := lib_utils.RunCmd(
+		"cp",
+		[]string{
+			"-R",
```
hsubbaraj-spiral, line 131: should we be using `cp -rT` here? What was the reason behind the change?

cw75: @hsubbaraj-spiral because Mac's `cp` doesn't support the `-T` flag...I can make a comment on this.

cw75:
@hsubbaraj-spiral addressed comment.

PR #1252, opened 2023-04-27
Improve logging when operator failed before generating artifact

likawind:
This PR improves logging when a operator failed without generating artifact metadata. The current situation being:
* operator failed without generating artifact metadata
* we attempt to upload artifact metadata
* we failed to read that metadata and thus result in showing a missing storage object failure in most error logs
As the improvements, we modifies the logging with minimum change on current life-cycle management behavior:
* If artif metadata is failed to read, we will log the error rather than fail the function call. We expect orchestration to handle downstream execution based on operator's state.
  * when reading the metadata, we assume empty metadata path + metadata field implies empty metadata, rather than an error state.
* In orchestrator, we log operator's failure reason together with operator's ID when wf execution is stopped due to operator failures.

ENG-2815

## Testings
We are not expecting to change any life-cycle management behavior. I'd consider it as success as long as all integration tests passed

### DIFF
--- a/src/golang/lib/workflow/artifact/artifact.go
+++ b/src/golang/lib/workflow/artifact/artifact.go
@@ -3,6 +3,7 @@ package artifact
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 
 	"github.com/aqueducthq/aqueduct/lib/database"
 	"github.com/aqueducthq/aqueduct/lib/models"
@@ -207,9 +208,7 @@ func (a *ArtifactImpl) updateArtifactResultAfterComputation(
 	execState *shared.ExecutionState,
 ) {
 	changes := map[string]interface{}{
-		models.ArtifactResultStatus:    execState.Status,
-		models.ArtifactResultExecState: execState,
-		models.ArtifactResultMetadata:  nil,
+		models.ArtifactResultMetadata: nil,
 	}
 
 	if a.Computed(ctx) {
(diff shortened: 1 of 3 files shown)


src/golang/lib/workflow/artifact/artifact.go
```diff
@@ -222,11 +221,18 @@ func (a *ArtifactImpl) updateArtifactResultAfterComputation(
 		)
 		if err != nil {
 			log.Errorf("Unable to read artifact result metadata from storage and unmarshal: %v", err)
```
kenxu95, line 223: Curious, is this error ever expected in any cases? If it's a really fatal error we should probably also add a Tip to the execution state?

src/golang/lib/workflow/artifact/artifact.go
```diff
@@ -222,11 +221,18 @@ func (a *ArtifactImpl) updateArtifactResultAfterComputation(
...
-			return
+			execState.Error.Context = fmt.Sprintf(
+				"%s\nError reading metadata: %v",
+				execState.Error.Context,
```
kenxu95, line 226: is `execState.Error != nil` in this case? I worry this will panic if there wasn't an issue with the computation, but something bad happened when we ReadFromStorage.

src/golang/lib/workflow/artifact/artifact.go
kenxu95: And make it a system error or something.

likawind: I can add a comment here

src/golang/lib/workflow/artifact/artifact.go
likawind: yeah that's a good catch, we should check the pointer state before assigning values

kenxu95 (approved):
Nice, very clean!

PR #1264, opened 2023-04-28
Allow users to specify custom Docker image when running operators on K8s

cw75:
Apologies for the long PR. Overall it contains 3 pieces.

UI: implements integration registration and editing. All of UI code changes are boiler plate that's borrowed from other integrations. In the future we should refactor and consolidate credentials for S3, AWS, and ECR.

SDK: contains similar logic as UI, plus allowing users to specify custom Docker image via the `image` field in the decorator. We also add syntax sugar `client.integration("ecr").image("image:tag")` that generates a dictionary that can serve as input to the `image` field above. This API also sends a request to the backend to validate if the image is accessible and returns the full url.

Backend:
1. A new route to serve the GetImageURL request from the SDK.
2. Helper functions to retrieve ECR credentials and refresh token ECR (`container_registry/ecr.go`).
3. k8s secret creation and deletion logic during job launch and teardown.


sdk/aqueduct/decorator.py
```diff
@@ -451,6 +454,50 @@ def _update_operator_spec_with_resources(
...
+        connected_integrations = globals.__GLOBAL_API_CLIENT__.list_integrations()
+        if registry_name not in connected_integrations.keys():
+            raise InvalidUserArgumentException(
+                "Registry name `%s` is not one of the connected resources." % registry_name,
+            )
+
+        if url is None:
+            raise InvalidUserArgumentException("`url` must be specified when `image` is set.")
+
+        if len(url.split(":")) != 2:
```
hsubbaraj-spiral, line 491: Reminder here to auto-add the `latest` tag when there is no tag provided by the user.

src/golang/lib/container_registry/ecr.go
```diff
@@ -0,0 +1,225 @@
...
+	log "github.com/sirupsen/logrus"
+)
+
+type ecrCredentials struct {
+	Token         string
+	ExpireAt      int64
+	ProxyEndpoint string
+}
+
+func createAWSSessionFromAccessKey(accessKeyID, secretAccessKey, region string) (*session.Session, error) {
```
hsubbaraj-spiral, line 31: Can we define these two functions in some shared lib? There are other places where we'll create sessions (I think S3 storage is one of them), and it'll be good to have one set of functions that handle the connection w/ varying credentials.

hsubbaraj-spiral (commented):
Looks great! We need to document our current Docker image and an example on how to modify it (maybe the LLM image?) in our docs. Just a few nits noted in the comments.

hsubbaraj-spiral (approved):

src/golang/lib/container_registry/ecr.go
cw75: The assumption is a bit different between the two. Specifically, when creating AWS session for S3, we assume we already has `region` available whereas for ECR, we need to retrieve it first. Then the only reusable part seems to be:
```
	sess, err := session.NewSession(&aws.Config{
		Region: aws.String(s3Config.Region),
		Credentials: credentials.NewSharedCredentials(
			s3Config.CredentialsPath,
			s3Config.CredentialsProfile,
		),
	})
```
Which is already a single function call so for now I don't feel super necessary to further put them into a shared lib.

PR #1232, opened 2023-04-21
ENG-2767 react hook form submission

agiron123:
## Describe your changes and why you are making these changes
Putting up a quick draft to review my work and collect thoughts on how to best refactor and document a reference implementation.

## Checklist before requesting a review
- [x] I have created a descriptive PR title. The PR title should complete the sentence "This PR...".
- [x] I have performed a self-review of my code.
- [x] I have included a small demo of the changes. For the UI, this would be a screenshot or a Loom video.
- [x] If this is a new feature, I have added unit tests and integration tests.
- [x] I have run the integration tests locally and they are passing.
- [x] I have run the linter script locally (See `python3 scripts/run_linters.py -h` for usage).
- [x] All features on the UI continue to work correctly.
- [x] Added one of the following CI labels:
    - `run_integration_test`: Runs integration tests
    - `skip_integration_test`: Skips integration tests (Should be used when changes are ONLY documentation/UI)

PR #1266, opened 2023-05-01
Implement the new k8s integration registration flow

cw75:
The functionality is there, but the code should need some refactoring. @agiron123 feedback welcome.

Loom demo: https://www.loom.com/share/8f7b7d0ff929466bb28e69f19f1e6292


kenxu95 (approved):
I'm good with this. Just make sure you rebase and merge onto `ui_reframing` branch.

agiron123:
There's quite a bit conflicting with the ui forms refactoring work that I've done here. I'm not really sure how to best proceed at the moment.

cw75:
@agiron123 I'm not exactly sure where the conflicts are without more details on the form refactoring work. Since we need to ship this today, I suggest we merge this version and create a task that tracks making this work with the new react form, and we can prioritize it and ship it in the next release.

src/ui/common/src/components/integrations/addIntegrations.tsx
```diff
@@ -144,43 +158,227 @@ const AddIntegrationListItem: React.FC<AddIntegrationListItemProps> = ({
...
-        anchorOrigin={{ vertical: 'top', horizontal: 'center' }}
-        open={showSuccessToast === service}
-        onClose={handleSuccessToastClose}
-        key={'integrations-dialog-success-snackbar'}
-        autoHideDuration={6000}
-      >
-        <Alert
+    );
+  } else {
+    const handleRegularK8s = () => {
```
agiron123, line 200: This Kubernetes dialog should probably go into it's own component. 

After doing so, I'd have the if statement above just check for Kubernetes and render the KubernetesDialog component. No need for an else since you're just returning.


Also, should we choose to use this multistep pattern more often in the future, it might be a good idea to use the Stepper component to show that there are multiple steps to connect.
https://mui.com/material-ui/react-stepper/#horizontal-stepper

agiron123: Something that we want to avoid in the future is checking for the service name and then choosing which component to show. 

We've already got a SupportedIntegrations map which contains the integrations that we support, so we can put the dialogs that we wish to show in there. This work is coming in #1232

PR #1269, opened 2023-05-02
Create user supplied root directory on s3 bucket if it does not exist

jpurusho65:
Create user supplied root directory (path) on s3 storage bucket if one does not exist. This object path is used for storing Aqueduct workflow artifacts.

ENG-1269

- [ x] I have created a descriptive PR title. The PR title should complete the sentence "This PR...".
- [x ] I have run the integration tests locally and they are passing.
- [x ] All features on the UI continue to work correctly.

### DIFF
--- a/src/python/aqueduct_executor/operators/connectors/data/s3.py
+++ b/src/python/aqueduct_executor/operators/connectors/data/s3.py
@@ -1,4 +1,5 @@
 import json
+from http import HTTPStatus
 from typing import Any, Dict, List, Optional
 
 import numpy as np
@@ -7,6 +8,8 @@ from aqueduct.utils.serialization import deserialize
 from aqueduct_executor.operators.connectors.data import connector, extract, load
 from aqueduct_executor.operators.connectors.data.config import S3Config
 from aqueduct_executor.operators.connectors.data.s3_serialization import (
+    S3InsufficientPermissionsException,
+    S3RootFolderCreationException,
     S3UnknownFileFormatException,
     S3UnsupportedArtifactTypeException,
     _s3_deserialization_function_mapping,
@@ -38,13 +41,27 @@ class S3Connector(connector.DataConnector):
                 % str(e)
             )
 
-        # Check that any user-supplied root directory exists.
-        if self.root_dir != "":
-            # If nothing is returned by this filter call, then the directory does not exist.
-            if len(list(self.s3.Bucket(self.bucket).objects.filter(Prefix=self.root_dir))) == 0:
-                raise Exception(
-                    "Supplied root directory `%s` does not exist in bucket %s."
-                    % (self.root_dir, self.bucket)
+        # Check that any user-supplied root directory exists. If the object key name does not exist,
+        # create a new one.
+        try:
+            if self.root_dir != "":
+                # If nothing is returned by this filter call, then the directory does not exist.
+                if len(list(self.s3.Bucket(self.bucket).objects.filter(Prefix=self.root_dir))) == 0:
+                    if not self.root_dir.endswith("/"):
+                        self.root_dir = self.root_dir + "/"
+                    bucket = self.s3.Bucket(self.bucket)
+                    bucket.put_object(Bucket=self.bucket, Key=self.root_dir)
+        except ClientError as e:
+            status_code = e.response["ResponseMetadata"]["HTTPStatusCode"]
+            if status_code == HTTPStatus.FORBIDDEN:
+                raise S3InsufficientPermissionsException(
+                    "The specified root folder {} does not exist and you do not have permission to modify the bucket {} to create the folder.".format(
+                        self.root_dir, self.bucket
+                    )
+                )
+            else:
+                raise S3RootFolderCreationException(
+                    "Failed to create root folder in bucket {} with {}".format(self.bucket, str(e))
                 )
 
     def discover(self) -> List[str]:
--- a/src/python/aqueduct_executor/operators/connectors/data/s3_serialization.py
+++ b/src/python/aqueduct_executor/operators/connectors/data/s3_serialization.py
@@ -160,6 +160,14 @@ class S3UnknownFileFormatException(Exception):
     pass
 
 
+class S3InsufficientPermissionsException(Exception):
+    pass
+
+
+class S3RootFolderCreationException(Exception):
+    pass
+
+
 def artifact_type_to_s3_serialization_type(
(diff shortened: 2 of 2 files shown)


src/python/aqueduct_executor/operators/connectors/data/s3.py
```diff
@@ -38,14 +38,15 @@ def authenticate(self) -> None:
...
             # If nothing is returned by this filter call, then the directory does not exist.
             if len(list(self.s3.Bucket(self.bucket).objects.filter(Prefix=self.root_dir))) == 0:
-                raise Exception(
-                    "Supplied root directory `%s` does not exist in bucket %s."
-                    % (self.root_dir, self.bucket)
-                )
+                if not self.root_dir.endswith("/"):
+                    self.root_dir = self.root_dir + "/"
+                bucket = self.s3.Bucket(self.bucket)
+                bucket.put_object(Bucket=self.bucket, Key=self.root_dir)
```
kenxu95, line 49: Curious what happens if the user doesn't sufficient permissions? I'd like us to surface that error a little more nicely here.

kenxu95 (changes requested):
Can you add a test for this in `integration_tests/sdk/data_integration_tests/s3.py`? Essentially just write and read to a directory that doesn't exist.

kenxu95:
Ah yes good question, so you'll want to take a look at `setup_storage_layer()` in `setup_integration.py`. The semantics I'd want is for the user to be able to define a section in `test-config.yml` like:
```
storage:
    test_s3: 
          root_dir: path/to/dir  # this is currently the unsupported part.
```

This will involve fetching information from both the test-credentials.yml and test-config.yml in that `setup_storage_layer()` method.

kenxu95 (commented):
Bumping off of queue for now while the testing gets sorted.

src/python/aqueduct_executor/operators/connectors/data/s3.py
```diff
@@ -38,13 +41,25 @@ def authenticate(self) -> None:
...
+            if self.root_dir != "":
+                # If nothing is returned by this filter call, then the directory does not exist.
+                if len(list(self.s3.Bucket(self.bucket).objects.filter(Prefix=self.root_dir))) == 0:
+                    if not self.root_dir.endswith("/"):
+                        self.root_dir = self.root_dir + "/"
+                    bucket = self.s3.Bucket(self.bucket)
+                    bucket.put_object(Bucket=self.bucket, Key=self.root_dir)
+        except ClientError as e:
+            status_code = e.response["ResponseMetadata"]["HTTPStatusCode"]
+            if status_code == HTTPStatus.FORBIDDEN:
```
kenxu95, line 56: interesting, is this condition guaranteed to be a permissions error?

jpurusho65: Yeah. If you access s3 as a client (as in bot3.client('s3'), then exception classes are very different.
In our case, we are accessing as resource (s3.resource('s3'), when user does not permission, we get 403 (forbidden) http response code.

jpurusho65:
> Ah yes good question, so you'll want to take a look at `setup_storage_layer()` in `setup_integration.py`. The semantics I'd want is for the user to be able to define a section in `test-config.yml` like:
> 
> ```
> storage:
>     test_s3: 
>           root_dir: path/to/dir  # this is currently the unsupported part.
> ```
> 
> This will involve fetching information from both the test-credentials.yml and test-config.yml in that `setup_storage_layer()` method.

Can I create a sub issue for this and fix that separately?

jpurusho65:
> Ah yes good question, so you'll want to take a look at `setup_storage_layer()` in `setup_integration.py`. The semantics I'd want is for the user to be able to define a section in `test-config.yml` like:
> 
> ```
> storage:
>     test_s3: 
>           root_dir: path/to/dir  # this is currently the unsupported part.
> ```
> 
> This will involve fetching information from both the test-credentials.yml and test-config.yml in that `setup_storage_layer()` method.

done

integration_tests/sdk/data_integration_tests/s3_test.py
```diff
@@ -569,3 +569,9 @@ def test_s3_save_failure(client, data_integration):
         InvalidUserArgumentException, match="Unsupported S3 file format `wrong format`."
     ):
         save(data_integration, hotel_reviews, generate_table_name(), format="wrong format")
+
+
+def test_s3_root_folder_present(client, data_integration):
```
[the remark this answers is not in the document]
kenxu95: Actually, because root_dir only matters as artifact storage, it may be pretty hard to test as a specific case... Maybe you can just do a manual `test-config.yml` test run with a custom `root_dir` and verify that it passes?

integration_tests/sdk/setup_integration.py
```diff
@@ -173,12 +173,28 @@ def _setup_relational_data(client: Client, db: RelationalDBResource) -> None:
...
         try:
-            s3.file(object_name, artifact_type=ArtifactType.TABLE, format="parquet")
+            root_dir = "/foo/"
+            config_data = _parse_config_file()
+            if "storage" in config_data:
+                if "root_dir" in config_data["storage"][integration_name]:
+                    root_dir = config_data["storage"][integration_name]["root_dir"]
+                    if not root_dir.endswith("/"):
+                        root_dir += "/"
+            s3.file(
```
kenxu95, line 195: ahhh I see what happened! Shoot I pointed you to this method instead of `setup_data_integrations()`, which is where this change should go. The `root_dir` should only matter when a connect a resource. Afterwards, all the s3.file() stuff should be relative to that (I think?)

kenxu95 (commented):
So I realize now that the testing work might be a bit confusing, sorry about that!! I think the S3 code looks good so can we actually just roll back to that and I'll stamp this PR? Let's just make a task for adding the `root_dir` to `test-config.yml` and put it on the backlog for now.

integration_tests/sdk/data_integration_tests/s3_test.py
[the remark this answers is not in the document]
jpurusho65: I have backed out the test-integration changes, and will create a separate issue for that.

kenxu95 (approved):
Thanks! Sorry about the confusion.

jpurusho65:
> Thanks! Sorry about the confusion.

NP Thanks. Will merge once integration test passes. I have opened a new issue for tracking the integration test, we can chat how we go about doing that.

PR #1289, opened 2023-05-04
Change aqueduct_demo name to Demo

kenxu95:
This PR does a number of things;
1) For aethestic purposes, we now call the demo db "Demo" instead of "aqueduct_demo".
2) We delete the "AqueductDemo" service type, which is a vestige of old and never used.
3) We add some logic to aqueduct startup, so that if a SQLite integration named "aqueduct_demo" exists, we'll delete it, with the assumption that the "Demo" entry will be created right afterwards. This, of course, means that we do not allow the user to create SQLite integrations named "aqueduct_demo".
4) The user can still use `client.resource("aqueduct_demo")` if they want. We'll translate it to "Demo" implicitly for them, after printing out a warning.
5) We remove the blocking of anything named "aqueduct_demo" on the frontend, and instead like the connect_integration route handle the validation. This improves error surfacing.

Screen Shot 2023-05-04 at 3 15 50 PM

ENG-2824


agiron123 (approved):
Frontend should be okay ... I think. I'm still working on getting the new Kubernetes dialog to work with my forms refactor :sadpanda:

PR #1295, opened 2023-05-08
[3/n] Go binary and executor changes for handling operator pypi version mismatch

likawind:
This PR introduces proper  go binary changes to handle operator version mismatch.

* added a `versionTag` field in global config for job schedulers to set up pypi packages for various engines. Engines uses slightly different way to deal with this new flag. 

We use gloabl config instead of server flag mainly to avoid massive code change that requires passing the flag around the system.

* We will add gh action to write this flag during server setup stage.

* added executor flag `--version-tag` that triggers installation from test pypi when enabled. This flag is not used in all engines as some (e.g. conda) has test pypi configured elsewhere. This change is still necessary for dkr based engines including k8s and lambda.


src/python/aqueduct_executor/operators/airflow/dag.template
```diff
@@ -24,6 +24,24 @@ VENV_REQUIREMENTS=[
 ]
 
 def invoke_task(spec, **kwargs):
+    '''
```
likawind, line 27: may be a blocker if not able to install from python process

kenxu95:
This looks fine to me but I'm not confident my review is very high quality - would highly recommend @saurav-c and @hsubbaraj-spiral take lead review here.

saurav-c:
@likawind  Heads up taking a look at this and related PRs this morning. Sorry for delay!

cw75 (approved):
LGTM overall. @likawind just making sure we validated that installing the package within the same Python process works without having to do anything extra right?

src/python/aqueduct_executor/operators/airflow/dag.template
```diff
@@ -24,6 +24,28 @@ VENV_REQUIREMENTS=[
...
+        install_process = subprocess.run([
+            sys.executable,
+            "-m",
+            "pip",
+            "install",
+            "--index-url",
+            "https://test.pypi.org/simple/",
+            "--extra-index-url", # allows dependencies from pypi
+            "https://pypi.org/simple",
+            f"aqueduct-ml=={{ version_tag }}",
```
saurav-c, line 43: Please change this to `"aqueduct-ml=={{ version_tag }}"` The prefix of `f` is not necessary since this is a jinja template. Found this during my testing,

saurav-c (changes requested):

src/python/aqueduct_executor/operators/airflow/dag.template
likawind: Thanks! Updated

saurav-c (commented):
Unblocking based on changes I requested

src/golang/lib/spark/environments.go
```diff
@@ -90,6 +90,8 @@ func RunSparkImage(
 		fmt.Sprintf("AWS_REGION=%s", region),
 		"-e",
 		fmt.Sprintf("S3_BUCKET=%s", bucket),
+		"-e",
+		fmt.Sprintf("VERSION_TAG=\"%s\"", config.VersionTag()),
```
hsubbaraj-spiral, line 94: Reminder to remove the quotes around the version number!

hsubbaraj-spiral (approved):

PR #1305, opened 2023-05-10
Implement feature that read and write Table Artifact in parquet format

Fanjia-Yan:
WIP: 
1. Handle outdated json serialized artifact. UI and SDK is good. Still need some work on server(Wednesday)
2. Test against S3 metadata storage and maybe compute engine + old table artifact+potential integration test(Thursday)
3. Benchmark parquet vs. json performance improvement(Friday)

Changes in 
1. SDK: (1) change serialization method in `serialization.py` specifically read and write table. (2) SDK uses `get_artifact_result_content` handler to get artifact info from flow-run. We need to adapt to the new return value

2. Executor: Executor uses serialization from `sdk/serialization.py` so we don't have to change anything here.

3. Server: (1)We uses https://github.com/xitongsys/parquet-go to read Parquet bytes string as a parquet reader and output json data.(2) We also construct schema from the parquet file metadata for UI display.

4. UI: We receive from `get_artifact_result_content` handler of a schema and data. The data format is consistent with our current UI but we need some processing for schema. 


src/golang/lib/workflow/artifact/artifact.go
```diff
@@ -376,31 +378,44 @@ func (a *ArtifactImpl) SampleContent(ctx context.Context) ([]byte, bool, error)
...
+		for i := 0; i < schemaFromParquet.NumField(); i++ {
+			schema[schemaFromParquet.Field(i).Name] = schemaFromParquet.Field(i).Type.String()[1:]
 		}
 
-		t.Data = t.Data[:sampleTableRow]
-		downsampledContent, err := json.Marshal(t)
+		t.Data = parquetTable
+		t.Schema = schema
+
+		jsonTable, err := json.Marshal(t)
```
saurav-c, line 410: Instead of JSON marshaling the Parquet data, can we convert the Parquet data into JSON. This would allow us to avoid any SDK or UI side changes for reading table artifacts for now.

PR #1320, opened 2023-05-15
Add ability to parameterize the save operator

kenxu95:
@likawind and @hsubbaraj-spiral as lead reviewers. Tagging @eunice-chan since you're more familiar with the `list_saved_objects()` part. 
@vsreekanti to watch the demo.

From an implementation perspective, for parameterized saves, we set the table name on the spec to a empty string if it's parameterized. This table name is then filled in on the spec at runtime, based on the input parameter, in the python executor.

The write part of this is actually quite easy. The tough part was getting this to work with read part of it, which is necessary for `delete_workflow()` and `list_saved_objects()`. This is because both these code paths rely on the LoadSpec in the database, which is empty. This means we have to do a ton of custom logic to go a find the input parameter artifact to these parameterized saves, and find the history of all parameter values of that artifact. See `GetDistinctSaveOpsByWorkflow()` for this logic.

ENG-2675

## Loom demo (if any)
https://www.loom.com/share/1f96a44998a042f39bd344aad6306810

### DIFF
--- a/src/golang/cmd/server/handler/delete_workflow.go
+++ b/src/golang/cmd/server/handler/delete_workflow.go
@@ -4,9 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/aqueducthq/aqueduct/lib/functional/slices"
-	"github.com/aqueducthq/aqueduct/lib/models"
-	"github.com/aqueducthq/aqueduct/lib/models/views"
 	"net/http"
 	"time"
 
@@ -16,9 +13,12 @@ import (
 	"github.com/aqueducthq/aqueduct/lib/database"
 	"github.com/aqueducthq/aqueduct/lib/engine"
 	exec_env "github.com/aqueducthq/aqueduct/lib/execution_environment"
+	"github.com/aqueducthq/aqueduct/lib/functional/slices"
 	"github.com/aqueducthq/aqueduct/lib/job"
+	"github.com/aqueducthq/aqueduct/lib/models"
 	"github.com/aqueducthq/aqueduct/lib/models/shared"
 	"github.com/aqueducthq/aqueduct/lib/models/shared/operator/connector"
+	"github.com/aqueducthq/aqueduct/lib/models/views"
 	"github.com/aqueducthq/aqueduct/lib/repos"
 	"github.com/aqueducthq/aqueduct/lib/vault"
 	"github.com/aqueducthq/aqueduct/lib/workflow/operator/connector/auth"
@@ -214,11 +214,11 @@ func (h *DeleteWorkflowHandler) Perform(ctx context.Context, interfaceArgs inter
 					}
 
 					for _, saveOp := range saveOpsList {
-						saveOpsByIntegrationName[integrationName] = append(saveOpsByIntegrationName[integrationName], saveOp)
+						saveOpsByIntegrationName[saveOp.IntegrationName] = append(saveOpsByIntegrationName[saveOp.IntegrationName], saveOpsList...)
 					}
 				}
 
-				// Check for existance in the parameter-expanded list.
+				// Check for existence in the parameter-expanded list.
 				for _, saveOp := range saveOpsByIntegrationName[integrationName] {
 					relationalLoad, ok := connector.CastToRelationalDBLoadParams(saveOp.Spec.Parameters)
 					if !ok {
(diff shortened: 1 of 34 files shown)


src/golang/cmd/server/handler/delete_workflow.go
```diff
@@ -183,26 +193,65 @@ func (h *DeleteWorkflowHandler) Perform(ctx context.Context, interfaceArgs inter
...
+				if saveOpsByIntegrationName == nil {
+					saveOpsByIntegrationName = make(map[string][]views.LoadOperator)
+					saveOpsList, err := GetDistinctSaveOpsByWorkflow(
+						ctx,
+						args.WorkflowID,
+						h.OperatorRepo,
+						h.DagRepo,
+						h.ArtifactResultRepo,
+						h.Database,
+					)
```
kenxu95, line 211: There may be an argument for using `GetDistinctSaveOpsByWorkflow` from the start, instead of going through `h.OperatorRepo.GetLoadOpsByWorkflowAndIntegration()` and then falling back to this method, but I don't understand the difference super well and didn't want to rock the boat too much. cc: @eunice-chan

eunice-chan: I think you should be able to use `GetDistinctSaveOpsByWorkflow` from the get-go. For consistency's sake it might be be good to call save ops, load ops.

kenxu95: Updated. Would appreciate another look! In particular I added a new `NonRelational` cast.

vsreekanti:
The demo looks good to me! One thing we might want to do in follow-on work is allow string interpolation in the save operator like we do in the load operator, so I can do something like `table_{{param}}`. Can we create a followup task to track that?

likawind (approved):
Approving to unblock. Based on the demo, we could run another test using regular save op to make sure we don't have any regression. Two other comments that I don't think are blocking this PR:
* What does it look like when you click the operator node? Does it show the parameter as a input in `inputs` section?
* It appears we don't do string replacement like how we use in extract operators. I think it's fine as an early feature and if it fit's Yext's current need. I'm wondering if we should have a plan making it consistent with extracts. It would enable use cases like `save(table="my_results_{{ today }}")` .

src/golang/cmd/server/handler/list_workflow_objects.go
```diff
@@ -80,16 +88,160 @@ func (h *ListWorkflowObjectsHandler) Prepare(r *http.Request) (interface{}, int,
...
+
+	// Use the save operator's ID as the primary key for organization when fetching the artifacts and then their corresponding
+	// artifact results.
+	paramArtifactIDBySaveOpID := make(map[uuid.UUID]uuid.UUID, len(saveOps))
+	for _, saveOp := range saveOps {
+		if len([]uuid.UUID(saveOp.Inputs)) < 2 {
+			return nil, errors.Newf("Expected parameterized save operator %s to have multiple inputs!", saveOp.ID)
+		}
+
+		tableNameParamArtifactID := saveOp.Inputs[0]
```
hsubbaraj-spiral, line 159: How do we know the parameter artifact is the first input to the save operator?

hsubbaraj-spiral (approved):
The logic makes sense to me. Thanks for extensively commenting on GetDistinctLoadOpsByWorkflow, it helped clarify my questions. Left one question as a comment. Would also like to see some testing (unit, backend integration, or sdk integration test) but I understand that may not happen with this PR given the time constraint. Maybe we can come back to that in a later PR.

kenxu95:
wrt to the string interpolation requests, the task is here: https://linear.app/aqueducthq/issue/ENG-2982/add-documentation-for-parameterized-sql-operators Note that we don't really use the `{{...}}` syntax anymore for interpolating extracts.

> What does it look like when you click the operator node? Does it show the parameter as input in inputs section?
Yeah I believe so.

> Would also like to see some testing (unit, backend integration, or sdk integration test) but I understand that may not happen with this PR given the time constraint.
There should be a fairly extensive integration test in this PR.

src/golang/cmd/server/handler/list_workflow_objects.go
kenxu95: That's just always the order that the save() on the SDK adds it in. I'll add a comment in both places clarifying this assumption.

PR #1339, opened 2023-05-18
[4/5] Replace all Golang + UI references to integration

kenxu95:
Turns out its a lot faster to simply rename everything all at once, and then check that nothing is messed up. For this PR, we do a dumb replace "integration" and "Integration" in all the code and filenames in src/golang and src/ui.

What is remaining
[5/5] The migration of all python code. This code is a little trickier to migrate since "integration tests" is a valid phrase.

ENG-2780


likawind:
I don't think we need to rename historical data versions for migrator? History stays in history :)

likawind (approved):

src/golang/lib/models/shared/engine.go
```diff
@@ -37,7 +37,7 @@ type AqueductCondaConfig struct {
 }
 
 type AirflowConfig struct {
-	ResourceID uuid.UUID `json:"integration_id"  yaml:"integration_id"`
```
saurav-c, line 40: Is this migration going to be taken care of as part of PR 6?

kenxu95: yep!

kenxu95: @saurav-c actually I think we're gonna give up on renaming the fields themselves due to time constraints. Synced with @likawind and it's actually quite a lot of work with low benefit, so I'll just make a follow up task.

PR #1343, opened 2023-05-18
ENG-2964 GetEnvironment RTKQuery

agiron123:
This PR moves the getEnvironment API call from the kubernetes dialog and the menusidebar to use RTKQuery. 

Also moves the getEnvironment route to the new v2 routes.

ENG-2964

### DIFF
--- a/src/golang/cmd/server/routes/routes.go
+++ b/src/golang/cmd/server/routes/routes.go
@@ -24,7 +24,7 @@ const (
 	NodeDagOperatorsRoute          = "/api/v2/workflow/{workflowID}/dag/{dagID}/node/operators"
 	NodeOperatorContentRoute       = "/api/v2/workflow/{workflowID}/dag/{dagID}/node/operator/{nodeID}/content"
 	NodesResultsRoute              = "/api/v2/workflow/{workflowID}/result/{dagResultID}/nodes/results"
-	GetServerEnvironmentRoute      = "/api/v2/environment"
+	EnvironmentRoute               = "/api/v2/environment"
 
 	// V1 routes
 	GetArtifactVersionsRoute = "/api/artifact/versions"
--- a/src/golang/cmd/server/routes/routes.go
+++ b/src/golang/cmd/server/routes/routes.go
@@ -24,6 +24,7 @@ const (
 	NodeDagOperatorsRoute          = "/api/v2/workflow/{workflowID}/dag/{dagID}/node/operators"
 	NodeOperatorContentRoute       = "/api/v2/workflow/{workflowID}/dag/{dagID}/node/operator/{nodeID}/content"
 	NodesResultsRoute              = "/api/v2/workflow/{workflowID}/result/{dagResultID}/nodes/results"
+	GetServerEnvironmentRoute      = "/api/v2/environment"
 
 	// V1 routes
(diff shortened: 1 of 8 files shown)


src/golang/cmd/server/routes/routes.go
```diff
@@ -24,6 +24,7 @@ const (
 	NodeDagOperatorsRoute          = "/api/v2/workflow/{workflowID}/dag/{dagID}/node/operators"
 	NodeOperatorContentRoute       = "/api/v2/workflow/{workflowID}/dag/{dagID}/node/operator/{nodeID}/content"
 	NodesResultsRoute              = "/api/v2/workflow/{workflowID}/result/{dagResultID}/nodes/results"
+	GetServerEnvironmentRoute      = "/api/v2/environment"
```
likawind, line 27: We will have one route handle multiple methods in v2 moving forward. So it's sufficient to call it `EnvironmentRoute`.

src/golang/cmd/server/routes/routes.go
```diff
@@ -73,6 +74,5 @@ const (
 	GetWorkflowDagResultRoute    = "/api/workflow/{workflowId}/result/{workflowDagResultId}"
 	GetWorkflowHistoryRoute      = "/api/workflow/{workflowId}/history"
 
-	GetServerVersionRoute     = "/api/version"
```
likawind, line 76: Is any SDK code still using this route? To be safe, we could keep these routes and map them to the same handler (similar to workflow trigger / delete / edit routes)

src/ui/common/src/components/layouts/menuSidebar.tsx
```diff
@@ -123,22 +123,19 @@ const MenuSidebar: React.FC<{
   const [versionNumber, setVersionNumber] = useState('');
   const location = useLocation();
 
+  const { data } = useEnvironmentGetQuery({ apiKey: user.apiKey } as any, {
```
likawind, line 126: no need to cast to `any` and you probably don't need the `skip` here, as the user is passed by caller

src/ui/common/src/components/layouts/menuSidebar.tsx
```diff
@@ -123,22 +123,19 @@ const MenuSidebar: React.FC<{
...
   useEffect(() => {
-    async function fetchVersionNumber() {
-      const res = await fetch(`${apiAddress}/api/version`, {
-        method: 'GET',
-        headers: { 'api-key': user.apiKey },
-      });
-      const versionNumberResponse = await res.json();
-      setVersionNumber(versionNumberResponse.version);
+    if (data) {
+      setVersionNumber(data.version);
```
likawind, line 136: will the user or other things set the version number? Otherwise we can just do `const versionNumber = data?.version`

src/golang/cmd/server/routes/routes.go
agiron123: Yes, the version route does have a usage. I will keep this one in place.

src/ui/common/src/components/layouts/menuSidebar.tsx
agiron123: Nope, I will remove this useEffect and state variable.
Vestiges here from the infinite loop that I was facing with the Kubernetes dialog :)

likawind (approved):

PR #1348, opened 2023-05-22
Adds Github Action for Airflow compute test

saurav-c:
This PR adds a github action for running the Airflow compute tests once a week. 

See https://github.com/aqueducthq/aqueduct/actions/runs/5043163833/jobs/9044608060 for an example run. Ignore the failures for now. These will all be filed as linear tasks that are beyond the scope of this task.

ENG 2599

### DIFF
--- a/.github/workflows/airflow-periodic-tests.yml
+++ b/.github/workflows/airflow-periodic-tests.yml
@@ -30,6 +30,10 @@ jobs:
       - name: Wait for Airflow cluster to be ready
         run: while ! echo exit | nc localhost 8000; do sleep 1; done
       
+      - name: Print the test-creds (1)
+        working-directory: integration_tests/sdk
+        run: cat test-credentials.yml
+      
       - name: Get Airflow password
         run: |
           airflow_password=$(cat ~/airflow/standalone_admin_password.txt)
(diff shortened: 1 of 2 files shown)


.github/workflows/periodic-integration-tests.yml
saurav-c, line 1: @reviewers Ignore these changes. This is just a mirror copy of the other file because github actions needs the file to already exist on main in order to test out an action.

.github/workflows/airflow-periodic-tests.yml
```diff
@@ -0,0 +1,77 @@
...
+      - name: Run the SDK Integration Tests
+        working-directory: integration_tests/sdk
+        run: pytest aqueduct_tests/ -rP -vv -n 1
+
+      - uses: ./.github/actions/upload-artifacts
+        if: always()
+        with:
+          prefix: Airflow Compute
+
+      # Sets it as an environmental variable.
```
saurav-c, line 59: I'll uncomment before merging

likawind (changes requested):
Mostly looks good! A few high / low level questions:
* confirming you are adding back other tests in  `periodic-integration-tests.yml`?
* should we publish pypi and start the test like how we do it in k8s / spark? What's the cadence of this test? If the same as k8s (and other periodic ones), we could add it as another job in `periodic-integration-tests` with `publish-pypi` as a dependency.
* I see a step copying over DAG template. I'm wondering if we should do this step in `install_local.py -e` rather than just in testing env?

saurav-c:
@likawind 
1. Yes, this is just for testing.
2. Sure, we can.
3. Sure, we can.

.github/workflows/airflow-periodic-tests.yml
saurav-c: Actually will leave this commented out until I finish triaging all of the test case failures. Otherwise there will be a lot of noise generated. See https://linear.app/aqueducthq/issue/ENG-3009/investigate-airflow-integration-test-failures-and-triage-accordingly

saurav-c:
@likawind Addressed everything, see reason why I am leaving slack notif commented out for now.
The Airflow DAG is already copied as part of install_local so no changes needed there.

likawind (approved):
:shipit: nice work!

PR #1374, opened 2023-05-25
ENG 3008 Fix registered with no run workflows

eunice-chan:
If there is no run, there is no dagResult to get the dag id from. Set the latest dag to the 0-th dag in workflow dags if there is no dag result.

ENG 3008

## Loom demo (if any)
no_run


src/ui/common/src/components/pages/workflows/index.tsx
```diff
@@ -148,13 +149,25 @@ const WorkflowsPage: React.FC<Props> = ({ user, Layout = DefaultLayout }) => {
...
       });
+
+      const {
+        data: dags,
+        error: dagsError,
+        isLoading: dagsLoading,
+      } = useDagsGetQuery({
+        apiKey: user.apiKey,
+        workflowId: workflowId,
+      });
```
likawind, line 160: I'd suggest skipping this unless `dagResults` is succeeded but empty, in this way we can have one less API call for workflows already have results

src/ui/common/src/components/pages/workflows/index.tsx
```diff
@@ -191,7 +215,6 @@ const WorkflowsPage: React.FC<Props> = ({ user, Layout = DefaultLayout }) => {
           skip: dagResultsLoading && latestDagId,
```
likawind, line 215: I feel this is probably wrong as it will not skip if `dagResultsLoading` is false but `latestDagId` is empty. I'd just use `skip: !latestDagId` to load dag whenever the ID is available.

likawind (changes requested):
Great work! Now that things are working let's spend a little time refactoring the codes here:
- just make sure you are on latest main so that you don't see potential large merge conflicts  from recent updates
- we can move duplicated dag / result getting code to a helper like `useWorkflowLatestDagOrResult()` that takes the wfID, call internally the `useDagResults` and `useDags`, and returns (potentially undefined) the latest `dag` and `dagResult` objects. In this way the caller should only care when these objects are available:
```
const { dag, dagResult } = useWorkflowLatestDagOrResult(apiKey, wfID)
const dagID = dag?.id
const dagResultID = dagResult?.id
// do things based on whether dag / dagResut / dagID / dagResultID is available
```

src/ui/common/src/utils/shared.ts
```diff
@@ -74,6 +75,29 @@ export const getArtifactResultTableRow = (
   };
 };
 
+export function getLatestDagResult(
+  dagResults: DagResultResponse[]
+): DagResultResponse {
```
likawind, line 80: I slightly prefer return an `undefined` here if there's no result available, and let the caller decide how to deal with `undefined` result

src/ui/common/src/components/pages/workflows/index.tsx
```diff
@@ -218,21 +193,18 @@ const WorkflowsPage: React.FC<Props> = ({ user, Layout = DefaultLayout }) => {
...
-
-      let latestDagResultId;
+      const { latestDagResult, dag: noRunDag } = useLatestDagResultOrDag(
+        user.apiKey,
+        workflowId
+      );
       let latestDagId;
-      if (!dagResultsLoading && !dagResultsError && dagResults.length > 0) {
-        const latestDagResult = getLatestDagResult(dagResults);
+      let latestDagResultId;
```
likawind, line 201: I'd prefer using `const` if possible: https://stackoverflow.com/questions/41086633/why-most-of-the-time-should-i-use-const-instead-of-let-in-javascript#:~:text=Using%20const%20instead%20of%20let,somewhere%20else%20in%20the%20code.

Here we can use `const latestDagResultID = latestDagResult?.id` and `const dagId = `latestDagResult?.dag_id ?? dag?.id`

src/ui/common/src/components/pages/workflow/id/hook.ts
```diff
@@ -187,3 +188,44 @@ export function useWorkflowNodesResults(
...
+    },
+    {
+      skip: hasRuns,
+    }
+  );
+
+  let latestDagResult;
+  let dag;
+  if (hasRuns) {
+    latestDagResult = getLatestDagResult(dagResults);
```
likawind, line 222: if `getLatestDagResult`handles situations where `dagResults` is `undefined` or zero-length, you don't have to use  a `let` here and it gives you cleaner code. Also, whether `dag` is available depends on `dagResult`, which may makes the caller confusing (since if `dagResult` is available, caller have to look for `dag` using an additional `useDagGetQuery`. So I'd suggest call the `useDagGetQuery` here directly. Putting everything together this hook should look like:
```
const latestDagResult = getLatestDagResult(dagResults ?? []) // undefined if not available
const { data: dags } = useDagsGetQuery(..., {skip: // when latestDagResult is available or is not loaded, see line 74 of this file})
const dagIdFromLatestDagResult = latestDagResult?.dag_id;
const { data: dag } = useDagGetQuery({ dag_id: dagIdFromLatestDagResult}, {skip: // when dagIdFromLatestDagResult is not available})
return { latestDagResult, dag?? (dags ?? [])[0] // fall back to dags when dag from latestDagResult is not available)}
```

likawind: With this change I believe you can remove most `useDagGetQuery` in the caller code

likawind (changes requested):
Thanks for re-organizing the code! I have a few nitpicking comments to improve the simplicity, making hooks less caller-dependent, and have better use of `const` over `let`

src/ui/common/src/components/pages/workflows/index.tsx
likawind: See if this is still relevant after the suggested change in `useLatestDagResultOrDag`

likawind (approved):

PR #1376, opened 2023-05-25
ENG 3042 Use checks status in workflow status

eunice-chan:
Use checks status as part of the workflow status.

ENG 3042

## Loom demo (if any)

Succeed with warning
Screen Shot 2023-05-25 at 10 33 02 AM
Failed due to check
Screen Shot 2023-05-25 at 11 19 09 AM


src/ui/common/src/components/pages/workflows/index.tsx
```diff
@@ -149,14 +150,40 @@ const WorkflowsPage: React.FC<Props> = ({ user, Layout = DefaultLayout }) => {
...
+        .forEach((op) => {
+          const artifactId = op.outputs[0]; // Assuming there is only one output artifact
+          const level = op.spec.check.level;
+          const value = nodesResults.artifacts[artifactId]?.exec_state?.status;
+          if (value === ExecutionStatus.Failed) {
+            if (level === CheckLevel.Error) {
+              status = ExecutionStatus.Failed;
+            } else if (status !== ExecutionStatus.Failed) {
+              // CheckLevel.Warning & status !== Failed
+              status = ExecutionStatus.Warning;
```
likawind, line 182: If both error and warning exists, this will assign the status to warning if the warning one get iterated later than the error one, while the actual status shown should be error. I think an easier solution is to only do this check if the workflow status is `success` and we override it to warning based on `nodes` values

likawind: Also, strongly recommend adding a `hasWarningCheck(nodes: NodeMap): boolean` helper in `responses/nodes` over `NodesMap` object and use it here

src/ui/common/src/components/pages/workflows/index.tsx
```diff
@@ -149,14 +150,40 @@ const WorkflowsPage: React.FC<Props> = ({ user, Layout = DefaultLayout }) => {
         workflowId: workflowId,
       });
       let status = ExecutionStatus.Unknown;
-
+      let latestDagId;
+      let latestDagResultId;
       if (!dagResultsLoading && !dagResultsError && dagResults.length > 0) {
         const latestDagResult = getLatestDagResult(dagResults);
         if (latestDagResult) {
           status = latestDagResult.exec_state.status;
```
likawind, line 158: I noticed that as we keep reassign this `status` variable, we will render different workflow status as these APIs are loading. This is a slight friction, and I think we can improve this by:
* Under `components/pages/workflow/hook` put together a helper like `useWorkflowStatus(workflow, dag, dagResult) -> status | undefined` that gather the status using these queries, and returns undefined if they are still loading. We can then render until the status is available

likawind (changes requested):
I do notice some edge case where the fix may not work, and have some suggestions on better organizing the code. I'd let you decide if code reorg could come separately as they may be easier when combining with https://github.com/aqueducthq/aqueduct/pull/1374 , but do consider them as immediate follow-ups if you decide not to do them in these two PRs

eunice-chan:
Workflow statuses look good for all manual QA test flows.

image
image
image

src/ui/common/src/components/pages/workflows/index.tsx
```diff
@@ -125,14 +126,36 @@ const WorkflowsPage: React.FC<Props> = ({ user, Layout = DefaultLayout }) => {
...
+
+      const nodes = useWorkflowNodes(user.apiKey, workflowId, latestDagId);
+      const nodesResults = useWorkflowNodesResults(
+        user.apiKey,
+        workflowId,
+        latestDagResultId
+      );
+
+      const ready =
+        Object.getOwnPropertyNames(nodes.operators).length !== 0 ? true : false;
```
likawind, line 142: What does `getOwnPropertyNames` do? Since we treat this as a map it would be more clear to use `Object.keys()`. Also, you don't need the `? :` sugar here and can directly do `const ready = ().length !== 0`.

likawind (approved):
Nice, just one minor comment!

PR #1382, opened 2023-05-28
ENG 3056 Cannot delete resource when no workflow is using it

eunice-chan:
Fix bug where the user cannot delete a resource when no workflow is using it. The cause is because the confirm dialog never pops up because resource operators.status is stuck at initial. This is due to the fact `handleLoadResourceOperators` is never dispatched. To fix this, I dispatch `handleLoadResourceOperators` in the resource details page.

--

Updated to V2 route.
Tested by successfully deleting MariaDB resource.


likawind (changes requested):
@kenxu95 I'm fairly sure that you've migrated this request and we should use one of `useResourceOperatorsGetQuery` or `useResourceWorkflowsGetQuery` here: https://github.com/aqueducthq/aqueduct/blob/main/src/ui/common/src/handlers/v2/ResourceOperatorsGet.ts

likawind (approved):
Great work and thanks for addressing feedbacks! Just a reminder to update PR description and also describe that you've done some testings to verify this feature (which I trusted you did :) )

PR #1386, opened 2023-05-30
Updated terraform scripts

jpurusho65:

1. To keep the number of variables to minimum
2. Keep it compatible with EKS terraform scripts

ENG-2918


src/terraform/gke/gke.tf
```diff
@@ -5,15 +5,14 @@
...
-  node_locations = local.node_locations
+  location = var.region
   resource_labels = var.cluster_resource_labels
   
   # We can't create a cluster with no node pool defined, but we want to only use
   # separately managed node pools. So we create the smallest possible default
   # node pool and immediately delete it.
   remove_default_node_pool = true
-  initial_node_count       = 1
+  initial_node_count       = var.initial_node_count
```
cw75, line 15: we can just keep this as "1" because we are removing this default node anyway.

src/terraform/gke/gke.tf
```diff
@@ -5,15 +5,14 @@
 resource "google_container_cluster" "primary" {
   name     = var.cluster_name
   description = var.description
-  location = local.location
-  node_locations = local.node_locations
+  location = var.region
```
cw75, line 8: Is our `zone` variable being used anywhere?

cw75: In my original code I had `node_locations = [var.zone]`. Maybe we can bring it back so that the cluster gets created in a specific zone that the user specified.

jpurusho65: > Is our `zone` variable being used anywhere?

zone is not valid for gcp resource block anymore. We use node_locations instead, whose type is a list(string). This list contains the zone names.

jpurusho65: Let me know, if you would like me to enforce type directly in resource block as in [var.zone], or define a variable of type list(string) for zone, and assign to node_locations

jpurusho65:
> @jpurusho65 let's make the following changes:
> 
> 1. I'm not sure what `initial_node_count` and `node_count` are used for, but let's use the `min_cpu_node` as the initial size of the CPU node pool.
done, node_count is needed for node pool creation

> 2. Since in this first cut we are not shipping GPU support, let's remove the relevant Terraform script for the GPU node pool (if any), and also because of that we also don't need `create_gpu_node_pool`. Let's keep the `gpu_node_type`, `min_gpu_node`, and `max_gpu_node` variables for now because they are currently passed in from the Golang side; they just won't be used in our Terraform script for now.
Sure, the block will not execute as the default value is set to false.

> 3. I'm not sure what's the goal of `cluster_resource_labels` is, but we probably don't need this for now.
They are meant for adding needed tags for the cluster, so when using gcloud cli, customers can filter output by specific resource labels. I have removed that for the initial release, we can add them later.

cw75:
@jpurusho65 looks good overall! A couple nits:
1: Let's use a default disk size of 50GB. If not I believe it'll use 15GB which is a bit too small.
2: I don't think the golang side currently pass in `description` to Terraform, so we can remove this variable and its current usage in the Terraform script. Otherwise without a default description value I don't think the script will work?

jpurusho65:
> @jpurusho65 looks good overall! A couple nits: 1: Let's use a default disk size of 50GB. If not I believe it'll use 15GB which is a bit too small. 2: I don't think the golang side currently pass in `description` to Terraform, so we can remove this variable and its current usage in the Terraform script. Otherwise without a default description value I don't think the script will work?

Done

cw75 (approved):

PR #1399, opened 2023-06-01
[2/n] Implement application logics to delete and handle deleted artifact contents

likawind:

This PR adds application logics to handle artifact content deletion if the user choose to opt-out snapshotting. The outline of the work lifecycle follows these steps:
* During wf registration, we assume client provides 
the right `should_persist` tag to each artifact based on user inputs
* During wf preview, no change is made there
* During wf execution, no change is made on the execution logic.
* During wf execution, we modified DB updates such that artifact with `should_persist` will be directly marked as 'deleted'
* After wf execution, we added a step in `eng.ExecuteWorkflow()` that deletes all necessary artf contents. Note that this cannot be done during execution as downstream op may need still need to read the contents.
* All APIs that read results returns content whenever possible regardless of exec state. Thus no change is necessary. The only caveat is that during execution, there will be time where the artf is marked as `deleted` while the content is still available, and we may want to have the client hide the content to provide consistent user view.

ENG-3000

## Tests
* Integration test should pass on this PR
* We will test through this feature in follow up PRs after client side changes are done


src/golang/lib/engine/aq_engine.go
```diff
@@ -362,6 +371,8 @@ func (eng *aqEngine) ExecuteWorkflow(
 		execState.Timestamps.FinishedAt = &now
 	}
 
+	eng.deleteTemporaryArtifactContents(ctx, dag)
```
kenxu95, line 374: hmm this should be in a defer() at the beginning of the method right? Do we not want to delete artifacts when the workflow fails?

src/golang/lib/models/shared/execution_status.go
```diff
@@ -23,7 +23,11 @@ const (
 	CanceledExecutionStatus   ExecutionStatus = "canceled"
 	FailedExecutionStatus     ExecutionStatus = "failed"
 	SucceededExecutionStatus  ExecutionStatus = "succeeded"
-	UnknownExecutionStatus    ExecutionStatus = "unknown"
+	// 'erased' refers to 'erased after success'.
```
kenxu95, line 26: What is "erased" here mean?

kenxu95: I think here is a good place to lay out what the lifecycle is, with an ascii diagram:
Registered -> Pending -> Running -> Failed
 .......................................................-> Cancelled
 .......................................................-> Succeeded -> Deleted [Optional]

src/golang/lib/workflow/artifact/artifact.go
```diff
@@ -25,6 +25,12 @@ type Artifact interface {
 	Signature() uuid.UUID
 	Type() shared.ArtifactType
 	Name() string
+	ShouldPersistContent() bool
+
+	// DeleteContent removes the artifact content from storage.
```
kenxu95, line 30: Can you note that this only deletes the latest artifact result? Also, what happens if the artifact content was never written due to workflow failure.

src/golang/lib/engine/aq_engine.go
```diff
@@ -147,6 +147,15 @@ func (eng *aqEngine) ScheduleWorkflow(
 	return nil
 }
 
+func (eng *aqEngine) deleteTemporaryArtifactContents(ctx context.Context, dag dag_utils.WorkflowDag) {
```
kenxu95, line 150: Can you note here that this is best effort. Also, that it only deletes the latest artifact result right?

kenxu95 (changes requested):
> All APIs that read results returns content whenever possible regardless of exec state. Thus no change is necessary. The only caveat is that during execution, there will be time where the artf is marked as deleted while the content is still available, and we may want to have the client hide the content to provide consistent user view.

Is this consistent view addressed in this PR, or is that a follow-up? I assume there will be additional PRs that deal with the client?

src/golang/lib/engine/aq_engine.go
likawind: I think it's a bit weird to explicitly note about the latest result. Similar to the other commend, current dag / op / artf objects have a result included and is used to handle the lifecycle of 'one workflow execution'. And all operations over these objects are tied to that execution. In this way I feel which result to delete is clear from the context? Also technically, if we reconstruct a workflow from an older run, we can use the same method to delete contents from that run as well

likawind:
@kenxu95 yes client change will be in future PRs

src/golang/lib/engine/aq_engine.go
kenxu95: > current dag / op / artf objects have a result included and is used to handle the lifecycle of 'one workflow execution'.

I think I'm confused cus you can execute the same dag / op/ artf multiple times. An artifact is different than an artifact result so as soon as I see this method signature I'm thinking "Which artifact result am I deleting". But ok I think I see what you mean, an artf in orchestration means something different than an artifact in our data model? This seemed like implicit context I wanted to call out.

likawind: I don't think our current abstraction allows us to re-execute a dag by calling `ExecuteWorkflow(dag)` multiple times? If you look at the constructor, we explicitly require a result ID to pass in to construct the `WorkflowDag` object: `NewWorkflowDag(..., dagResultID,...)` . In this way I think our abstraction is explicitly designed to have one object track one result.

I think it's a good idea to clarify this context somewhere, and I will add some comments to `WorkflowDag` object's definition. I just don't feel it's specific to this feature so we don't have to call out that on the specific API we added

kenxu95: Yeah I'm not saying I want us to change the way we execute, I just wanted that context clarified somewhere cus I do think it's an important callout. Agreed we can put it at a higher layer (maybe in orchestration or as you said, in workflowDag)

likawind: Sg! I updated comments for wfDag / artf / op objects, lmk if you have suggestions on where else we can clarify this

src/golang/lib/workflow/operator/base.go
```diff
@@ -372,6 +372,10 @@ func (bo *baseOperator) PersistResult(ctx context.Context) error {
 			artifactExecState.Status = shared.CanceledExecutionStatus
 		}
 
+		if artifactExecState.Status == shared.SucceededExecutionStatus && !outputArtifact.ShouldPersistContent() {
+			artifactExecState.Status = shared.DeletedExecutionStatus
```
kenxu95, line 376: This setting of status to Deletion before the artifact is even deleted seems confusing actually, can we not update the status in `DeleteContent()` instead? That way all the deletion logic is handled in one place.

likawind: In this way the artifact will be marked as succeeded during execution and deleted later. I'm not sure if it's going to confuse users since they will see the intermediate succeed state

kenxu95: Yes - I actually find that less confusing since it very clearly follows the lifecycle, where deleted is just a future state of succeeded. Right now, we technically would have operators using artifacts already marked as deleted, which seems weirder?

likawind: sounds good, let's do what you suggested

kenxu95 (approved):
Unblocking with a couple final thoughts. Thanks!

PR #1400, opened 2023-06-01
[3/n] Allows user to disable snapshots from SDK

likawind:
This PR adds SDK changes to disable snapshots on both artf and wf level. The API is implemented based on https://aqueducthq.slack.com/archives/C01U6RREC7M/p1685489357239879 . And the behavior looks like the following:
* user can enable / disable snapshots each individual artifact (enabled by default)
* when user turns off workflow level snapshot at `publish_flow`, this is equivalent to call `disable_snapshots` for all artifacts from non param / metrics / checks operators. In other words, if the user explicitly disable snapshots for those excluded artifacts, the option will be respected.
* See the manual QA example as a reference to the above behavior

ENG-3000

## Tests
Verified the manual QA as a part of #1401 .

### DIFF
--- a/manual_qa_tests/workflows/succeed_with_snapshots_disabled.py
+++ b/manual_qa_tests/workflows/succeed_with_snapshots_disabled.py
@@ -1,0 +1,47 @@
+import aqueduct as aq
+
+NAME = "succeed_with_snapshots_disabled"
+DESCRIPTION = """* Workflows Page: should succeed.
+* Workflow Details Page: everything should succeed except for 'check_disabled_artf' and op_disabled_artf'."""
+
+
+@aq.op(requirements=[])
+def op_disabled(df):
+    return df
+
+
+@aq.metric(requirements=[])
+def metric(df):
+    return df.shape[0]
+
+
+@aq.check(requirements=[])
+def check(count):
+    return count > 10
+
+
+@aq.check(requirements=[])
+def check_disabled(count):
+    return count < 10
+
+
+def deploy(client, resource_name):
+    resource = client.resource(resource_name)
+    reviews = resource.sql("SELECT * FROM hotel_reviews")
+    op_artf = op_disabled(reviews)
+    metric_artf = metric(op_artf)
+    check(metric_artf)
+    check_disabled_artf = check_disabled(metric_artf)
+
+    op_artf.enable_snapshot()
+    check_disabled_artf.disable_snapshot()
+
+    resource.save(op_artf, "succeed_with_snapshots_disabled_tmp", aq.LoadUpdateMode.REPLACE)
+
+    client.publish_flow(
+        artifacts=[op_artf],
+        name=NAME,
+        description=DESCRIPTION,
+        schedule="",
+        disable_snapshots=True,
+    )
--- a/manual_qa_tests/workflows/succeed_with_snapshots_disabled.py
+++ b/manual_qa_tests/workflows/succeed_with_snapshots_disabled.py
@@ -2,7 +2,10 @@ import aqueduct as aq
 
 NAME = "succeed_with_snapshots_disabled"
 DESCRIPTION = """* Workflows Page: should succeed.
-* Workflow Details Page: everything should succeed except for 'check_disabled_artf' and op_disabled_artf'."""
+* Workflow Details Page:
+  * There artifacts are shown deleted: op_disabled artf, Demo query artf
+  * Both checks should show 'passed'.
+"""
 
 
 @aq.op(requirements=[])
@@ -22,7 +25,7 @@ def check(count):
 
 @aq.check(requirements=[])
 def check_disabled(count):
-    return count < 10
+    return count > 10
 
 
 def deploy(client, resource_name):
--- a/sdk/aqueduct/artifacts/base_artifact.py
+++ b/sdk/aqueduct/artifacts/base_artifact.py
@@ -29,6 +29,9 @@ class BaseArtifact(ABC):
     def type(self) -> ArtifactType:
         return self._dag.must_get_artifact(artifact_id=self._artifact_id).type
 
+    def snapshot_enabled(self) -> bool:
+        return self._dag.must_get_artifact(artifact_id=self._artifact_id).should_persist
+
     def _get_content(self) -> Any:
         return self._content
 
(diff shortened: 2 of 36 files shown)


manual_qa_tests/workflows/succeed_with_snapshots_disabled.py
```diff
@@ -0,0 +1,47 @@
...
+def check(count):
+    return count > 10
+
+
+@aq.check(requirements=[])
+def check_disabled(count):
+    return count < 10
+
+
+def deploy(client, resource_name):
```
likawind, line 28: @vsreekanti would like your input to confirm this call chain should result in:
* reviews, op_artf and check_disabled_artf to be deleted
* outputs of `metrics` and `check` are persisted

likawind: Actually I feel allowing disabling check artf is very confusing. I'm wondering if we should just return an error if the user attempts to do so. Maybe the same applies to params.

Disabling metric artf makes sense to me though, since esp. table metrics can have sensitive content as they are just a grammar sugar over regular operators.

sdk/aqueduct/client.py
```diff
@@ -602,6 +603,9 @@ def publish_flow(
                 behavior is 'True'.
             use_local:
                 Must be set if any artifact in the flow is derived from local data.
+            disable_snapshots:
+                If set 'True', disables snapshots for all artifacts not generating from params / metrics / checks
+                by calling `.disable_snapshot()` on each of these artifacts.
```
cw75, line 608: This sentence seems fragmented and I am a bit confused about what it means. Pasting it to GPT gave me the following:
```
disable_snapshots:
    If set to 'True', this option disables snapshots for all artifacts that are not generated from parameters, metrics, or checks. It achieves this by calling the `.disable_snapshot()` method on each of these artifacts.
```
Is this what you're trying to say? @likawind

likawind: Great, thank you :)

sdk/aqueduct/artifacts/base_artifact.py
```diff
@@ -41,6 +44,12 @@ def set_operator_type(self, operator_type: OperatorType) -> None:
     def set_name(self, name: str) -> None:
         self._dag.update_artifact_name(self._artifact_id, sanitize_artifact_name(name))
 
+    def enable_snapshot(self) -> None:
+        self._dag.update_artifact_should_persist(self._artifact_id, True)
+
+    def disable_snapshot(self) -> None:
```
kenxu95, line 50: Can we use the `set_` prefix for any write methods?

sdk/aqueduct/models/dag.py
```diff
@@ -457,6 +457,20 @@ def update_artifact_name(self, artifact_id: uuid.UUID, new_name: str) -> None:
         artifact.name = new_name
         artifact.explicitly_named = True
 
+    def update_artifact_should_persist(self, artifact_id: uuid.UUID, should_persist: bool) -> None:
+        self.must_get_artifact(artifact_id).should_persist = should_persist
+
+    def disable_snapshots_from_non_params_metrics_checks(self) -> None:
```
kenxu95, line 463: I personally find this name a little confusing. I'd prefer if we kept the method more generic `disable_snapshots()` and added a nice docstring here indicating that parameters, metric and check artifacts are excluded.

kenxu95 (commented):
Can you paste a loom demo of the manual_qa_tests?

sdk/aqueduct/models/dag.py
likawind: I actually prefer the naming to be self-descriptive esp. that here the underlying assumption is not clear to whoever scan through the caller codes.

Screenshot 2023-06-05 at 2 21 06 PM

likawind: how about `disable_snapshots_not_from_params_metrics_checks` ?

kenxu95: I also prefer my methods to be meaningful, but the description needs to stop somewhere. In this case, it feels like we're leaking implementation details into the method name. Specifically, I'd rather not be listing things in my names (params, metrics, checks), since it's not scalable or very readable imo. Nbd if you want to keep it, it just jumped out at me.

likawind: I'm more worried that `disable_snapshot` is misleading because we don't actually disable snapshots. I do agree that this name is too long and what you mentioned about listing stuff makes sense. Welcome to advices as long as we somehow indicate only some snapshots get disabled here :)

sdk/aqueduct/artifacts/base_artifact.py
likawind: It's been discussed in https://aqueducthq.slack.com/archives/C01U6RREC7M/p1685489357239879 and given it's this close to release, I'd stick with the decision than making more back-and-forth to the channel. We've been using other verbs like `bound` `max` over artifacts so I don't see it as a huge misalignment as long as they start with verbs

sdk/aqueduct/models/dag.py
likawind: Updated to `disable_sensitive_snapshots`. `sensitive` may confuse reader but that should lead them to the comment where we clarify things

likawind:
@kenxu95 manual QA is updated in #1401

sdk/aqueduct/models/dag.py
kenxu95: mm does `disable_op_snapshots` or `disable_op_output_snapshots` work? No strong opinions as long as we document it in the function.

sdk/aqueduct/artifacts/base_artifact.py
kenxu95: oh I didn't realize this was user-facing actually. That's good for now yea

kenxu95 (approved):

sdk/aqueduct/models/dag.py
likawind: yeah sounds good, I guess I just don't want it look too harmless so that call reader would try to look at the docs, rather than skip over it with some wrong assumptions

cw75 (approved):

PR #1401, opened 2023-06-02
[4/n] Add UI changes to render deleted artifacts

likawind:
This PR adds necessary changes to render artifacts who has snapshots disabled. Note that we show the results that get disabled but not implying that future workflow run will have the snapshot disabled.

We also added some small fix so that we show check status that aligns with op results rather than artf results
ENG-3000

## Loom demo (if any)
Verified in manual QA that no regressions to relevant features:
https://www.loom.com/share/12a3015af65b47e6b2511c17aeab1c6b

Verified on the manual QA with disabled API calls:
Screenshot 2023-06-05 at 5 00 08 PM

### DIFF
--- a/src/ui/common/src/components/tables/CheckTableItem.tsx
+++ b/src/ui/common/src/components/tables/CheckTableItem.tsx
@@ -1,71 +1,31 @@
-import {
-  faCircleCheck,
-  faCircleExclamation,
-  faMinus,
-  faTriangleExclamation,
-} from '@fortawesome/free-solid-svg-icons';
-import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import Box from '@mui/material/Box';
 import React from 'react';
 
-import { theme } from '../../styles/theme/theme';
-import { stringToExecutionStatus } from '../../utils/shared';
+import ExecutionStatus from '../../utils/shared';
 import { StatusIndicator } from '../workflows/workflowStatus';
 
 interface CheckTableItemProps {
-  checkValue: string;
-  status?: string;
+  status: ExecutionStatus;
+  value?: string;
 }
 
 export const CheckTableItem: React.FC<CheckTableItemProps> = ({
-  checkValue,
   status,
+  value,
 }) => {
-  let iconColor = theme.palette.black;
-  let checkIcon = faMinus;
+  if (value) {
+    return <>{value}</>;
+  }
 
-  if (checkValue) {
-    switch (checkValue.toLowerCase()) {
-      case 'true': {
-        checkIcon = faCircleCheck;
-        iconColor = theme.palette.Success;
-        break;
-      }
-      case 'false': {
-        checkIcon = faCircleExclamation;
-        iconColor = theme.palette.Error;
-        break;
-      }
-      case 'warning': {
-        checkIcon = faTriangleExclamation;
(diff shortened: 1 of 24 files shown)


src/ui/common/src/components/tables/CheckTableItem.tsx
```diff
@@ -1,71 +1,31 @@
...
 
 export const CheckTableItem: React.FC<CheckTableItemProps> = ({
-  checkValue,
   status,
+  value,
 }) => {
-  let iconColor = theme.palette.black;
-  let checkIcon = faMinus;
+  if (value) {
+    return <>{value}</>;
```
agiron123, line 16: Does this need to be a Typography component? Or does the parent component take care of this?

src/ui/common/src/components/tables/CheckTableItem.tsx
```diff
@@ -1,71 +1,31 @@
...
-  } else {
-    // Check value not found, render the status indicator for this check.
-    return (
-      <StatusIndicator
-        status={stringToExecutionStatus(status)}
-        size={'16px'}
-        monochrome={false}
-      />
-    );
+  if (status) {
```
agiron123, line 19: This can be combined with the return statement below as a ternary to check for status.

```status={status ? status : ExecutionStatus.Unknown}```

File: src/ui/common/src/components/workflows/artifact/content.tsx [hunk 3]
```diff
@@ -45,7 +45,7 @@ const ArtifactContent: React.FC<Props> = ({
     );
   }
 
-  if (!content || !artifactResult) {
+  if (!content || !content.content || !artifactResult) {
```
agiron123, line 48: Could this become ```!content?.content || !artifactResult``` ?

agiron123 (commented):
Code looks good, left a couple small suggestions.
Happy to help with some manual testing after these are resolved.

src/ui/common/src/components/tables/CheckTableItem.tsx
likawind: This follows what has been used in main version that we returns the raw string for check names, etc.

likawind: We could use a Typography here but I feel that's an independent change and not necessarily need to spend much time in this PR

likawind:
Based on discussion with @vsreekanti , we make the delete artifact green but showing a different icon / description to distinguish from the succeeded one. I deliberately make the icon gray so that the node is not 'as green as' a succeeded one

src/ui/common/src/components/tables/CheckTableItem.tsx
agiron123: no worries, just checking what's going on here.

agiron123:
Nice work! Should we gray out / add some sort of opacity filter over the side sheet so that users know that the artifact has been deleted.

I find it a little hard to distinguish that this artifact is no longer there.

agiron123:
Maybe something that we can also do is add a strikethrough to the text of the artifact name in the DAG.

Open to ideas here @vsreekanti

agiron123 (approved):
Nice work! Left some comments, but we can resolve them in a later PR.

PR #1404, opened 2023-06-03
[5/n] Adds integration test for disabling snapshots

likawind:
This PR adds integration tests for disabling snapshots. To properly examine the exec states of flows, we also added exec state tracking for all artifact types.

Also included a fix for `ENG-3103` to fix datetime parsing

ENG-3000 ENG-3103

## Tests
* Integration tests on this PR should succeed, implying new feature working and no regressions.
* Manual UI testing covered in previous PR #1401
* Tested on a cleaner build to verify requirements.txt are updated (install only using `pip install .` and `install_local` script, without install from a pypi package).

### DIFF
--- a/integration_tests/sdk/aqueduct_tests/flow_test.py
+++ b/integration_tests/sdk/aqueduct_tests/flow_test.py
@@ -529,22 +529,27 @@ def test_flow_with_disabled_snapshots(client, flow_name, data_resource, engine):
     assert (
         latest_run.artifact("extract artifact").execution_state().status == ExecutionStatus.DELETED
     )
+    assert latest_run.artifact("extract artifact").get() is None
     assert (
         latest_run.artifact("op_disabled_by_wf artifact").execution_state().status
         == ExecutionStatus.DELETED
     )
+    assert latest_run.artifact("op_disabled_by_wf artifact").get() is None
     assert (
         latest_run.artifact("metric_enabled artifact").execution_state().status
         == ExecutionStatus.SUCCEEDED
     )
+    assert latest_run.artifact("metric_enabled artifact").get() == 100
     assert (
         latest_run.artifact("metric_disabled artifact").execution_state().status
         == ExecutionStatus.DELETED
     )
+    assert latest_run.artifact("metric_disabled artifact").get() is None
     assert (
         latest_run.artifact("check_enabled artifact").execution_state().status
         == ExecutionStatus.SUCCEEDED
     )
+    assert latest_run.artifact("check_enabled artifact").get()
 
 
 def test_artifact_set_name(client, flow_name, engine):
--- a/integration_tests/sdk/aqueduct_tests/flow_test.py
+++ b/integration_tests/sdk/aqueduct_tests/flow_test.py
@@ -495,6 +495,58 @@ def test_flow_list_saved_objects_none(client, flow_name, engine):
     assert len(flow.list_saved_objects()) == 0
 
 
+def test_flow_with_disabled_snapshots(client, flow_name, data_resource, engine):
+    @op
+    def op_disabled_by_wf(df):
+        return df
+
+    @metric
+    def metric_enabled(df):
+        return df.shape[0]
+
+    @metric
+    def metric_disabled(df):
+        return df.shape[0]
+
+    @check
+    def check_enabled(count):
+        return count > 10
+
+    reviews = extract(data_resource, DataObject.SENTIMENT, output_name="extract artifact")
+    op_artf = op_disabled_by_wf(reviews)
+    metric_enabled(op_artf)
+    metric_disabled_artf = metric_disabled(op_artf)
+    metric_disabled_artf.disable_snapshot()
+    check_enabled(metric_disabled_artf)
+
+    op_artf.enable_snapshot()
+    save(data_resource, op_artf)
+
+    flow = publish_flow_test(
+        client, op_artf, name=flow_name(), engine=engine, disable_snapshots=True
+    )
+    latest_run = flow.latest()
+    assert (
+        latest_run.artifact("extract artifact").execution_state().status == ExecutionStatus.DELETED
+    )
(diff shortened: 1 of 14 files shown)


integration_tests/sdk/aqueduct_tests/flow_test.py
```diff
@@ -495,6 +495,58 @@ def noop():
...
+
+    op_artf.enable_snapshot()
+    save(data_resource, op_artf)
+
+    flow = publish_flow_test(
+        client, op_artf, name=flow_name(), engine=engine, disable_snapshots=True
+    )
+    latest_run = flow.latest()
+    assert (
+        latest_run.artifact("extract artifact").execution_state().status == ExecutionStatus.DELETED
```
kenxu95, line 530: What does `.get()` return? Can we just check that `latest_run.artifact(...).get() == None`? Seems good to actually check the data directly.

sdk/aqueduct/artifacts/generic_artifact.py
```diff
@@ -62,7 +54,10 @@ def get(self, parameters: Optional[Dict[str, Any]] = None) -> Any:
         self._dag.must_get_artifact(self._artifact_id)
 
         if self._from_flow_run:
-            if self._execution_status != ExecutionStatus.SUCCEEDED:
+            if (
+                not self._execution_state
+                or self._execution_state.status != ExecutionStatus.SUCCEEDED
```
kenxu95, line 59: Um shouldn't this return None if the artifact was deleted? The error message talks about "computed successfully!" which is true when status is Deleted.

sdk/aqueduct/backend/api_client.py
```diff
@@ -669,10 +668,10 @@ def get_artifact_result_data(
         if "data" in parsed_response:
             return_value = deserialize(serialization_type, artifact_type, parsed_response["data"])
 
-        if execution_status != ExecutionStatus.SUCCEEDED:
+        if not execution_state or execution_state.status != ExecutionStatus.SUCCEEDED:
```
kenxu95, line 671: Same question here about handling of DELETED status. Can we get away with just returning `return_value` in that case (if it's None?)

sdk/aqueduct/artifacts/generic_artifact.py
likawind: That's a good catch! It's interesting that we are not handling other state though. Following the same argument I think we should. I will just address the DELETE state in this PR though

likawind:
@kenxu95 thanks for the feedbacks! updated

sdk/aqueduct/artifacts/generic_artifact.py
kenxu95: What do you mean by handling other state? None of the other states computed the artifact successfully.

sdk/aqueduct/artifacts/generic_artifact.py
```diff
@@ -59,10 +51,16 @@ def get(self, parameters: Optional[Dict[str, Any]] = None) -> Any:
             InternalServerError:
                 An unexpected error occurred in the server.
         """
+        if self._is_content_deleted():
+            return None
```
kenxu95, line 55: I think it's important to point out to the reader that execution status can never be DELETED (since we condition on status later in the method) after passing this check.. Can we make that an assertion after each of these?

sdk/aqueduct/artifacts/table_artifact.py
```diff
@@ -128,6 +129,9 @@ def head(self, n: int = 5, parameters: Optional[Dict[str, Any]] = None) -> pd.Da
             A dataframe containing the table contents of this artifact.
         """
         df = self.get(parameters=parameters)
+        if df is None:
+            return None
```
kenxu95, line 133: nit: slightly more python for `return df.head(n) if df else None?

sdk/aqueduct/backend/api_client.py
kenxu95: @likawind was this addressed? I'm not sure it's clear to me what happens when you call this on a deleted artifact. It would be great if that was in the function's docstring too

kenxu95 (commented):
It's good, I just had one blocking comment about **get_artifact_result_data()**

sdk/aqueduct/backend/api_client.py
likawind: oh oh this line just prints a warning and doesn't affects the execution flow

kenxu95: oh I missed that lol. Can you note that DELETED -> None in the docstring though.

kenxu95 (approved):

sdk/aqueduct/artifacts/generic_artifact.py
likawind: I mean for fetched workflow runs, it's reasonable to call `.get()` even if the artifact is with fail / cancel / pending state with less blocking behavior like throwing this error

sdk/aqueduct/artifacts/table_artifact.py
likawind: oh actually I do this intentionally as I believe `if df` will be `False` if df is empty. But we can shrink that to one line though

sdk/aqueduct/artifacts/generic_artifact.py
kenxu95: yeah I think we just wanted to disambiguate a failed artifact vs an operator that returned None.

likawind: I see, in that way I will add a few lines to print warning message if the artifact is deleted

sdk/aqueduct/backend/api_client.py
likawind: yeah updated the warning message and docstring

cw75 (approved):
LGTM!

